PR #67, opened 2023-05-29
Bidirectional Forwarding Detection

### LINKED ISSUE #65: Want bidirectional forwarding detection (BFD)
As [discussed in RFD 267](https://rfd.shared.oxide.computer/rfd/0267#_basic) we need to implement BFD to support HA with static routing.

rcgoodfellow:
This is an initial implementation of the [RFC 5880: Bidirectional Forwarding Detection (BFD)](https://datatracker.ietf.org/doc/html/rfc5880#autoid-15) protocol. This PR is limited in scope to [RFC 5883: Bidirectional Forwarding Detection (BFD) for Multihop Paths](https://datatracker.ietf.org/doc/html/rfc5883) and [RFC 5881: Bidirectional Forwarding Detection (BFD) for IPv4 and IPv6 (Single Hop)](https://datatracker.ietf.org/doc/html/rfc5881). Authentication is not covered in this initial PR but will likely be added in a follow-on based on recommendations from RFC 5883.

~~This is purely a protocol implementation. No attempt is made to influence an underlying forwarding platform or interact with routing daemons.~~ BFD is now fully integrated with the Maghemite RIB.

The driving use case for BFD, as described in [RFD 276 Section 3.1](https://rfd.shared.oxide.computer/rfd/0267#_basic), is providing an HA path _into_ an Oxide rack for customers who use static routing. BFD is commonly implemented in enterprise-grade networking equipment for precisely this purpose. To make this direction work, we only need to implement the upper half of the protocol. This will inform upstream routers/firewalls about what sidecar to send packets to in the event of an outage. Integration with the Maghemite RIB allows our switches to detect upstream forwarding faults and react accordingly.

In addition to the CI-based tests that have been added. I have also tested this against the [VyOS/FRR BFD implementation](https://docs.vyos.io/en/latest/configuration/protocols/bfd.html), and the basic detection mechanisms look good. As we move forward, I'll work on integrating testing for FortiGate, PanOS, and others.

The BFD functionality is plumbed into the `mgd` API is designed to be driven by an overarching control plane. No static configuration is available whatsoever - everything is configured through the API. A basic CLI commands are provided in `mgadm` along with additions to the progenitor-based client library for `mfd`. ~~The daemon is also stateless, meaning that a higher-level entity is responsible for managing state across daemon restarts.~~ The deamon stores session information in the Maghemite RDB. It will restore BFD sessions on restart. The RDB does not survive a sled reboot, as its on the ramdisk. At that point, it's the control plane's responsibility to repopulate things. The I've tested that the Omicron BFD RPW performs this function as expected.

Finally, the core functionality in the BFD library is independent of network transport. This was done to make lightweight testing possible without having to set up complete networks with a bunch of zones/VMs and the whole nine yards. BFD sessions are linked to an underlying UDP mechanism through bidirectional channels that carry packet data and destination addresses. This makes it really simple to [define multi-daemon protocol tests](https://github.com/oxidecomputer/maghemite/blob/bfd/bfd/lib/src/lib.rs#L255-L280).

- Fixes #65


bfd/cli/src/main.rs
```diff
@@ -0,0 +1,124 @@
...
+        SubCommand::RemovePeer { peer } => {
+            client.remove_peer(&peer).await?;
+        }
+    }
+
+    Ok(())
+}
+
+pub fn get_styles() -> clap::builder::Styles {
+    clap::builder::Styles::styled()
```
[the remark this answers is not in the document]
rcgoodfellow: This code no longer exists as the bfd-specific CLI has been merged into `mgadm`

bfd/daemon/src/admin.rs
```diff
@@ -0,0 +1,182 @@
...
+    let ds_log = ConfigLogging::StderrTerminal {
+        level: ConfigLoggingLevel::Error,
+    }
+    .to_logger("admin")?;
+
+    let api = api_description();
+
+    info!(log, "admin: listening on {}", sa);
+
+    let log = log.clone();
```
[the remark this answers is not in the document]
rcgoodfellow: This got cleaned up in the move from a standalone daemon to running within the context of `mgd`.

bfd/daemon/src/udp.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/// Run an egress handler, taking BFD control packets from a session and sending
+/// the out to the peer over UDP.
+fn egress(rx: Receiver<(IpAddr, packet::Control)>, sk: Socket, log: Logger) {
+    spawn(move || loop {
+        let (addr, pkt) = match rx.recv() {
+            Ok(result) => result,
+            Err(e) => {
+                warn!(log, "udp egress channel closed: {e}");
```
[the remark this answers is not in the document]
rcgoodfellow: In the current code we do issue a `break` right below this line to stop the handler.

bfd/lib/src/sm.rs
```diff
@@ -0,0 +1,430 @@
...
+            }
+        }
+    }
+
+    fn state(&self) -> PeerState {
+        PeerState::Init
+    }
+}
+
+/// The BFD up state. The following is take verbatim from the RFC.
```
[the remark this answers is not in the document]
rcgoodfellow: Done

bfd/lib/src/packet.rs
```diff
@@ -0,0 +1,460 @@
...
+    /// 6. Concatenated Path Down
+    /// 7. Administratively Down
+    /// 8. Reverse Concatenated Path Down
+    /// 9-31. -- Reserved for future use
+    ///
+    pub vers_diag: u8,
+
+    /// This field contains both the status and flag values.
+    ///
+    /// The statis is the current BFD session state as seen by the transmitting
```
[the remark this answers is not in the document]
rcgoodfellow: Done

bfd/peer-test-config/vyos.vbash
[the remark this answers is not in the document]
rcgoodfellow: Added a comment explaining that to the beginning of this file.

bfd/lib/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+    Copy,
+    Clone,
+    TryFromPrimitive,
+    JsonSchema,
+    Serialize,
+    Deserialize,
+)]
+#[repr(u8)]
+pub enum PeerState {
+    /// A stable down state. Non-responsive to incomming messages.
```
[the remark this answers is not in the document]
rcgoodfellow: Thanks. Time to get a spell checker for `helix` ....

bfd/lib/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+    pub demand_mode: bool,
+
+    /// When multiplied against required_min_rx, defines the detection threshold
+    /// connectivity status.
+    pub detection_multiplier: u8,
+}
+
+impl Default for PeerInfo {
+    fn default() -> Self {
+        Self {
```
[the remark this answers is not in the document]
rcgoodfellow: The comment here was poorly worded. I've updated it to more closely follow the RFC.

> This is the minimum interval, in microseconds, between received
>  BFD Control packets that this system is capable of supporting,

bfd/lib/src/bidi.rs
```diff
@@ -0,0 +1,20 @@
...
+    pub tx: Sender<T>,
+}
+
+impl<T> Endpoint<T> {
+    fn new(rx: Receiver<T>, tx: Sender<T>) -> Self {
+        Self { rx, tx }
+    }
+}
+
+/// Analsgous to std::sync::mpsc::channel for bidirectional endpoints.
```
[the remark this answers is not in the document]
rcgoodfellow: Done

bfd/daemon/src/udp.rs
```diff
@@ -0,0 +1,111 @@
...
+    } else {
+        Domain::IPV6
+    };
+    let sk = Socket::new(domain, Type::DGRAM, Some(Protocol::UDP))?;
+    let sa = SocketAddr::new(listen, BFD_MULTIHOP_PORT).into();
+    sk.bind(&sa)?;
+
+    let (local, remote) = bidi::channel();
+
+    ingress(peer, remote.tx, sk.try_clone()?, log.clone());
```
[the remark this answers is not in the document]
rcgoodfellow: Done

bfd/daemon/src/udp.rs
```diff
@@ -0,0 +1,111 @@
...
+use socket2::{Domain, Protocol, Socket, Type};
+use std::mem::MaybeUninit;
+use std::net::{IpAddr, SocketAddr};
+use std::sync::mpsc::{Receiver, Sender};
+use std::thread::spawn;
+
+/// Port to be used for BFD multihop per RFC 5883.
+const BFD_MULTIHOP_PORT: u16 = 4784;
+
+/// Create a bidirectional chennel linking a peer session to an underlying BFD
```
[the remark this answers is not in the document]
rcgoodfellow: done

bfd/daemon/src/admin.rs
```diff
@@ -0,0 +1,182 @@
...
+            HttpError::for_internal_error(e.to_string())
+        })?;
+
+    let timeout = Duration::from_micros(rq.required_rx);
+    daemon.add_peer(rq.peer, timeout, rq.detection_threshold, ch);
+
+    Ok(HttpResponseUpdatedNoContent {})
+}
+
+/// Request to remove a peer form the daemon.
```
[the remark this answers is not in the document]
rcgoodfellow: done

bfd/lib/src/sm.rs
```diff
@@ -0,0 +1,430 @@
...
+    /// Create a new state machine, but do not start it.
+    pub fn new(
+        peer: IpAddr,
+        required_rx: Duration,
+        detection_multiplier: u8,
+        log: Logger,
+    ) -> Self {
+        let state = Down::new(peer, log.clone());
+        Self {
+            state: Arc::new(RwLock::new(Box::new(state))),
```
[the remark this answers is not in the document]
rcgoodfellow: > OK, we need the Box because the state is a dyn trait rather than a struct?

Yeah, that's the reason.

On the trait versus enum approaches, I've added a comment about this and filed a follow-up issue #153

bfd/lib/src/sm.rs
```diff
@@ -0,0 +1,430 @@
...
+use std::sync::mpsc::Sender;
+use std::sync::{Arc, Mutex, RwLock};
+use std::thread::sleep;
+use std::thread::spawn;
+use std::time::Duration;
+
+/// This state machine implements the BFD state machine. The following is taken
+/// directly from RFC 5880
+///
+/// ```text
```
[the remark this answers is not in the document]
rcgoodfellow: I've added a comment to this effect in the text before the diagram.

bfd/lib/src/util.rs
```diff
@@ -0,0 +1,159 @@
...
+            "[{:?}][{}] {}",
+            $self.state(),
+            $self.peer,
+            format!($($args)+)
+        )
+    }
+}
+
+#[macro_export]
+macro_rules! recv {
```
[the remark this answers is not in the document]
rcgoodfellow: I've transitioned this to a method on the `State` trait.

bfd/lib/src/util.rs
```diff
@@ -0,0 +1,159 @@
+use crate::{packet, PeerInfo};
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+#[macro_export]
```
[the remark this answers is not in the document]
rcgoodfellow: Agreed. Done.

bfd/lib/src/packet.rs
```diff
@@ -0,0 +1,460 @@
...
+            Err(_) => State::Unknown(status),
+        }
+    }
+
+    /// Set control packet state.
+    pub fn set_state(&mut self, ps: PeerState) {
+        self.flags = (self.flags & 0b00111111) | ((ps as u8) << 6);
+    }
+
+    /// Returns true of the poll flag is set.
```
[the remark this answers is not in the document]
rcgoodfellow: Perhaps, but let's just leave this as-is for the time being.

bfd/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+    pub fn add_peer(
+        &mut self,
+        peer: IpAddr,
+        required_rx: Duration,
+        detection_multiplier: u8,
+        mode: SessionMode,
+        endpoint: bidi::Endpoint<(IpAddr, packet::Control)>,
+        db: rdb::Db,
+    ) {
+        if self.sessions.contains_key(&peer) {
```
Nieuwejaar, line 48: I'm just getting started, so maybe there will be something that makes this obvious later... Is it expected that we would add the same peer multiple times?  If so, that might be worth commenting.  If not, then this should either be a log message or an error.

bfd/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+    /// The interval at which the peer would _like_ to receive BFD control
+    /// packets.
+    pub desired_min_tx: Duration,
+
+    /// This is the minimum interval, in microseconds, between received
+    /// BFD Control packets that this system is capable of supporting,
+    pub required_min_rx: Duration,
+
+    /// A unique identifer for the peer.
+    pub discriminator: u32,
```
Nieuwejaar, line 113: Assigned by this daemon or self-assigned by the peer?  What guarantees uniqueness?

bfd/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+}
+
+/// Information about a BFD peer.
+#[derive(Clone, Copy)]
+pub struct PeerInfo {
+    /// The interval at which the peer would _like_ to receive BFD control
+    /// packets.
+    pub desired_min_tx: Duration,
+
+    /// This is the minimum interval, in microseconds, between received
```
Nieuwejaar, line 108: I'm guessing that the two endpoints communicate in terms of microseconds, but the `Duration` we're using here can be on any scale.

bfd/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+            required_min_rx,
+            detection_multiplier,
+            discriminator: rand::random(),
+            ..Default::default()
+        }
+    }
+}
+
+/// The possible peer states. See the `State` trait implementations `Down`,
+/// `Init`, and `Up` for detailed semantics.
```
Nieuwejaar, line 158: Worth mentioning that the `repr(u8)` is because this is part of the wire protocol, not just internal usage?

mg-lower/src/dendrite.rs
```diff
@@ -55,19 +53,18 @@ pub(crate) fn ensure_tep_addr(
...
-        dpd.loopback_ipv6_create(&types::Ipv6Entry {
-            tag: MG_LOWER_TAG.into(),
-            addr: tep,
-        })
-        .await
-    }) {
-        if e.status() != Some(StatusCode::CONFLICT) {
-            warn!(log, "failed to ensure TEP address {tep} on ASIC: {e}");
+) -> Result<(), Error> {
+    rt.block_on(async {
```
Nieuwejaar, line 57: I don't understand the change.  The check-and-set operation isn't atomic, so we could still have a conflict.  If we expect there to be conflicts, then it seems like the old code is a better fit: Why flag an error if this is actually expected behavior?  If we don't expect conflicts, then why bother with the initial lookup operation?  Why not just do the `loopback_ipv6_create()`?

bfd/src/packet.rs
```diff
@@ -0,0 +1,464 @@
...
+    /// 7. Administratively Down
+    /// 8. Reverse Concatenated Path Down
+    /// 9-31. -- Reserved for future use
+    ///
+    pub vers_diag: u8,
+
+    /// This field contains both the status and flag values.
+    ///
+    /// The status is the current BFD session state as seen by the transmitting
+    /// system. It's values are:
```
Nieuwejaar, line 70: s/It's/Its

bfd/src/packet.rs
```diff
@@ -0,0 +1,464 @@
...
+
+// Control packet flags.
+const POLL: u8 = 1 << 5;
+const FINAL: u8 = 1 << 4;
+const CONTROL_PLANE_INDEPENDENT: u8 = 1 << 3;
+const AUTHENTICATION_PRESENT: u8 = 1 << 2;
+const DEMAND: u8 = 1 << 1;
+const MULTIPOINT: u8 = 1;
+
+/// The BFD control packet. The commentary on this structure comes directly from
```
Nieuwejaar, line 16: This is all very helpful.  Thanks for including it.

bfd/src/packet.rs
```diff
@@ -0,0 +1,464 @@
...
+            required_min_rx: 0,
+            required_min_echo_rx: 0,
+            auth: None,
+        }
+    }
+}
+
+impl Control {
+    /// Wrap an array of bytes as a `Control`.
+    pub fn wrap(d: &[u8]) -> Result<Self> {
```
Nieuwejaar, line 179: It seems like this should be `from_bytes` for symmetry with the next fucntion.

Nieuwejaar (commented):
I still have a few files to get through, but I'm done for the night.  So far, this looks really solid.

bfd/src/packet.rs
rcgoodfellow: Agreed. Done

bfd/src/packet.rs
rcgoodfellow: Done

mg-lower/src/dendrite.rs
rcgoodfellow: Yeah, the old code was a better overall fit. I was trying to avoid an unfortunate interaction between dendrite table size tracking and softnpu's lack of table conflict errors. Basically, when softnpu has a table update that conflicts with a current entry, it overwrites that entry instead of returning an error, but the dendrite table size tracking assumes there will be an error on conflict. So conflicting entries that are overwritten by softnpu increase dendrites table size counters without actually increasing the table size. I'll go fix this in softnpu/dendrite instead.

rcgoodfellow: I've reverted this back to its previous form.

bfd/src/lib.rs
rcgoodfellow: Done

bfd/src/lib.rs
rcgoodfellow: Fixed.

bfd/src/lib.rs
rcgoodfellow: I've expanded on the doc comment explaining this.

bfd/src/lib.rs
rcgoodfellow: Generally speaking, this should not happen. The omicron RPW should calculate additions in a way that does not result in peers being added multiple times. I've added a warning-level log here.

mgd/src/bfd_admin.rs
```diff
@@ -0,0 +1,393 @@
...
+
+    // Spawn an egress thread to take packets from the session and send them
+    // out a UDP socket.
+    egress(remote.rx, listen, src_port, dst_port, log.clone());
+
+    Ok(local)
+}
+
+/// Run an egress handler, taking BFD control packets from a session and sending
+/// the out to the peer over UDP.
```
Nieuwejaar, line 220: s/the/them

mgd/src/bfd_admin.rs
```diff
@@ -0,0 +1,393 @@
...
+    ) {
+        loop {
+            if kill_switch.load(std::sync::atomic::Ordering::Relaxed) {
+                break;
+            }
+            // Maximum length of a BFD packet is on the order of 100 bytes (RFC
+            // 5880).
+            let mut buf = [0; 1024];
+
+            let (n, sa) = match sk.recv_from(&mut buf) {
```
Nieuwejaar, line 362: Since the state machine is isolated from the network thread by a channel, I don't see how a timeout in the state machine propagates back to this thread.  Why don't we block in this receive forever?

Update: oh, because we've set a 1 second timeout above.  Does that mean we're writing out an `error` log message every second?

mgd/src/bfd_admin.rs
```diff
@@ -0,0 +1,393 @@
...
+                None => {
+                    warn!(log, "unknown peer {}, dropping", sa.ip(),);
+                    continue;
+                }
+            };
+
+            let pkt = match packet::Control::from_bytes(&buf[..n]) {
+                Ok(pkt) => pkt,
+                Err(e) => {
+                    error!(log, "parse control packet: {e}");
```
Nieuwejaar, line 382: Add the peer IP to the error log?

bfd/src/sm.rs
```diff
@@ -0,0 +1,558 @@
...
+use std::thread::spawn;
+use std::time::Duration;
+
+/// This state machine implements the BFD state machine. The following is taken
+/// directly from RFC 5880. The arrows indicate the state machine transition
+/// that is taken when a packet containing the peer state on the label is
+/// received.
+///
+/// ```text
+///                             +--+
```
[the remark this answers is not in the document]
rcgoodfellow: Yeah the phrasing of the state machine and its transitions is a bit strange. This diagram comes directly from the RFC. Here is the associated explanation.

> A session remains in Down state until the remote system indicates
   that it agrees that the session is down by sending a BFD Control
   packet with the State field set to anything other than Up.  If that
   packet signals Down state, the session advances to Init state; if
   that packet signals Init state, the session advances to Up state.

bfd/src/sm.rs
```diff
@@ -0,0 +1,558 @@
...
+            peer,
+            endpoint.tx.clone(),
+            local,
+            remote.clone(),
+            kill_switch.clone(),
+            log.clone(),
+        );
+
+        spawn(move || loop {
+            let prev = state.read().unwrap().state();
```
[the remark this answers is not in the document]
rcgoodfellow: Yeah, there is definitely room for improvement here. However, I think it's not quite this simple since the run functions of individual states vary in both preamble code and the transitions they make within their own processing loops. I'm going to revisit the state machine design in #153, which was a result of the first code review on BFD, and would like to defer refactoring the state machine code till then if possible.

bfd/src/sm.rs
```diff
@@ -0,0 +1,558 @@
...
+        sleep(self.0);
+    }
+}
+
+pub enum RecvResult {
+    MessageFrom((IpAddr, Control)),
+    TransitionTo(Box<dyn State>),
+}
+
+//TODO consider using a `State` enum instead of identical structs that
```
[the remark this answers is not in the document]
rcgoodfellow: Yeah, this comment was a result of your feedback on the previous review.

mgd/src/bfd_admin.rs
rcgoodfellow: Done. Also added to the other log entires in this scope.

mgd/src/bfd_admin.rs
rcgoodfellow: I've added a predicate to only log if this is not a timeout. I think, typically, we will not see this as BFD communications are usually inside one second.

mgd/src/bfd_admin.rs
rcgoodfellow: Done

PR #156, opened 2024-01-25
backwards compatibility with ddm protocol version 1

rcgoodfellow:
This PR now covers backward compatibility with version one of the ddm exchange protocol.

Ddm has two protocols, a router discovery protocol and a prefix exchange protocol. The tunnel routing updates did not change the discovery protocol - and that protocol carries an 8-bit version in it. I updated the exchange client/server code to speak either v1 or v2 of the protocol based on what the peer advertised in the discovery phase. I also added backward compatibility tests to CI.

### DIFF
--- a/ddm/src/discovery.rs
+++ b/ddm/src/discovery.rs
@@ -103,9 +103,15 @@ use thiserror::Error;
 
 const DDM_MADDR: Ipv6Addr = Ipv6Addr::new(0xff02, 0, 0, 0, 0, 0, 0, 0xdd);
 const DDM_PORT: u16 = 0xddd;
-const VERSION_1: u8 = 1;
 const SOLICIT: u8 = 1;
 const ADVERTISE: u8 = 1 << 1;
+
+#[derive(Debug, Copy, Clone)]
+#[repr(u8)]
+pub enum Version {
+    V1 = 1,
+    V2 = 2,
+}
 
 #[derive(Error, Debug)]
 pub enum DiscoveryError {
@@ -128,7 +134,7 @@ pub struct DiscoveryPacket {
 impl DiscoveryPacket {
     pub fn new_solicitation(hostname: String, kind: RouterKind) -> Self {
         Self {
-            version: VERSION_1,
+            version: Version::V2 as u8,
             flags: SOLICIT,
             hostname,
             kind,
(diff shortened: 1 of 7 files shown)


ddm/src/exchange.rs
```diff
@@ -166,8 +166,14 @@ pub enum ExchangeError {
     #[error("hyper error: {0}")]
     Hyper(#[from] hyper::Error),
 
+    #[error("hyper http error: {0}")]
+    HyperHttp(#[from] hyper::http::Error),
+
     #[error("timeout error: {0}")]
     Timeout(#[from] tokio::time::error::Elapsed),
+
+    #[error("timeout error: {0}")]
```
leftwo, line 175: This should probalby say json error or something (it's not a timeout, right?)

rcgoodfellow: Good call. Fixed.

ddm/src/discovery.rs
```diff
@@ -128,15 +134,15 @@ pub struct DiscoveryPacket {
 impl DiscoveryPacket {
     pub fn new_solicitation(hostname: String, kind: RouterKind) -> Self {
         Self {
-            version: VERSION_1,
+            version: Version::V2 as u8,
```
rmustacc, line 137: Apologies as I've not spent much time in this, but how does hardcoding version 2 for the different discovery packets actually fit in terms of backwards compatibility? If someone only understood version 1, they'd presumably ignore the solicitation or advertisement, no?

rcgoodfellow: Version 1 does not do anything with the version value in the discovery header, except for fill it in with the value `1` and send it out. The backwards compatibility part comes in with how the routers behave when they get to the exchange phase, which comes after discovery. If a version `2` router detects a version `1` peer through discovery, it will speak the version `1` protocol for exchange.

rcgoodfellow: I've added a comment on this in [1f95198](https://github.com/oxidecomputer/maghemite/pull/156/commits/1f95198d71a84344f225e7047b84497d3965d601)

rcgoodfellow: Follow in issue in 
- #157

ddm/src/discovery.rs
```diff
@@ -396,6 +396,16 @@ fn handle_advertisement(
 ) {
     trc!(&ctx.log, ctx.config.if_name, "advert from {}", &hostname);
 
+    // TODO: version negotiation
+    //
+    // Things currently work because ddm v1 one does no version checking at all.
```
rmustacc, line 401:
```suggestion
// Things currently work because ddm v1 does no version checking at all.
```

PR #168, opened 2024-02-08
ddm: persist originated prefixes and tunnel endpoints

### LINKED ISSUE #140: ddmd: persist originated prefixes
When `ddmd` restarts, it loses all state, relying on a higher level control plane to re-establish state. For locally established state, originated prefixes in particular, this complicates daemon restart robustness. The `ddmd` daemon should persist originated prefixes to disk and reload them on restart.

rcgoodfellow:
- fixes #140

### DIFF
--- a/ddm/src/db.rs
+++ b/ddm/src/db.rs
@@ -5,30 +5,61 @@
 use schemars::{JsonSchema, JsonSchema_repr};
 use serde::{Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
+use slog::{error, Logger};
 use std::collections::{HashMap, HashSet};
 use std::net::{AddrParseError, IpAddr, Ipv4Addr, Ipv6Addr};
 use std::num::ParseIntError;
 use std::sync::{Arc, Mutex};
 use thiserror::Error;
 
-#[derive(Default, Clone)]
+/// The handle used to open a persistent key-value tree for originated
+/// prefixes.
+const ORIGINATE: &str = "originate";
+
+/// The handle used to open a persistent key-value tree for originated
+/// tunnel endpoints.
+const TUNNEL_ORIGINATE: &str = "tunnel_originate";
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("datastore error {0}")]
+    DataStore(#[from] sled::Error),
+
+    #[error("db key error{0}")]
+    DbKey(String),
+
+    #[error("db value error{0}")]
+    DbValue(String),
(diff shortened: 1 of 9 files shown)


ddm/src/db.rs
```diff
@@ -221,6 +332,31 @@ pub struct Ipv6Prefix {
...
+        buf.push(self.len);
+        buf
+    }
+
+    pub fn from_db_key(v: &[u8]) -> Result<Self, Error> {
+        if v.len() < 5 {
+            Err(Error::DbKey(format!(
+                "buffer to short for prefix 4 key {} < 5",
+                v.len()
+            )))
```
FelixMcFelix, line 347: Should this length check be against 17? `std::mem::size_of::<Self>()` would work, but that's maybe a bit fragile. The error message below also mentions prefix4s, when I think you're only storing v6 prefixes.

ddm/src/db.rs
```diff
@@ -221,6 +332,31 @@ pub struct Ipv6Prefix {
...
+        } else {
+            Ok(Self {
+                addr: Ipv6Addr::from([
+                    v[0], v[1], v[2], v[3], v[4], v[5], v[6], v[7], v[8], v[9],
+                    v[10], v[11], v[12], v[13], v[14], v[15],
+                ]),
```
FelixMcFelix, line 353:
```suggestion
                addr: Ipv6Addr::from(<[u8; 16]>::try_from(&v[..16]).unwrap()),
```
I couldn't quite get the cleaner `&v[..16].try_into()` to typecheck here, but this might be a bit less repetitive.

FelixMcFelix (commented):
Thanks, I think there's just one fix on a parse length check? Otherwise it seems sensible to use the same sled db as you're already storing BGP state in. :)

ddm/src/db.rs
```diff
@@ -5,30 +5,61 @@
...
+/// The handle used to open a persistent key-value tree for originated
+/// tunnel endpoints.
+const TUNNEL_ORIGINATE: &str = "tunnel_originate";
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("datastore error {0}")]
+    DataStore(#[from] sled::Error),
+
+    #[error("db key error{0}")]
```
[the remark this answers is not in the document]
rcgoodfellow: Done

ddm/src/db.rs
rcgoodfellow: Done

FelixMcFelix (approved):
Great, looks good!

PR #180, opened 2024-03-21
track ddm state durations

### LINKED ISSUE #54: ddm: track peer establishment time
This way we can see how long sessions have been alive.

rcgoodfellow:
This PR tracks the times of DDM session state transitions internally and provides that information through an new `GET /peers-v2` API endpoint. The existing `GET /peers` endpoint that control planes depend on remains in place.

- Fixes #54


taspelund:
Closing, as this work was re-implemented in #689

PR #182, opened 2024-03-25
mgd and mg-ddm: make rack-uuid and sled-uuid optional

jgallagher:
If these are not provided, we will not start an `oxstats` server.

Part of fixing https://github.com/oxidecomputer/omicron/issues/5318


bnaecker:
I missed this in review, that was my mistake. I think in the long-run we should operate the way `dpd` does, which is watch SMF properties for changes, and then start the metrics producer server when they become available. That is more work, so could be a separate issue.

bnaecker (approved):
Looks good to me, thanks for fixing this!

PR #184, opened 2024-03-25
SMF refresh

### LINKED ISSUE #183: mgd, mg-ddm: Support `smf refresh` to start oxstats server
I missed this in review, that was my mistake. I think in the long-run we should operate the way `dpd` does, which is watch SMF properties for changes, and then start the metrics producer server when they become available. That is more work, so could be a separate issue.

_Originally posted by @bnaecker in https://github.com/oxidecomputer/maghemite/issues/182#issuecomment-2018268223_

rcgoodfellow:
- Fixes #183

### DIFF
--- a/ddmd/src/signal.rs
+++ b/ddmd/src/signal.rs
@@ -1,0 +1,34 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use crate::smf::smf_refresh;
+use ddm::admin::HandlerContext;
+use slog::{error, info, Logger};
+use std::sync::{Arc, Mutex};
+use tokio::signal::unix::{signal, SignalKind};
+use tokio::sync::mpsc::Receiver;
+
+pub(crate) async fn handle_signals(
+    mut ch: Receiver<Arc<Mutex<HandlerContext>>>,
+    log: Logger,
+) {
+    let mut sigusr1 = signal(SignalKind::user_defined1()).unwrap();
+    let mut future_signal = sigusr1.recv();
+
+    info!(log, "signal handler waiting for context");
+    let ctx = loop {
+        match ch.recv().await {
+            Some(ctx) => break ctx,
+            None => continue,
+        }
+    };
+
+    loop {
+        future_signal.await;
+        if let Err(e) = smf_refresh(ctx.clone(), log.clone()).await {
+            error!(log, "smf update on sigusr1 failed: {e}");
+        }
+        future_signal = sigusr1.recv();
+    }
+}
--- a/ddmd/src/signal.rs
+++ b/ddmd/src/signal.rs
@@ -13,22 +13,24 @@ pub(crate) async fn handle_signals(
     mut ch: Receiver<Arc<Mutex<HandlerContext>>>,
     log: Logger,
 ) {
-    let mut sigusr1 = signal(SignalKind::user_defined1()).unwrap();
-    let mut future_signal = sigusr1.recv();
+    tokio::spawn(async move {
+        let mut sigusr1 = signal(SignalKind::user_defined1()).unwrap();
+        let mut future_signal = sigusr1.recv();
 
-    info!(log, "signal handler waiting for context");
-    let ctx = loop {
-        match ch.recv().await {
-            Some(ctx) => break ctx,
-            None => continue,
-        }
-    };
+        info!(log, "signal handler waiting for context");
+        let ctx = loop {
+            match ch.recv().await {
+                Some(ctx) => break ctx,
+                None => continue,
+            }
+        };
 
-    loop {
-        future_signal.await;
-        if let Err(e) = smf_refresh(ctx.clone(), log.clone()).await {
-            error!(log, "smf update on sigusr1 failed: {e}");
+        loop {
+            future_signal.await;
+            if let Err(e) = smf_refresh(ctx.clone(), log.clone()).await {
+                error!(log, "smf update on sigusr1 failed: {e}");
+            }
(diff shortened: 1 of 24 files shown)


mgd/src/smf.rs
```diff
@@ -0,0 +1,182 @@
...
+    name: &str,
+    pg: &PropertyGroup<'_>,
+) -> anyhow::Result<String> {
+    let prop = match pg
+        .get_property(name)
+        .map_err(|e| anyhow!("smf get-prop {name}: {e}"))?
+    {
+        Some(p) => p,
+        None => anyhow::bail!("smf property {name} does not exist"),
+    };
```
jclulow, line 133: Perhaps:

```suggestion
    let prop = pg.get_property(name)
        .map_err(|e| anyhow!("smf get-prop {name}: {e}"))?
        .ok_or_else(|| anyhow!("smf property {name} does not exist"))?;
```

mgd/src/smf.rs
```diff
@@ -0,0 +1,182 @@
...
+pub(crate) async fn smf_refresh(
+    ctx: Arc<HandlerContext>,
+    log: Logger,
+) -> anyhow::Result<()> {
+    let scf = smf::Scf::new()
+        .map_err(|e| anyhow::anyhow!("create scf handle: {e}"))?;
+
+    let instance = scf
+        .get_self_instance()
+        .map_err(|e| anyhow::anyhow!("create smf instance: {e}"))?;
```
jclulow, line 24: I've been considering reworking the errors here a little so that they include the name of the function in the string that would come out of `Display`; e.g.,

`Scf::get_self_instance: Scf::get_instance_from_frmi: scf_handle_decode_fmri: no memory available`

In the debug rendering it might display `SCF_ERROR_NOT_SET` instead of `no memory available`.

Would that mean you could cut down on all the `.map_err(...anyhow...)` stuff in here?

rcgoodfellow: Yeah, it would. The worry here is getting log messages like "error: 22 EINVAL" which are meaningless without context.

mgd/src/smf.rs
rcgoodfellow: Yeah, that's nicer. Done. Thanks

rcgoodfellow:
Still need to run through testing on madrid where this all began - but I think this is ready for a look

ddmd/src/signal.rs
```diff
@@ -0,0 +1,37 @@
...
+use std::sync::{Arc, Mutex};
+use tokio::signal::unix::{signal, SignalKind};
+use tokio::sync::mpsc::Receiver;
+
+pub(crate) async fn handle_signals(
+    mut ch: Receiver<Arc<Mutex<HandlerContext>>>,
+    log: Logger,
+) {
+    tokio::spawn(async move {
+        let mut sigusr1 = signal(SignalKind::user_defined1()).unwrap();
```
jgallagher, line 17: Not sure about this unwrap - does this repo build with [panic=abort](https://github.com/oxidecomputer/omicron/blob/1244d5918dabf8804f5ab5570ef6cee1d3c02b9f/Cargo.toml#L498-L499)? If not, I think this panic would cause the signal handler to just not run, and the only note about it would be a panic message in the logfile.

I think you could call `signal()` before entering the `tokio::spawn`, allowing `handle_signals` to return an error that could bubble up and out of `main()` if we can't attach the signal handler? Alternatively, if that's what we want (i.e., a failure to attach the signal handler isn't fatal), maybe log the error and return instead of unwrapping?

ddmd/src/signal.rs
```diff
@@ -0,0 +1,37 @@
...
+        let mut future_signal = sigusr1.recv();
```
jgallagher, line 18: Tiny nit - I don't think this variable necessary; you could make the first line of the `loop`

```rust
sigusr1.recv().await;
```

and the behavior would be the same. IIUC, it's calling the `signal()` function that attaches the handler; if the process gets a signal in between that and the first call to `recv()`, the `recv()` will return immediately whenever it does get called.

ddmd/src/smf.rs
```diff
@@ -0,0 +1,85 @@
...
+                DDM_STATS_PORT,
+                context.peers.clone(),
+                context.stats.clone(),
+                props.dns_servers,
+                hostname,
+                props.rack_uuid,
+                props.sled_uuid,
+                log.clone(),
+            )
+            .unwrap(),
```
jgallagher, line 78: Similar comment about this `unwrap()` - if it fails, maybe instead log an error and leave the `handler` as `None`? At least then it _could_ try again on the next sigusr1.

mgd/src/signal.rs
```diff
@@ -0,0 +1,37 @@
...
+use std::sync::Arc;
+use tokio::signal::unix::{signal, SignalKind};
+use tokio::sync::mpsc::Receiver;
+
+pub(crate) async fn handle_signals(
+    mut ch: Receiver<Arc<HandlerContext>>,
+    log: Logger,
+) {
+    tokio::spawn(async move {
+        let mut sigusr1 = signal(SignalKind::user_defined1()).unwrap();
```
jgallagher, line 17: (Identical comments as those left on `ddmd::handle_signals`)

mgd/src/smf.rs
```diff
@@ -0,0 +1,81 @@
...
+        crate::oxstats::start_server(
+            props.admin_addr,
+            ctx.clone(),
+            props.dns_servers,
+            hostname,
+            props.rack_uuid,
+            props.sled_uuid,
+            log.clone(),
+        )
+        .unwrap();
```
jgallagher, line 74: Same note about replacing this with an error log and leaving `is_running` `false`.

jgallagher (approved):
Looks great - thanks! Happy to give this a spin on madrid.

PR #199, opened 2024-04-07
0xPop

### LINKED ISSUE #147: bgp: override configured hold/keepalive time with proposal from peer if smaller
According to the RFC section 4.2

```
Hold Time:

         This 2-octet unsigned integer indicates the number of seconds
         the sender proposes for the value of the Hold Timer.  Upon
         receipt of an OPEN message, a BGP speaker MUST calculate the
         value of the Hold Timer by using the smaller of its configured
         Hold Time and the Hold Time received in the OPEN message.  The
         Hold Time MUST be either zero or at least three seconds.  An
         implementation MAY reject connections on the basis of the Hold
         Time.  The calculated value indicates the maximum number of
         seconds that may elapse between the receipt of successive
         KEEPALIVE and/or UPDATE messages from the sender.
```

Then according to Section 10.

```
   The KeepaliveTime is a mandatory FSM attribute that stores the
   initial value for the KeepaliveTimer.  The suggested default value
   for the KeepaliveTime is 1/3 of the HoldTime.
```

So if the hold time is modified, we should set the keepalive to `hold_time / 3`.

### LINKED ISSUE #151: Allow operator to specify ASN of BGP Peer
Currently the `bgp_peers` config does not have a parameter to specify a remote ASN, leaving it to be learnt dynamically via the exchange of BGP Open messages.

This means the BGP stack will always dynamically learn the peer's ASN + accept it unconditionally, which could potentially risk allowing a misconfig or security event to occur.

### LINKED ISSUE #152: Add basic BGP security features
Issue is to track the addition of basic BGP security features.
1. TTL Security
2. MD5 Auth on TCP session

Item (1) allows an operator to define a minimum TTL value on received BGP packets in order for them to be considered valid.  Typically this is used on point-to-point eBGP sessions w/ a value of 255 or similar, to prove that the packet isn't being spoofed or routed from a source elsewhere in the network.

Item (2) allows an operator to define an auth key to be used with the TCP MD5 Option.  This ensures authentication of BGP packets will fail unless the peer has also defined the same auth key, i.e. it is an anti-spoofing measure.

### LINKED ISSUE #194: bgp: multi-exit discriminator support
The multi-exit discriminator attribute is defined in [RFC 4271 section 5.1.4](https://datatracker.ietf.org/doc/html/rfc4271#section-5.1.4) as

> The MULTI_EXIT_DISC is an optional non-transitive attribute that is
   intended to be used on external (inter-AS) links to discriminate
   among multiple exit or entry points to the same neighboring AS.  The
   value of the MULTI_EXIT_DISC attribute is a four-octet unsigned
   number, called a metric.  All other factors being equal, the exit
   point with the lower metric SHOULD be preferred.

This allows Oxide rack operators to influence traffic ingress from adjacent ASes. In particular it allows control of active/active configurations versus active/passive configurations. It can be used in concert with graceful shutdown to avoid packet loss during maintenance operations, manage contingencies, and give operators control over traffic (a)symmetry.

### LINKED ISSUE #195: bgp: operator defined communities
When peering with external parties, it's often required to leverage communities supported by adjacent ASes to get to desired or optimal traffic flows. A nice example of this is described in the Cloudflare blog post [Why BGP communities are better than AS-path prepends](https://blog.cloudflare.com/prepends-considered-harmful). Mgd should provide an API for operators to selectively add custom communities chosen by operators based on ASN, nexthop, prefix, etc.

### LINKED ISSUE #196: bgp: add support for import/export policy
We need support for filtering inbound and outbound prefix updates by peer, ASN, prefix, etc. 

As a simple use case, consider the scenario where an Oxide rack is peered with both transit providers and private VPCs within public clouds. The former are on public address spaces, and the latter are on private address spaces that only make sense within the context of the public cloud VPC and a private on-prem network address space. Blasting out all routes to both types of peers makes no sense, will certainly result in notifications, and potentially BGP session failure.

### LINKED ISSUE #197: bgp: add support for operator defined local preferences
Operators should be able to define local preferences for

- routes learned via BGP based on ASN, nexthop, prefix, etc
- the value of LOCAL_PREF carried in outgoing IBGP updates

### LINKED ISSUE #200: bgp: idle hold timer should not be running outside idle state
During testing I'm seeing this in the `mgd` logs

```
WARN slog-rs: [transit] event idle hold timer expires not allowed in connect
```

This timer should not be running unless we're in the idle state.

### LINKED ISSUE #205: Compute active prefix ECMP group based on multiple independent factors
Currently, things that have nothing to do with one another, like path length and MED, are being considered simultaneously under a single priority metric. We should probably have a more structured selection process that first partitions based on deal breakers like GSHUT or BFD-disqualified routes, then prioritizes based on path length, then on MED or local priority, etc.

### LINKED ISSUE #208: bgp: enforce-first-as
`enforce-first-as` is a common security policy option in BGP implementations that ensures routes received from an eBGP peer in ASN X carry an AS-Path whose left-most ASN is X.

This is useful as a baseline mechanism to ensure a peer isn't prepending the AS-Path with ASNs they don't reside within, however must be disabled when peering with an eBGP route-server (e.g. at an IXP) since they do not prepend their own ASN to the AS-Path.

As we add more security options to this BGP implementation, this may be a useful one to add as well.

### LINKED ISSUE #209: bgp: add martian detection
As self-preservation mechanisms, many BGP implementations support the detection and invalidation/discarding of "martian" entries learned from BGP peers.
i.e.
If a BGP peer advertises an NLRI or Path Attribute that describes something we believe to be local to the router itself, we should detect this and (in most cases) invalidate/discard it.

It's usually a much more sane approach to have the control plane detect and react to these conditions (e.g. by marking NLRI ineligible for install while a corresponding local entry exists, installing just these previously-ineligible NLRI when a corresponding local entry is removed) than it is to rely on the data plane to do these validations + handle error conditions on the fly. 

Examples of "martian" entries received from a peer:
1. Prefix within IPv4/IPv6 NLRI matches a local interface/TEP address (generally /32 or /128)
2. NEXT-HOP matches a local interface/TEP address
3. RD (Route Distinguisher) within NLRI matches a locally configured VRF
4. EVPN MAC within NLRI matches a local interface MAC
5. EVPN RMAC Extended Community matches a local interface MAC (mapping to an IP-VRF)
6. Site-of-Origin Extended Community matches a locally configured "Site" ID

In the current maghemite implementation, likely the only applicable martian types are (1) and (2), however that might expand over time as the implementation evolves.

### LINKED ISSUE #210: bgp: ipv4 unicast NRLI sanity checking
Add sanity checks for the contents of IPv6 Unicast NLRI, i.e. making sure they only contain valid unicast addresses.

Ranges to drop as invalid:

IPv4:
- 127.0.0.0/8
- 224.0.0.0/4
- 240.0.0.0/4
- 169.254.0.0/16 (optional, as this range is sometimes used as loopback ranges similar to RFC 1918)

### LINKED ISSUE #215: bgp: route refresh / enhanced route refresh
Route refresh is needed to implement dynamic inbound routing policy for peers.

Route refresh is defined in

- [RFC 2918](https://datatracker.ietf.org/doc/html/rfc2918)
- [RFC 7313](https://datatracker.ietf.org/doc/html/rfc7313)

### LINKED ISSUE #234: Building mgd on macOS with `--no-default-features` stopped working
As of https://github.com/oxidecomputer/maghemite/commit/8207cb9c I can't build maghemite on macOS anymore.

```
$ cargo build --no-default-features --bin mgd

.........

error: cannot find macro `compile_fail` in this scope
   --> /Users/david/.cargo/git/checkouts/netadm-sys-25ac2db9b1b244c6/1d75565/libnet/src/sys.rs:703:13
    |
703 |             compile_fail!("only linux and illumos are currently supported")
    |             ^^^^^^^^^^^^

   Compiling ciborium-ll v0.2.2
error[E0412]: cannot find type `ioc_t` in this scope
  --> /Users/david/.cargo/git/checkouts/netadm-sys-25ac2db9b1b244c6/1d75565/libnet/src/sys.rs:54:20
   |
54 | pub const DLD_IOC: ioc_t = 0x0D1D;
   |                    ^^^^^ not found in this scope
```

I think this is because a dependency on `libnet` was added to `mg-common`. It seems designed to be left out of non-illumos builds, but the dependency on libnet is not wrapped in a default feature.

https://github.com/oxidecomputer/maghemite/blob/77246a4e07c86782313184cd1aa73bd2d162db70/mg-common/src/nexus.rs#L99-L116

Maybe I'm misunderstanding how this is supposed to work, because this change did not fix it.

```diff
diff --git a/mg-common/Cargo.toml b/mg-common/Cargo.toml
index 5f3abcc..9101b6c 100644
--- a/mg-common/Cargo.toml
+++ b/mg-common/Cargo.toml
@@ -19,4 +19,7 @@ oximeter.workspace = true
 backoff.workspace = true
 smf.workspace = true
 uuid.workspace = true
-libnet.workspace = true
+libnet = { workspace = true, optional = true }
+
+[features]
+default = ["libnet"]
```

rcgoodfellow:
This PR knocks off the following issues.

- Fixes #151
- Fixes #152
- Fixes #147
- Fixes #194 
- Fixes #195
- Fixes #197
- Fixes #205
- Fixes #200
- Fixes #208
- Fixes #210
- Fixes #209
- Fixes #196
- Fixes #215 
- Fixes #234

Additionally, the BGP API is "normalized" in the sense that everything is now organized around resources, and every resource is a full set of CRUD operations. This is to facilitate better interactions with automation frameworks like Terraform. There is currently a [work in progress terraform provider](https://github.com/oxidecomputer/terraform-provider-maghemite) for maghemite. This also greatly helped in standing up testing environments where we iterate through a number of configuration states and make sure things land where we expect in container lab and bespoke falcon environments.


rdb/src/db.rs
```diff
@@ -581,4 +596,74 @@ impl Db {
...
+            let targets: Vec<Path> = path
+                .iter()
+                .filter_map(|p| {
+                    if let Some(bgp) = p.bgp.as_ref() {
+                        if bgp.id == id {
+                            let mut marked = p.clone();
+                            marked.bgp = Some(bgp.as_stale());
+                            return Some(marked);
+                        }
+                    }
```
taspelund, line 624: Is this relying on the peer's Router ID to identify routes learned from a particular BGP session?  That would probably suffice in the short term, but may pose problems if we run multiple sessions to the same peer (e.g. over redundant L3 links)

rcgoodfellow: Created a follow-up issue.

- https://github.com/oxidecomputer/maghemite/issues/241

mgd/src/bgp_admin.rs
```diff
@@ -52,226 +51,97 @@ impl BgpContext {
...
+    register!(api, delete_checker);
+
+    // Policy shaper configuration
+    register!(api, create_shaper);
+    register!(api, read_shaper);
+    register!(api, update_shaper);
+    register!(api, delete_shaper);
+
+    // Omicron API XXX? use normal API now that it's somewhat civilized?
+    register!(api, bgp_apply);
```
internet-diglett, line 106: We need to refactor some things in that omicron RPW anyway, so we can move away from calling this api endpoint when we do that.

rdb/src/bestpath.rs
```diff
@@ -0,0 +1,194 @@
...
+    // Partition the choice space on whether routes are shutdown or not. If we
+    // only have shutdown routes then use those. Otherwise use active routes
+    let (active, shutdown): (BTreeSet<&Path>, BTreeSet<&Path>) =
+        candidates.iter().partition(|x| x.shutdown);
+    let candidates = match (active.len(), shutdown.len()) {
+        (0, _) => shutdown,
+        (_, _) => active,
+    };
+
+    // Filter down to paths that are not stale
```
internet-diglett, line 49: It wasn't very obvious to me how this worked at first, but maybe others are more
familiar with `itertools`. If you think this comment is valuable feel free to add it :)

```suggestion
    // Filter down to paths that are not stale
    // The `min_set_by_key` method allows us to assign "not stale" paths to the `0`
    // set, and "stale" paths to the `1` set. The method will then return the `0` set.
```

interop-lab/create-containers.sh
```diff
@@ -0,0 +1,36 @@
...
+        -e container=docker \
+        -i \
+        -t ceos:4.28.3M \
+        /sbin/init \
+        systemd.setenv=INTFTYPE=eth \
+        systemd.setenv=ETBA=1 \
+        systemd.setenv=SKIP_ZEROTOUCH_BARRIER_IN_SYSDBINIT=1 \
+        systemd.setenv=CEOS=1 \
+        systemd.setenv=EOS_PLATFORM=ceoslab \
+        systemd.setenv=container=docker
```
internet-diglett, line 22: If you're finding the docker containers valuable, I can probably replace a lot of these setup scripts with a `docker-compose.yml` file that will make it easier to set up, run, stop, and cleanup the container environment.

rdb/src/db.rs
```diff
@@ -45,22 +48,35 @@ const TEP_KEY: &str = "tep";
 /// information.
 const BFD_NEIGHBOR: &str = "bfd_neighbor";
 
+//TODO as parameter
+const BESTPATH_FANOUT: usize = 1;
```
internet-diglett, line 52: Are we only supporting one path per prefix at the moment?

internet-diglett (commented):
Amazing work, this is a huge PR! I've reached a stopping point for today's review session and leaving some notes. So far I only have some very minor items and a few questions.

rdb/src/bestpath.rs
```diff
@@ -0,0 +1,194 @@
...
+///   local preference
+/// - filter the selection group to the set of paths with the smallest
+///   AS path lenght
+/// - filter the selection group to the set of paths with the smallest
+///   multi-exit discriminator (MED) on a per-AS basis.
+///
+/// Upon completion of these filtering operations, if the selection group
+/// is larger than `max`, return the first `max` entries. This is a set,
+/// so "first" has no semantic meaning, consider it to be random. If the
+/// selection group is smaller than `max`, the entire group is returned.
```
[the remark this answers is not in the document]
rcgoodfellow: We are very much in all other things being equal just pick something territory here, as the bestpath algorithm has narrowed things down to a set it considers truly equal. But yeah, the ordering should be deterministic. You can find the `Ord` trait implementation for `Path` in `rdb/src/types.rs`.

interop-lab/create-containers.sh
rcgoodfellow: I think if we keep bespoke containers around, not within containerlab, I'd like to use terraform to set things up.

rdb/src/db.rs
rcgoodfellow: For now yeah. Eventually, we'll gain a cisco-style `maximum-paths` parameter.

rdb/src/bestpath.rs
```diff
@@ -0,0 +1,194 @@
...
+    };
+
+    // Partition the choice space on whether routes are shutdown or not. If we
+    // only have shutdown routes then use those. Otherwise use active routes
+    let (active, shutdown): (BTreeSet<&Path>, BTreeSet<&Path>) =
+        candidates.iter().partition(|x| x.shutdown);
+    let candidates = match (active.len(), shutdown.len()) {
+        (0, _) => shutdown,
+        (_, _) => active,
+    };
```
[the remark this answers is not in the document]
rcgoodfellow: This is clearer. Thanks. Updated.

rdb/src/bestpath.rs
rcgoodfellow: Added

mgadm/src/bgp.rs
```diff
@@ -5,78 +5,297 @@
...
+pub struct OmicronSubcommand {
+    #[command(subcommand)]
+    command: OmicronCmd,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum OmicronCmd {
+    /// Apply a BGP peer group configuration.
+    Apply { filename: String },
+}
```
internet-diglett, line 89: :thinking: 

Are peer groups specially related to Omicron in some way? Not a big deal, just curious why peer group configurations are under this command hierarchy and not one of the BGP neighbor related ones.

internet-diglett (approved):

mgadm/src/bgp.rs
rcgoodfellow: This is a bit more than a peer group in the typical sense, that comment is very misleading will fix. Thanks for pointing that out.

PR #239, opened 2024-05-02
[smf] Update manifests to reflect self assembling switch zone

karencfv:
**DO NOT MERGE** - let's wait until release 8 is out the door

In Omicron, we are converting all zones to be self assembling. This means we are no longer having to boot the zone, configure zone networking, configure all the services and finally refresh each service.

For the switch zone we have two new services that configure common networking in https://github.com/oxidecomputer/omicron/pull/5593. This PR adds these dependencies to the SMF files that require them, and sets the service default instance as enabled by default.

This PR should be merged in conjunction with https://github.com/oxidecomputer/omicron/pull/5593

### DIFF
--- a/smf/ddm/manifest.xml
+++ b/smf/ddm/manifest.xml
@@ -4,10 +4,18 @@
 <service_bundle type='manifest' name='mg-ddm'>
 
 <service name='oxide/mg-ddm' type='service' version='1'>
-  <create_default_instance enabled='false' />
+  <create_default_instance enabled='true' />
   <dependency name='network' grouping='require_all' restart_on='none'
     type='service'>
   <service_fmri value='svc:/milestone/network:default' />
+  </dependency>
+
+  <dependency name='zone_network_setup' grouping='require_all' restart_on='none' type='service'>
+    <service_fmri value='svc:/oxide/zone-network-setup:default' />
+  </dependency>
+
+  <dependency name='switch_zone_setup' grouping='require_all' restart_on='none' type='service'>
+    <service_fmri value='svc:/oxide/switch_zone_setup:default' />
   </dependency>
 
   <exec_method type='method' name='start'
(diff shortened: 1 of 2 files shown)


karencfv:
```
| + ./build.sh
| Unable to generate SMF dependency on svc:/oxide/zone-network-setup:default declared in /wo [...]
| Unable to generate SMF dependency on svc:/oxide/switch_zone_setup:default declared in /wor [...]
```

Oof, this is unfortunate. Other repos seem to not have a problem with those services living in omicron :thinking: From the contents of the build.sh file I can't really tell what's going on. @jclulow, is this a buildomat thing? Do you have a clue about what's going on?

karencfv:
Ok, after looking into this a bit further, this issue appears to stem from https://github.com/oxidecomputer/pkg5. 

I'm guessing we probably don't want to change any functionality there directly :thinking: @citrus-it do you have any suggestions on how to best resolve this issue?

jclulow:
If it's IPS (pkg) saying that, it's probably sad that the things you're depending on are not delivered in a package themselves, so it cannot create an appropriate dependency.  It's possible to have IPS just ignore certain files from a dependency resolution perspective, I believe.

jclulow:
Looking at [the manual](https://man.omnios.org/pkgdepend):

> Delivered SMF service manifests that include **require_all** dependencies result in dependencies on the packages that deliver the SMF manifests that provide those FMRIs.

If you look in that manual, I believe the `pkg.depend.bypass-generate` directive would allow you to just ignore those manifests.

It's possible, if the package is also supposed to work _outside_ the switch zone, that the right thing to do is actually change them from being **require_all** dependencies to being something more relaxed about missing services.  The manual page for [smf(7)](https://illumos.org/man/7/smf) describes the depenency grouping types.  If you  look at the descriptions there it's possible that **optional_all** might be more appropriate?  I'm not sure!

karencfv:
Thanks @jclulow! Looks like I was looking at the wrong build.sh file :smile: `.github/buildomat/jobs/build.sh` instead of `pkg/build.sh` :woman_facepalming: 

> If you look in that manual, I believe the pkg.depend.bypass-generate directive would allow you to just ignore those manifests.

hmmm, that's interesting! That could be a good workaround. I'll try it out.

PR #257, opened 2024-05-20
Update oximeter dep, bump collection interval

bnaecker:


bnaecker:
I updated the collection interval here from 1s to 10s. It looks like each collection triggers the generation of the stats, so we're collecting every second to get stats every second. I'd probably recommend that we decouple these: collect less often, but generate stats on whatever interval we'd like them. Or we can just roll that bit back, it's not a huge deal right now.

rcgoodfellow:
I think I'd prefer to keep collection at 1 second. A lot of the phenomena an operator may be interested in observing in real time happens inside of 10 seconds. For example a common BGP configuration uses keepalive timers of 1 second and hold timers of 3 seconds.

bnaecker:
Reset to original interval in b1303ab

rcgoodfellow (approved):

PR #260, opened 2024-05-24
Add support for parsing vlan-tagged tfports

Nieuwejaar:


mg-lower/src/dendrite.rs
```diff
@@ -64,6 +64,8 @@ impl RouteHash {
         path: Path,
     ) -> Result<RouteHash, Error> {
         let (port_id, link_id) = get_port_and_link(path.nexthop)?;
+
+        // XXX: verify that the link vlan ID matches the route's vlan ID?
```
rcgoodfellow, line 68: I'm not sure if this is strictly a requirement. One could imagine a deployment where a BGP control plane is on one VLAN and the data plane is on a different VLAN.

rcgoodfellow (approved):
LGTM, but I think we can drop the XXX comment.

PR #266, opened 2024-06-05
migrate net/prefix types to oxnet

ahl:
I've hit at least two interesting problem (as distinct from the countless uninteresting problems I've encountered).

1. This v1 stuff? It seems that we've baked in an old ddm somewhere to ensure compatibility... compatibility which I've broken by changing the json schema for IpNet/Prefix. I guess this means we're heading to v3, but I don't understand this mechanism. **Update** addressed by #276 and opened #275.

3. Dependencies, how do they work? `mgd` depends transitively on `libnet` which ends up dumping in a bunch of dependencies on illumos-only shared libraries. It builds on Linux... somehow? It's surprising (baffling) to me how that works. Or worked. So I put in some proper feature flags to help make sure this actually works as intended.


rcgoodfellow:
> Dependencies, how do they work? mgd depends transitively on libnet which ends up dumping in a bunch of dependencies on illumos-only shared libraries. It builds on Linux... somehow? It's surprising (baffling) to me how that works. Or worked. So I put in some proper feature flags to help make sure this actually works as intended.

libnet should not be pulling in much in the way of shared library dependencies. It interacts with illumos using ioctl and doors interfaces. I think the only illumos library that has snuck in there is the `kstat`, and I'm not sure if any of the kstat-based stuff is used in maghemite.

ahl:
@rcgoodfellow this failure led me to the feature: https://buildomat.eng.oxide.computer/wg/0/details/01HZN7DPHGCA4NTWA88AY1GYTG/K6UXLFfvHdWloZLAjDs8BKluNUnVgGx6uh4sGUn9CiTCsI2z/01HZN7EFMT4RNCBZX20PAZGAS6

In particular if you scroll to the bottom of this Linux build:

```
  = note: /usr/bin/ld: cannot find -lkstat: No such file or directory
          /usr/bin/ld: cannot find -lnvpair: No such file or directory
          /usr/bin/ld: cannot find -lzfs: No such file or directory
          /usr/bin/ld: cannot find -ldlpi: No such file or directory
          /usr/bin/ld: cannot find -ldoor: No such file or directory
          collect2: error: ld returned 1 exit status
```

I think there was some dead-code analysis that cause these **not** to be included that somehow I disrupted.

mgd/Cargo.toml
```diff
@@ -32,3 +32,4 @@ smf.workspace = true
 
 [features]
 default = ["mg-lower"]
+mg-lower = ["dep:mg-lower"]
```
rcgoodfellow, line 35: Is this just to pick up the post rust 1.60 explicit dependency syntax - or is there something more to what's going on here?

rcgoodfellow (approved):
Thanks Adam! LGTM just a few small questions.

ddm-admin-client/build.rs
```diff
@@ -1,3 +0,0 @@
-fn main() {
-    println!("cargo:rerun-if-changed=../openapi/ddm-admin.json");
```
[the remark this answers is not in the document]
ahl: To my knowledge it has been for quite a while. Did you find this not to be the case? https://github.com/oxidecomputer/progenitor/pull/16

mgd/Cargo.toml
ahl: I thought this was the (new) right way to specify the inclusion of optional dependencies.

rcgoodfellow:
Ah right - beyond `kstat`, those other deps are being picked up by other crates like `rusty-doors` and `dlpi-sys`. The whole compiling on Linux and Mac thing has been in service of supporting control plane build and test environments that require compiling and running somewhat vacuous maghemite daemon builds as placeholders that can answer API requests but don't really do much.

PR #283, opened 2024-06-28
compat between ddm exchange protocol v2/v3

rcgoodfellow:
This PR provides compatibility between DDM exchange protocol versions 2 and 3. When the transition to `oxnet` data structures took place, some members of data structures in the DDM exchange API were changed. In particular, members representing IPv4 and IPv6 prefixes. Because the `oxnet` data structures (de)serialize differently than the bespoke data structures that were there before, an incompatibility across DDM versions was introduced. 

This PR brings back the basic IP prefix data structures that were in use before and provides both V2 and V3 exchange API methods. Like prior cross-compatibility between V2 and V1, the version of a peer is determined during discovery, and then the exchange protocol implementation will use the appropriate method versions and deserialization mechanisms based on the discovered version of the peer.

### DIFF
--- a/.github/buildomat/jobs/test-ddm-trio-v2-server.sh
+++ b/.github/buildomat/jobs/test-ddm-trio-v2-server.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #:
-#: name = "test-ddm-trio-v1-server"
+#: name = "test-ddm-trio-v2-server"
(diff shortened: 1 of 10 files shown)


FelixMcFelix (approved):
I think the type transforms look correct, and the test changes seem reasonable to me.

PR #292, opened 2024-07-10
apply changes to peers on `bgp_apply`

### LINKED ISSUE #291: The `bgp_apply` endpoint does not sync changes to peers
The `bgp_apply` endpoint adds and removes peers as they are added and removed from the Omicron control plane. However, when changes are made to peers in the control plane, they are not synchronized by the `bgp_apply` endpoint in Maghemite.

rcgoodfellow:
- Fixes #291
- Fixes deadlock in `ExportPolicyChanged` handler in the `established` state

### DIFF
--- a/bgp/src/session.rs
+++ b/bgp/src/session.rs
@@ -1456,6 +1456,7 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
                         .filter(|x| list.contains(&Prefix::from(*x)))
                         .collect(),
                 };
+                drop(session);
 
                 let to_withdraw: BTreeSet<&Prefix4> =
                     originated_before.difference(&originated_after).collect();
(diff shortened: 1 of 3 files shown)


bgp/src/session.rs
```diff
@@ -1456,6 +1456,7 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
                         .filter(|x| list.contains(&Prefix::from(*x)))
                         .collect(),
                 };
+                drop(session);
```
FelixMcFelix, line 1459: I spent some time trying to understand this, I assume the deadlock is due to us attempting a re-entrant hold on `self.session` when we `self.send_update(...)`?

Are the extra drops in `update_session_info` a necessary part of this fix, or are they just cleaning up lock scope?

FelixMcFelix (approved):
Looks good! I have one non-blocking question to make sure I've understood the deadlock fix.

bgp/src/session.rs
rcgoodfellow: Yeah that's correct. The eager drop in the update function is just tidying things up a bit. That lock was blocked on the reentrant deadlock in the pstack I was looking at, so was just trying to scope things as tightly as possible while debugging.

PR #298, opened 2024-07-16
Move DDM timeseries to new TOML definitions

bnaecker:
- Update `oximeter` dependency and any related
- Swap out hand-rolled structs for the ones generated from the new TOML definitions
- Fixup Dropshot error type handling

### DIFF
--- a/ddm/src/oxstats.rs
+++ b/ddm/src/oxstats.rs
@@ -5,15 +5,11 @@
 use crate::{admin::RouterStats, sm::SmContext};
 use chrono::{DateTime, Utc};
 use dropshot::{ConfigLogging, ConfigLoggingLevel};
-use mg_common::{
-    counter,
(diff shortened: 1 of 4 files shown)


ddm/src/oxstats.rs
```diff
@@ -32,36 +48,6 @@ pub(crate) struct Stats {
...
-}
-
-#[derive(Debug, Clone, Target)]
-struct DdmRouter {
-    hostname: String,
-    rack_id: Uuid,
-    sled_id: Uuid,
-}
-
-counter!(SolicitationsSent);
```
rcgoodfellow, line 50: I presume the idea is to do the same thing from `mgd`, and then we can get rid of these macros altogether?

rcgoodfellow (approved):
LGTM

ddm/src/oxstats.rs
bnaecker: Yessir!

PR #301, opened 2024-07-18
Use toml for mgd timeseries

bnaecker:


rcgoodfellow:
With this I think we can get rid of `mg_common::{counter, quantity};` completely?

bnaecker:
Done in 0fefde4

rcgoodfellow (approved):

PR #309, opened 2024-07-24
remove prefixes when nexthops have been exhausted

### LINKED ISSUE #308: Tunnel routes flapping on BGP session shutdown
When a BGP peer shuts down, and that peer is providing the only path to a given prefix, that prefix should be withdrawn from tunnel routes advertised on the rack underlay network.

Testing that I'm currently doing is showing that this is not working as expected. Instead of withdraws, we see a withdraw and then and advertise around a second later.

```
20:20:29.238Z INFO slog-rs: withdraw tunnel: {
        TunnelOrigin {
            overlay_prefix: V4(
                Ipv4Net {
                    addr: 0.0.0.0,
                    width: 0,
                },
            ),
            boundary_addr: fd21:ab92:6d64:93df::1,
            vni: 99,
            metric: 18446744073709551615,
        },
    }
```
```
20:20:30.436Z INFO slog-rs: advertise tunnel: {
        TunnelOrigin {
            overlay_prefix: V4(
                Ipv4Net {
                    addr: 0.0.0.0,
                    width: 0,
                },
            ),
            boundary_addr: fd21:ab92:6d64:93df::1,
            vni: 99,
            metric: 18446744073709551615,
        },
    }
```

Running `mgadm bgp status imported` shows there are no imported routes. So we should not be re-announcing here.

Interestingly, we see this for the static RIB.

```
$ mgadm static get-v4-routes
{
    "0.0.0.0/0": [],
}
```

which is problematic .... I'm guessing what is happening here is that the BGP daemon is removing the nexthop from the RIB with the session shuts down, but leaves the prefix with no nexthops and that is confusing mg-lower in terms of synchronizing tunnel routes to DDM in the sense that we have the prefix in the RIB, but there are no nexthops.

### LINKED ISSUE #310: bgp: changing peer vlan-id does not change entries in RIB
When a BGP peer config is changed to have a different VLAN id, the entries that correspond to that peer in the RIB are not changed. Changing the VLAN ID should result in a session reset, or maybe a session refresh, but a reset seems prudent for the moment.

rcgoodfellow:
- Fixes #308
- Fixes #310

### DIFF
--- a/rdb/src/db.rs
+++ b/rdb/src/db.rs
@@ -554,6 +554,8 @@ impl Db {
             None => true,
         });
 
+        rib.retain(|&_, paths| !paths.is_empty());
+
         Self::update_loc_rib(&rib, &mut lock!(self.rib_loc), prefix);
         self.notify(prefix.into());
     }
@@ -574,6 +576,8 @@ impl Db {
             result.insert(*prefix, paths.clone());
             pcn.changed.insert(*prefix);
         }
+
+        rib.retain(|&_, paths| !paths.is_empty());
 
         for prefix in pcn.changed.iter() {
             Self::update_loc_rib(&rib, &mut lock!(self.rib_loc), *prefix);
--- a/bgp/src/session.rs
+++ b/bgp/src/session.rs
@@ -2354,6 +2354,11 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
             refresh_needed = true;
         }
 
+        if current.vlan_id != info.vlan_id {
+            current.vlan_id = info.vlan_id;
+            reset_needed = true;
+        }
+
         if current.allow_export != info.allow_export {
             let previous = current.allow_export.clone();
             current.allow_export = info.allow_export;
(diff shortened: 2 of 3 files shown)


rdb/src/db.rs
```diff
@@ -575,6 +577,8 @@ impl Db {
             pcn.changed.insert(*prefix);
         }
 
+        rib.retain(|&_, paths| !paths.is_empty());
```
taspelund, line 580: We already loop over `rib.iter_mut()` earlier in this function, so this will mean we effectively walk all prefixes in the RIB twice.

I think for simplicity and correctness sake this is fine for now, but we'll want to revisit this at a later date to optimize so this doesn't become a performance/convergence issue at high scale (like when receiving a full bgp feed).
e.g.
Only walk the rib once but consolidate conditional prefix removal into the path removal logic

taspelund (approved):

rdb/src/db.rs
rcgoodfellow: Let's get this captured in a follow-on issue.

PR #323, opened 2024-08-10
handle unknown capability codes gracefully

### LINKED ISSUE #322: BGP capability parsing should tolerate unknown capability codes better
When we receive a BGP open message with an unknown capability code we immediately send a notification to the peer and tear down the session. That's not terribly robust. We should instead chalk the capability up to an unknown capability (and not advertise it ourselves).

An example of how this can have unfortunate consequences is FRR 10.1 [started advertising the dynamic capability by default](https://www.frrouting.org/release/10.1/) and this prevented us from peering with 10.1 with default settings :/

rcgoodfellow:
- Fixes #322


elaine-oxide (approved):
LGTM, I see that the error in `code` is sent for handling.

PR #338, opened 2024-08-16
expose local_pref in static routing API

### LINKED ISSUE #262: Allow specification of custom route priorities
Specification of custom route priorities will let operators define preference among routes across routing protocols. For example, the operator may say that BGP routes have a priority of 100 (but could be modified on a peer-by-peer or prefix-by-prefix basis based on policy), static routes 50, but certain static routes have a priority of 200. This would allow for static routes to take precedence generally (lowest priority wins at 50), BGP would be the second choice at 100, and then the set of static routes with priority 200 would be fallbacks if nothing else is viable.

rcgoodfellow:
- Fixes #262

### DIFF
--- a/mgadm/src/static_routing.rs
+++ b/mgadm/src/static_routing.rs
@@ -34,6 +34,7 @@ pub struct StaticRoute4 {
     pub destination: Ipv4Net,
     pub nexthop: Ipv4Addr,
     pub vlan_id: Option<u16>,
+    pub local_pref: Option<u32>,
 }
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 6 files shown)


taspelund (approved):
Code-wise, this LGTM.  However, I'm not a fan of re-using the `local_pref` nomenclature for static routing, since the name conflicts with the BGP attribute.
I think an operator could easily intuit a RIB comparison between the local pref of a static route and the local pref attribute of a BGP route, rather than what's effectively a protocol priority (i.e. "admin distance" in Cisco-speak).

I don't think it's a blocker per se (so I'll still hit approve), but unless this is directly meant to be evaluated against one specific BGP attribute (local_pref) I think we should consider naming this differently.

rcgoodfellow:
Sounds good on cleaning up terminology. I've logged #343 to track this.

Issue #318, opened 2024-08-07
Add API endpoint for showing exported BGP prefixes


rcgoodfellow:
The `mgd` BGP API provides a way to originate a set of prefixes and also to limit what originated prefixes are exported on a peer-by-peer basis. It would be good to have an API endpoint that returns a per-peer list of exported prefixes given the origin set and per-peer filters.

PR #342, opened 2024-08-20
BGP: Add API endpoint for showing exported BGP prefixes

taspelund:
Fixes: #318

### DIFF
--- a/mgd/src/bgp_admin.rs
+++ b/mgd/src/bgp_admin.rs
@@ -2,6 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+#![allow(clippy::type_complexity)]
 use crate::bgp_param as resource;
 use crate::{admin::HandlerContext, bgp_param::*, error::Error, register};
 use bgp::router::LoadPolicyError;
--- a/mgd/src/bgp_admin.rs
+++ b/mgd/src/bgp_admin.rs
@@ -5,6 +5,7 @@
 use crate::bgp_param as resource;
 use crate::{admin::HandlerContext, bgp_param::*, error::Error, register};
 use bgp::router::LoadPolicyError;
+use bgp::session::FsmStateKind;
 use bgp::{
     config::RouterConfig,
     connection::BgpConnection,
@@ -18,7 +19,7 @@ use dropshot::{
     HttpResponseUpdatedNoContent, Query, RequestContext, TypedBody,
 };
 use http::status::StatusCode;
-use rdb::{Asn, BgpRouterInfo};
+use rdb::{Asn, BgpRouterInfo, ImportExportPolicy, Prefix};
 use slog::info;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::hash::{Hash, Hasher};
@@ -110,6 +111,7 @@ pub(crate) fn api_description(api: &mut ApiDescription<Arc<HandlerContext>>) {
     //
 
     register!(api, get_neighbors);
+    register!(api, get_exported);
     register!(api, get_imported);
     register!(api, get_selected);
     register!(api, message_history);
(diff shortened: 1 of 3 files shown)


mgd/src/bgp_admin.rs
```diff
@@ -351,6 +353,54 @@ pub async fn delete_origin4(
...
+) -> Result<HttpResponseOk<HashMap<IpAddr, Vec<Prefix>>>, HttpError> {
+    let rq = request.into_inner();
+    let ctx = ctx.context();
+    let mut exported = HashMap::new();
+
+    let r = get_router!(ctx, rq.asn)?.clone();
+    let orig4 = r.db.get_origin4();
+    let neighs = r.db.get_bgp_neighbors();
+    if orig4.is_err() || neighs.is_err() {
+        // not configuring peers or announce-sets is not an error
```
rcgoodfellow, line 369: These should only return an error if something goes wrong. They should return empty vectors if there is nothing there.

I'd recommend something like this

```rust
let orig4 = r.db.get_origin4().map_err(|e| {
    HttpError::for_internal_error(format!("error getting origin: {e}"))
})?;
let neighs = r.db.get_bgp_neighbors().map_err(|e| {
    HttpError::for_internal_error(format!("error getting neighbors: {e}"))
})?;
```

And that way we can avoid the unwraps below.

mgd/src/bgp_admin.rs
```diff
@@ -351,6 +353,54 @@ pub async fn delete_origin4(
...
+        return Ok(HttpResponseOk(exported));
+    }
+    let orig4 = orig4.unwrap();
+
+    for n in neighs.unwrap() {
+        let sess = r.get_session(n.host.ip());
+
+        if sess.is_none() || sess.unwrap().state() != FsmStateKind::Established
+        {
+            break;
```
rcgoodfellow, line 379: I think we'd want to `continue` instead of breaking here.

mgd/src/bgp_admin.rs
```diff
@@ -351,6 +353,54 @@ pub async fn delete_origin4(
...
+    if orig4.is_err() || neighs.is_err() {
+        // not configuring peers or announce-sets is not an error
+        return Ok(HttpResponseOk(exported));
+    }
+    let orig4 = orig4.unwrap();
+
+    for n in neighs.unwrap() {
+        let sess = r.get_session(n.host.ip());
+
+        if sess.is_none() || sess.unwrap().state() != FsmStateKind::Established
```
rcgoodfellow, line 377: This `unwrap` can be avoided by filtering the `get_session` result to established sessions.

```rust
if r.get_session(n.host.ip())
    .filter(|s| s.state() == FsmStateKind::Established)
    .is_none()
{
    continue;
}
```

rcgoodfellow (commented):
Thanks Trey! Just a few comments below.

mgd/src/bgp_admin.rs
taspelund: In some iteration, I had basically this same code but I ended up convincing myself it was indicative of missing config rather than a failure of the getter functions.  I'll fix it, thanks!

mgd/src/bgp_admin.rs
taspelund: Yep, that was definitely a brain fart there.  `continue` is what I wanted, but not what I typed.
Will address this too

mgd/src/bgp_admin.rs
taspelund: That's nice and clean, I like this a lot.  Thanks for the suggestion, I'll get this integrated!

taspelund:
Comments have been resolved.  The only item I wasn't sure about was that clippy is complaining about a complex type being returned by this function.  Should I create something similar to `resource::Origin4` to address the complaint?

rcgoodfellow:
Let's add a `#![allow(clippy::type_complexity)]` at the top of the file for now, as I'm not sure how a type alias is going to interact with dropshot/progenitor.

rcgoodfellow (approved):

PR #359, opened 2024-09-09
RIB Priority Changes

taspelund:
This primarily makes 5 changes:

1. Move BGP's `local_pref` into `BgpPathProperties` (this is a BGP-specific attribute, let's store it with the other BGP-specific attributes)
2. Rename the RIB's `local_pref` to `rib_priority` (since this is used by the RIB to prioritize certain protocols/paths over others)
3. Fixes up tests to use the new nomenclature/structure
4. Introduces new constants to represent default RIB priorities for Static and BGP protocols
5. Enables RIB priority comparison in the RIB's bestpath selection algorithm

### DIFF
--- a/bgp/src/session.rs
+++ b/bgp/src/session.rs
@@ -16,8 +16,8 @@ use crate::policy::{CheckerResult, ShaperResult};
 use crate::router::Router;
 use crate::{dbg, err, inf, to_canonical, trc, wrn};
 use mg_common::{lock, read_lock, write_lock};
-pub use rdb::DEFAULT_ROUTE_PRIORITY;
 use rdb::{Asn, BgpPathProperties, Db, ImportExportPolicy, Prefix, Prefix4};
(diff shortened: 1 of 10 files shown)


bgp/src/session.rs
```diff
@@ -2070,13 +2070,14 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
                 let path = rdb::Path {
                     nexthop: nexthop.into(),
                     shutdown: update.graceful_shutdown(),
-                    local_pref: update.local_pref(),
+                    rib_priority: DEFAULT_RIB_PRIORITY_BGP,
```
rcgoodfellow, line 2073: Can we create a follow on issue to make this configurable? It seems to me that we'd want this to be configurable at the router level. Perhaps it could be useful as a per-session option. But for the purposes of this PR, a static value is fine.

rcgoodfellow (approved):
Just one comment on `rib_priority` configurability for BGP learned routes. Otherwise LGTM.

PR #364, opened 2024-09-12
Migrate the custom ddm client to hyper v1

ahl:
I'm working on moving all of the control plane to hyper v1... albeit slowly. I think this is a nibble that will advance the general effort.


rcgoodfellow (approved):
LGTM

Issue #369, opened 2024-09-17
Static route key retained despite deleting all paths


taspelund:
Building from `main`, I see this when manually testing adding/removing static routes:
```
➜  maghemite git:(f5dbb6a) ./target/debug/mgadm static add-v4-route 1.1.1.1/32 100.64.0.0
➜  maghemite git:(f5dbb6a) ./target/debug/mgadm static add-v4-route 1.1.1.1/32 100.64.0.1
➜  maghemite git:(f5dbb6a) ./target/debug/mgadm static get-v4-routes
{
    "1.1.1.1/32": [
        Path {
            bgp: None,
            local_pref: None,
            nexthop: 100.64.0.0,
            shutdown: false,
            vlan_id: None,
        },
        Path {
            bgp: None,
            local_pref: None,
            nexthop: 100.64.0.1,
            shutdown: false,
            vlan_id: None,
        },
    ],
}
➜  maghemite git:(f5dbb6a) ./target/debug/mgadm static remove-v4-routes 1.1.1.1/32 100.64.0.0
➜  maghemite git:(f5dbb6a) ./target/debug/mgadm static get-v4-routes
{
    "1.1.1.1/32": [
        Path {
            bgp: None,
            local_pref: None,
            nexthop: 100.64.0.1,
            shutdown: false,
            vlan_id: None,
        },
    ],
}
➜  maghemite git:(f5dbb6a) ./target/debug/mgadm static remove-v4-routes 1.1.1.1/32 100.64.0.1
➜  maghemite git:(f5dbb6a) ./target/debug/mgadm static get-v4-routes
{
    "1.1.1.1/32": [],
}
```

I don't think we should be retaining 1.1.1.1/32 after deletion of the last path.

Issue #330, opened 2024-08-13
When Export Policy is configured, withdrawn NLRI can only ever contain routes allowed by Export Policy


taspelund:
```
   1886         if let ImportExportPolicy::Allow(ref policy) =
   1887             lock!(self.session).allow_export
   1888         {
   1889             let message_policy = policy
   1890                 .iter()
   1891                 .filter_map(|x| match x {
   1892                     rdb::Prefix::V4(x) => Some(x),
   1893                     _ => None,
   1894                 })
   1895                 .map(|x| crate::messages::Prefix::from(*x))
   1896                 .collect::<BTreeSet<crate::messages::Prefix>>();
   1897
   1898             update.nlri.retain(|x| message_policy.contains(x));
   1899             update.withdrawn.retain(|x| message_policy.contains(x));
   1900         };
```
bgp/src/session.rs

Line 1899 is only retaining routes in the withdrawn nlri list that pass the export policy.
This means that reconfiguring the export policy will not result in previously-allowed routes getting withdrawn.

Issue #241, opened 2024-05-02
bgp: router id insufficient discriminator for route-refresh stale marking


rcgoodfellow:
This is a follow up for this review comment

- https://github.com/oxidecomputer/maghemite/pull/199/files#r1585208765

We need to mark routes as stale for route refresh based on session, not on router id, as we may have multiple sessions with a single router and thus single router id.

taspelund:
This discriminator should be something specific to the BGP session, since that should have a 1:1 mapping to a configured peer.
I did an initial code-dive to see what value might make sense to use in our BGP implementation, and it seems like `Cnx: BgpConnection` is the likely candidate. A tuple of local/remote sockaddrs might also be an option here, but that just seems like peeling one layer of abstraction off of the `BgpConnection` trait without an immediately obvious upside.

PR #385, opened 2024-10-03
Embiggen the goodness of maghemite

taspelund:
This addresses a few different items found in Maghemite.

1) remove_prefix_path() did not delete the prefix key from `rib_in` after the last path was removed.
Fixes: #369

2) remove_prefix_path() was removing all paths for a prefix that happened to share a nexthop.  This is bad because a single call to `static_remove_v4_route()` would delete all paths with that nexthop even when fields like rib_priority didn't match. This would likely also allow the deletion of a floating static route to inadvertently clobber a BGP route via the same next-hop.

3) apply_update() and send_update() were incorrectly applying the relevant ImportExportPolicy to the list of withdrawn routes.  Filtering route withdrawals causes stale routing information, as we can end up ignoring a legitimate withdrawal from a peer or we could forget to inform a peer that we're no longer originating a route (e.g. if we first set an export policy matching N routes in an announce-set, then removed a route from that announce-set).  In either case, this is incorrect behavior that can lead to forwarding issues.
Fixes: #330


taspelund:
Before:
```
➜  maghemite git:(f5dbb6a) ./target/debug/mgadm static get-v4-routes
{
    "1.1.1.1/32": [
        Path {
            bgp: None,
            local_pref: None,
            nexthop: 100.64.0.1,
            shutdown: false,
            vlan_id: None,
        },
    ],
}
➜  maghemite git:(f5dbb6a) ./target/debug/mgadm static remove-v4-routes 1.1.1.1/32 100.64.0.1
➜  maghemite git:(f5dbb6a) ./target/debug/mgadm static get-v4-routes
{
    "1.1.1.1/32": [],
}
```

After:
```
➜  maghemite git:(trey/cleanup_static_route_key) ./target/debug/mgadm static get-v4-routes
{
    "1.1.1.1/32": [
        Path {
            bgp: None,
            nexthop: 100.64.0.0,
            rib_priority: 2,
            shutdown: false,
            vlan_id: None,
        },
        Path {
            bgp: None,
            nexthop: 100.64.0.0,
            rib_priority: 10,
            shutdown: false,
            vlan_id: None,
        },
    ],
}
➜  maghemite git:(trey/cleanup_static_route_key) ✗ ./target/debug/mgadm static remove-v4-routes 1.1.1.1/32 100.64.0.0
➜  maghemite git:(trey/cleanup_static_route_key) ✗ ./target/debug/mgadm static get-v4-routes
{}
```

zeeshanlakhani (commented):
This seems fine to me, but can we add a test demonstrating the example you showed for removing the prefix from the rib?

taspelund:
I added two more commits + rebased off main.

The second commit adds the requested tests, plus additional unit testing for rdb::Db methods.

The first commit is a refactor of various rdb::Db methods and their callers.  Highlights below:
- Rename of various add/remove methods + helper functions, ideally helping to keep naming more consistent (e.g. using "peer" instead of "id", adding "bgp"/"static" to method names)
- In the add/remove codepaths, there is now a single PrefixChangeNotification used to batch all pcn's together instead of firing them off individually. I tried to keep the updated routes together in a collection up until the point of processing them, so it will be more clear in the future where to make changes to support bulk prefix updates.
- Delete codepath now accepts a closure that determines what paths are removed
- removed an unused method `get_imported()` and consolidated the `enable_nexthop()`/`disable_nexthop()` methods into a single method accepting a bool: `set_nexthop_shutdown()`.
- Changed return types from Result -> void, since callers were not actually checking the Result anyway
- `loc_rib()` now returns a `Rib`, which aligns with `full_rib()`, `static_rib()`, and `bgp_rib()`
- Moved BGP attribute -> Path parsing out of a for loop, since these are per-update not per-route
- Guarded an illumos-specific debug import
- Added some TODOs into the comments, primarily for #241 and to check that we cover all cases for nexthop parsing
- Introduced `get_selected_prefix_paths()` to get an entry from rib_loc

Other refactoring we might consider later:
- Moving test helpers into a common spot (e.g. `wait_for_eq` macros in bgp
- Moving logging helpers into a common spot (e.g. all of bgp/src/log.rs)
- Cleaning up remaining instances of `.lock().unwrap()` by moving to `lock!()`
- More ergonomic conversions between messages::Prefix and rdb::Prefix types/subtypes (e.g. Prefix4)

rdb/src/db.rs
```diff
@@ -378,8 +379,13 @@ impl Db {
         }
     }
 
-    pub fn get_imported(&self) -> Rib {
-        lock!(self.rib_in).clone()
+    pub fn get_selected_prefix_paths(&self, prefix: &Prefix) -> Vec<Path> {
```
rcgoodfellow, line 382: If this function is only used for tests, it should probably have a ` #[cfg(test)]` attribute.

rdb/src/db.rs
```diff
@@ -474,118 +502,108 @@ impl Db {
...
                     let mut replacement = p.clone();
-                    replacement.shutdown = true;
+                    replacement.shutdown = shutdown;
                     paths.insert(replacement);
                     pcn.changed.insert(*prefix);
                 }
             }
         }
 
+        //TODO loc_rib updater as a pcn listener?
```
rcgoodfellow, line 519: It'd be nice to keep PCNs directional to the extent possible. Reasoning about things could become a bit difficult if the DB were producing and consuming PCNs.

rdb/src/db.rs
```diff
@@ -474,118 +502,108 @@ impl Db {
...
+                    "{}: remove from tree STATIC4_ROUTES failed: {e}",
+                    route.prefix
+                );
+                continue;
+            }
         }
 
-        Self::update_loc_rib(&rib, &mut lock!(self.rib_loc), prefix);
-        self.notify(prefix.into());
+        self.notify(pcn);
```
rcgoodfellow, line 574: I think we may want to flush before sending out the PCN.

rdb/src/db.rs
```diff
@@ -474,118 +502,108 @@ impl Db {
...
+            pcn.changed.insert(prefix);
         }
         self.notify(pcn);
-        result
+    }
+
+    // helper function to remove all routes learned from a given peer
+    // e.g. when peer is deleted or exits Established state
+    pub fn remove_bgp_peer_prefixes(&self, id: u32) {
+        // TODO: don't rely on BGP ID for path definition.
```
rcgoodfellow, line 601: Let's post an issue for this.

rdb/src/db.rs
```diff
@@ -617,8 +635,13 @@ impl Db {
         Ok(())
     }
 
-    pub fn mark_bgp_id_stale(&self, id: u32) {
-        let mut rib = self.rib_loc.lock().unwrap();
+    pub fn mark_bgp_peer_stale(&self, id: u32) {
+        // TODO: don't rely on BGP ID for path definition.
```
rcgoodfellow, line 639: I think this is #241; if so, let's leave a breadcrumb here.

rcgoodfellow (commented):
Thanks Trey! Some great improvements here. Just a few comments.

rdb/src/db.rs
taspelund: Today it's only used in tests, but it seems like this would be a useful function for exposing per-prefix details up through mgadm/nexus.  I can add the attribute if you'd like, but if so then I'd probably want to add a TODO comment along with it.

bgp/src/session.rs
```diff
@@ -2039,6 +2042,37 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
         };
 
         if !update.nlri.is_empty() {
+            // TODO: parse and prefer nexthop in MP_REACH_NLRI
```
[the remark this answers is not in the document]
taspelund: https://github.com/oxidecomputer/maghemite/issues/397

rdb/src/db.rs
taspelund: This TODO comment was present before my changes.  I didn't know if it was valid or not, so I just retained it through my refactoring.  Do you want me to delete the comment? or file an issue about it?

rdb/src/db.rs
taspelund: I moved the flush after the notification because an error can cause an early return.  My thought was that a successful update to the loc_rib should trigger a PCN even if the flush of the persistent route DB fails later, since the "running state" is what other listeners are likely intending to react to.

You can correct me if my read on the situation is wrong, or if we need to give some more thought to robust handling here.

rdb/src/db.rs
taspelund: Yep, that's the one.  I'll update the comment with a pointer to the issue.

rdb/src/db.rs
taspelund: This is also #241.  Will copy the breadcrumb to all the TODOs related to this.

rdb/src/db.rs
rcgoodfellow: Let's just remove it.

rdb/src/db.rs
rcgoodfellow: The thing I worry about here is sending out a PCN that causes a persistent DB read and the data not being flushed yet. I think there is a deeper issue here, as well as in several other methods that modify in-memory state and then write the result to disk (that were also there before this PR). We need transactional semantics for these methods, either everything succeeds or nothing does. If it's possible to run all the fallible commands that persist state up front and then run the soft state updates that don't have error conditions, that would be ideal.

taspelund: Probably not an important detail in this context, but it doesn't look like PCNs trigger a read of the persistent DB.
mg-lower seems to be the only watcher and it reacts to PCNs by kicking off `sync_prefix()`, which looks at running state known in dpd (ASIC) or the loc_rib.

All that seems appropriate to me, as I don't think we want mg-lower paying attention to rib_in or state changes that don't make their way to the set of best paths.  And since the persistent bits here are specific to the static route DB, I also think it makes sense to only update the running state if the persistent change succeeds.

I do agree we should have transactional semantics (especially when the config and running state overlap so closely), but I'm not sure how to go about reverting a failed operation, or what an effective way to propagate errors for a collection of operands (routes) would be.
e.g.
If flush() fails after making a file smaller (removing a route), would we expect the revert action (presumably re-adding the route back into the DB and calling flush() again) to succeed?  What happens if the subsequent flush() to get us into the last known good state also fails?
The API also allows for a collection of static routes to come in through this delete path.  What happens if an error happens for just a subset of those routes? Should we bail out upon first error (which I think we did prior to this PR), stop and try to revert the operations we've already done, or try to complete the operations for as many routes as we can (aggregating or transforming the errors we've gotten up until this point)?

For the revert action, I think it might behoove us to understand what guarantees flush() makes when it returns an Err.  Do we have any guarantees which prior writes have/haven't made it to disk?  If not, then I'm not sure what other approach to take beyond calling flush() again... which feels a bit like the definition of insanity (something something... expecting a different outcome)

rdb/src/db.rs
taspelund: `get_prefix_paths()` is used in the same way today, except it's being imported by `#[test]` functions in bgp/src/test.rs... and `mod test;` in bgp/src/lib.rs is also guarded by `#[cfg(test)]`.

For whatever reason, clippy refuses to accept the import of `get_prefix_paths()` in bgp/src/test.rs while its definition is guarded with `#[cfg(test)]`.

I think that's probably something that could be addressed while doing cleanup of the existing test code.  If you feel that's important to address in this PR, I can do that.  Otherwise I can follow up on that later.

rdb/src/db.rs
rcgoodfellow: I think we can frontload the fallible DB stuff and leverage sled transactions. Like diff below. Also happy to make this a follow-on issue.

```diff
diff --git a/rdb/src/db.rs b/rdb/src/db.rs
index 6d4e612..ccef6b7 100644
--- a/rdb/src/db.rs
+++ b/rdb/src/db.rs
@@ -541,37 +541,32 @@ impl Db {
         &self,
         routes: Vec<StaticRouteKey>,
     ) -> Result<(), Error> {
-        let mut pcn = PrefixChangeNotification::default();
         let tree = self.persistent.open_tree(STATIC4_ROUTES)?;
+
+        let mut route_keys = Vec::new();
+        for route in &routes {
+            let key = serde_json::to_string(route)?;
+            route_keys.push(key);
+        }
+
+        tree.transaction(|tx_db| {
+            for key in &route_keys {
+                tx_db.remove(key.as_str())?;
+            }
+            Ok(())
+        })?;
+        tree.flush()?;
+
+        let mut pcn = PrefixChangeNotification::default();
+
         for route in routes {
             self.remove_prefix_path(route.prefix, |rib_path: &Path| {
                 rib_path.cmp(&Path::from(route)) == CmpOrdering::Equal
             });
             pcn.changed.insert(route.prefix);
-
-            let key = match serde_json::to_string(&route) {
-                Err(e) => {
-                    error!(
-                        self.log,
-                        "{} serde_json::to_string conversion failed: {e}",
-                        route.prefix
-                    );
-                    continue;
-                }
-                Ok(s) => s,
-            };
-            if let Err(e) = tree.remove(key.as_str()) {
-                error!(
-                    self.log,
-                    "{}: remove from tree STATIC4_ROUTES failed: {e}",
-                    route.prefix
-                );
-                continue;
-            }
         }

         self.notify(pcn);
-        tree.flush()?;
         Ok(())
     }

diff --git a/rdb/src/error.rs b/rdb/src/error.rs
index 666811a..6df5472 100644
--- a/rdb/src/error.rs
+++ b/rdb/src/error.rs
@@ -7,6 +7,9 @@ pub enum Error {
     #[error("datastore error {0}")]
     DataStore(#[from] sled::Error),

+    #[error("data store transaction {0}")]
+    DataStoreTransaction(#[from] sled::transaction::TransactionError),
+
     #[error("serialization error {0}")]
     Serialization(#[from] serde_json::Error),
```

rdb/src/db.rs
rcgoodfellow: Following up on this later is fine.

rcgoodfellow (approved):

PR #391, opened 2024-10-16
guard against excessive sa keepalive spawning

### LINKED ISSUE #390: SA keepalive spawn guard not initialized
When spawning SA keepalive threads, there is a guard in place to prevent the thread from being spawned multiple times, but that guard is never initialized.

https://github.com/oxidecomputer/maghemite/blob/056283eb02b6887fbf27f66a215662520f7c159c/bgp/src/connection_tcp.rs#L666-L688

rcgoodfellow:
- fixes #390

### DIFF
--- a/bgp/src/connection_tcp.rs
+++ b/bgp/src/connection_tcp.rs
@@ -12,7 +12,7 @@ use crate::session::FsmEvent;
 use crate::to_canonical;
 use libc::{c_int, sockaddr_storage};
 use mg_common::lock;
-use slog::{error, info, trace, warn, Logger};
+use slog::{debug, error, info, trace, warn, Logger};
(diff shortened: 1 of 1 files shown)


taspelund (commented):
How often will this timer fire? Do we want an info log to be written for this every time it fires, or would this be better as a debug?

rcgoodfellow:
This code will only run when we're actively trying to set up a connection with a peer. I've made the logging a bit less aggressive.

taspelund:
If it's a (relatively) one shot operation, then I'm not worried about the logging. I just didn't want to spam the logs if this was going to be run consistently on a timer

taspelund (approved):
Sorry for the delay in the detailed review.  LGTM

PR #398, opened 2024-10-25
Stop using Router-ID to distinguish between BGP sessions

taspelund:
BGP-ID is insufficient to distinguish between BGP sessions, as it's perfectly valid/reasonable to have multiple sessions to the same peer. Or an operator can (incorrectly or otherwise) assign the same BGP-ID to multiple routers.
This ensures that Session-A to Peer-A going down does not impact Session-B to Peer-A.

Fixes: #241


rdb/src/types.rs
```diff
@@ -24,6 +24,12 @@ pub struct Path {
     pub vlan_id: Option<u16>,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Eq, PartialEq)]
+pub struct SocketAddrPair {
```
zeeshanlakhani, line 28: minor: (since we control the struct), prob worth impl'ing a new function that automatically sets local & peer (also handling the option).

rdb/src/types.rs
```diff
@@ -61,6 +67,8 @@ impl From<StaticRouteKey> for Path {
 pub struct BgpPathProperties {
     pub origin_as: u32,
     pub id: u32,
+    // pub conn: (Option<SocketAddr>, SocketAddr),
```
zeeshanlakhani, line 70: remove comment?

taspelund: Done. Thanks!

rdb/src/types.rs
taspelund: Done!

taspelund:
I'd be interested in some feedback on whether or not it would be worth destructuring the Option wrapping the local sockaddr.
e.g.
```
➜  maghemite git:(trey/unique_neigh) ✗ rustfmt $(git status | grep modified | grep -v 'json\|Cargo.lock\|Cargo.toml' | awk '{print $2}' | tr '\n' diff --git a/rdb/src/types.rs b/rdb/src/types.rs
index 7c1211c..db3d0a4 100644
--- a/rdb/src/types.rs
+++ b/rdb/src/types.rs
@@ -26,13 +26,19 @@ pub struct Path {

 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Eq, PartialEq)]
 pub struct SocketAddrPair {
-    pub local: Option<SocketAddr>,
+    pub local: SocketAddr,
     pub peer: SocketAddr,
 }

 impl SocketAddrPair {
     pub fn new(local: Option<SocketAddr>, peer: SocketAddr) -> Self {
-        Self { local, peer }
+        Self {
+            local: local.unwrap_or(SocketAddr::new(
+                IpAddr::from_str("0.0.0.0").unwrap(),
+                0,
+            )),
+            peer,
+        }
     }
 }
```

I'm not sure if this is a reasonable approach or if there's a more idiomatic way to do this.
The `TcpStream` type has a local_addr() method to query the kernel for the local SockAddr, but that's already used by BgpConnectionTcp to populate that field + store the return as an Option.

I'm just not sure if there's a better way to handle this, or if leaving it as an Option makes the most sense.

internet-diglett:
> I'd be interested in some feedback on whether or not it would be worth destructuring the Option wrapping the local sockaddr. e.g.
> 
> ```
> ➜  maghemite git:(trey/unique_neigh) ✗ rustfmt $(git status | grep modified | grep -v 'json\|Cargo.lock\|Cargo.toml' | awk '{print $2}' | tr '\n' diff --git a/rdb/src/types.rs b/rdb/src/types.rs
> index 7c1211c..db3d0a4 100644
> --- a/rdb/src/types.rs
> +++ b/rdb/src/types.rs
> @@ -26,13 +26,19 @@ pub struct Path {
> 
>  #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Eq, PartialEq)]
>  pub struct SocketAddrPair {
> -    pub local: Option<SocketAddr>,
> +    pub local: SocketAddr,
>      pub peer: SocketAddr,
>  }
> 
>  impl SocketAddrPair {
>      pub fn new(local: Option<SocketAddr>, peer: SocketAddr) -> Self {
> -        Self { local, peer }
> +        Self {
> +            local: local.unwrap_or(SocketAddr::new(
> +                IpAddr::from_str("0.0.0.0").unwrap(),
> +                0,
> +            )),
> +            peer,
> +        }
>      }
>  }
> ```
> 
> I'm not sure if this is a reasonable approach or if there's a more idiomatic way to do this. The `TcpStream` type has a local_addr() method to query the kernel for the local SockAddr, but that's already used by BgpConnectionTcp to populate that field + store the return as an Option.
> 
> I'm just not sure if there's a better way to handle this, or if leaving it as an Option makes the most sense.

This could make sense. In what situations do we expect the `local SocketAddr` to be `None`?

taspelund:
>This could make sense. In what situations do we expect the local SocketAddr to be None?

It looks like local_addr() is the top of a bunch of layers wrapping libc `getsockname()`:
https://stdrs.dev/nightly/x86_64-pc-windows-gnu/src/std/sys_common/net.rs.html#309-311

So if the illumos [man page](https://illumos.org/man/3SOCKET/getsockname) is to be trusted on this:
```
ERRORS

       The call succeeds unless:

       EBADF
                   The argument s is not a valid file descriptor.


       ENOMEM
                   There was insufficient memory available for the operation to
                   complete.


       ENOSR
                   There were insufficient STREAMS resources available for the
                   operation to complete.


       ENOTSOCK
                   The argument s is not a socket.
```
It seems like the only time we'd realistically expect this to fail maghemite for reasons within our control is if we call local_addr() for an invalid fd/socket... which I'd guess would only happen if the connect() call fails or if the socket gets closed out from under us?

From the API handler side, mgd checks whether or not there's a valid `SessionRunner<Cnx: BgpConnection>` for the peer we're attempting to delete before grabbing the local/peer sockaddr:
```
    pub(crate) async fn remove_neighbor(
        ctx: Arc<HandlerContext>,
        asn: u32,
        addr: IpAddr,
    ) -> Result<HttpResponseDeleted, Error> {
        info!(ctx.log, "remove neighbor: {}", addr);

        let session = get_router!(ctx, asn)?
            .get_session(addr)
            .ok_or(Error::NotFound("session for bgp peer not found".into()))?;
        let conn =
            SocketAddrPair::new(session.bind_addr, session.neighbor.host);
        ctx.db.remove_bgp_peer_prefixes(conn);
        ctx.db.remove_bgp_neighbor(addr)?;
        get_router!(&ctx, asn)?.delete_session(addr);

        Ok(HttpResponseDeleted())
    }
```

It looks like the only time mgd calls into the kernel to query the local sockaddr is during Update generation -- to figure out what next-hop to set on routes we originate.

I don't imagine we need to do this more than once (store the local sockaddr after a successful `connect()` call), so long as we properly handle the socket dying underneath us.  I'm not sure if/how we handle those scenarios today (maybe we just need to check for an err every time we interact with the socket/`BgpConnection`?) but maybe that's an indication I need to get a better understanding of our socket/error handling.

rcgoodfellow (approved):

taspelund:
@rcgoodfellow apparently I just needed a week or so to let the thoughts percolate, but I think you were right about using the peer IP as the distinguisher...

The biggest benefit I see to this is that the IP a stateless object (it's either configured/present or it's not) so we won't have to think about socket state changing. No error handling if we can't query the socket state from the kernel. No changes to the hash input, possibly leaving us unable to cleanup entries inserted with the old socket state (mainly the ephemeral port).

I've updated the code to use an IpAddr instead of a SocketAddrPair, which I think has simplified things quite a bit.
I've also cleaned up logging in the rdb tests.

Another review over the latest changes would be appreciated.

taspelund:
One other comment I'll leave regarding the session discriminator: if we ever decide to support peering to the same address multiple times (e.g. via explicit ipv6 link-local addresses), this will likely need to be updated again to include a "scope" in addition to the peer address

taspelund:
Output from manual testing in a4x2 against Arista peers (datacenter topology) below.

The only change to the default a4x2 configuration that was made is to add the following on cr2:
```
router bgp 51
router-id 198.51.102.1  # give cr2 the same router-id cr1
```
On Arista, all sessions are automatically reset after issuing this -- no need for `clear ip bgp *` -- so all sessions come back up with the new RID in place.
Now both cr1 and cr2 are using RID 198.51.102.1.

To trigger the session going down, I am running the following on cr1:
```
router bgp 50
neighbor 198.51.101.2 shutdown
```

Before these changes (maghemite built from main):
```
root@oxz_switch:~# mgadm bgp status neighbors 47
Peer Address  Peer ASN  State        State Duration  Hold   Keepalive
198.51.101.9  Some(51)  Established  3m 55s 710ms    6s/6s  2s/2s
198.51.101.1  Some(50)  Idle         891ms           6s/6s  2s/2s
root@oxz_switch:~# mgadm bgp status imported 47
root@oxz_switch:~#
```
We can see that even though only the session to cr1 went down (IP 198.51.101.1), the imported paths from cr2 (IP 198.51.101.9) have been removed.


After these changes (maghemite built from `da9305638385cb0fd28e4a48055e87b7942e3e8c`):
```
root@oxz_switch:~# mgadm bgp status neighbors 47
Peer Address  Peer ASN  State        State Duration  Hold   Keepalive
198.51.101.9  Some(51)  Established  2m 30s 764ms    6s/6s  2s/2s
198.51.101.1  Some(50)  Idle         2s 890ms        6s/6s  2s/2s
root@oxz_switch:~# mgadm bgp status imported 47
BGP Routes
=============
Prefix     Nexthop       RIB Priority  Local Pref  Origin AS  Peer ID       MED   AS Path   Stale
0.0.0.0/0  198.51.101.9  20            None        51         198.51.102.1  None  [51, 52]  None
root@oxz_switch:~#
```
The paths learned from cr2 (IP 198.51.101.9, also using RID 198.51.102.1) have not been removed from the RIB.

taspelund:
Similar broken behavior before these changes:  if one of multiple sessions to a single RID flaps, routes learned from all other sessions are lost indefinitely (presumably unless all peers were to flap at the same time).
I only swapped out my build of maghemite on g0, but on g3 I see this behavior upon running `clear ip bgp *` on cr1:
```
root@oxz_switch:~# mgadm bgp status neighbors 48
Peer Address   Peer ASN  State        State Duration  Hold   Keepalive
198.51.101.13  Some(51)  Established  31m 51s 407ms   6s/6s  2s/2s
198.51.101.5   Some(50)  Established  5s 253ms        6s/6s  2s/2s
root@oxz_switch:~# mgadm bgp status imported 48
BGP Routes
=============
Prefix     Nexthop       RIB Priority  Local Pref  Origin AS  Peer ID       MED   AS Path   Stale
0.0.0.0/0  198.51.101.5  20            None        50         198.51.102.1  None  [50, 52]  None
```

taspelund:
@rcgoodfellow confirmed in chat that we're good to merge.  Going forward with that now

PR #402, opened 2024-10-30
[metrics] add switch info/meta to ddmd/mgd related timeseries

zeeshanlakhani:
Add additional switch and sled metadata to bfd/bgp/routing/ddm timeseries. 

Related:

- https://github.com/oxidecomputer/dendrite/pull/1033
- https://github.com/oxidecomputer/omicron/pull/6955

### DIFF
--- a/mgd/src/oxstats.rs
+++ b/mgd/src/oxstats.rs
@@ -1,15 +1,16 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
-
 use crate::admin::HandlerContext;
 use crate::bfd_admin::BfdContext;
 use crate::bgp_admin::BgpContext;
 use chrono::{DateTime, Utc};
+use dpd_client::types;
 use mg_common::lock;
 use mg_common::nexus::{local_underlay_address, run_oximeter};
 use mg_common::stats::MgLowerStats;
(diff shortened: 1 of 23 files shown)


zeeshanlakhani:
Will update cargo.lock post review.

mgd/src/oxstats.rs
```diff
@@ -95,21 +98,50 @@ pub(crate) struct Stats {
...
+                    .into(),
+                asic_lot: $switch_idents
+                    .lot
+                    .clone()
+                    .map(|c| c.to_string())
+                    .unwrap_or_else(|| $switch_idents.asic_backend.to_string())
+                    .into(),
+                asic_wafer: $switch_idents.wafer.unwrap_or(0),
+                asic_wafer_loc_x: $switch_idents
+                    .wafer_loc
```
taspelund: Maybe we can extract all of this logic into a single macro that is called by all the other `_counter` macros?
Unless I'm reading it wrong, it seems like we're repeating the same logic in each macro

mgd/src/oxstats.rs
```diff
@@ -95,21 +98,50 @@ pub(crate) struct Stats {
...
+                    .clone()
+                    .map(|c| c.to_string())
+                    .unwrap_or_else(|| $switch_idents.asic_backend.to_string())
+                    .into(),
+                asic_lot: $switch_idents
+                    .lot
+                    .clone()
+                    .map(|c| c.to_string())
+                    .unwrap_or_else(|| $switch_idents.asic_backend.to_string())
+                    .into(),
```
taspelund: Is it intentional that both fab and lot default to a backend string instead of a different sentinel value?

Issue #405, opened 2024-11-12
BGP: all NLRI in an Update are dropped when a single prefix is invalid


taspelund:
If mgd receives an Update message with an invalid prefix, all NLRI in that Update are ignored rather than just the offending prefixes.

tcpdump output below shows routes sent to mgd from an instance of FRR configured to send 2 locally originated routes (`192.0.2.0/24` and `224.0.0.0/24`) and one route from an eBGP peer (`100.64.1.0/24`).

Since the route FRR learned from an eBGP peer has different path attributes (AS-Path differs), the routes are bin-packed into two separate Update messages:
1) `100.64.1.0/24`
2) `192.0.2.0/24` + `224.0.0.0/24`

tcpdump showing the Updates sent by FRR:
```
22:30:18.886973 a8:40:25:f4:4d:ab > a8:40:25:ff:77:77, ethertype IPv4 (0x0800), length 177: (tos 0xc0, ttl 1, id 45643, offset 0, flags [DF], proto TCP (6), length 163)
    172.30.0.13.179 > 172.30.0.15.59944: Flags [P.], cksum 0x8867 (correct), seq 119:230, ack 90, win 509, options [nop,nop,TS val 1224249282 ecr 2431360110], length 111: BGP
	Update Message (2), length: 52
	  Origin (1), length: 1, Flags [T]: IGP
	    0x0000:  00
	  AS Path (2), length: 10, Flags [TE]: 65002 65001
	    0x0000:  0202 0000 fdea 0000 fde9
	  Next Hop (3), length: 4, Flags [T]: 172.30.0.13
	    0x0000:  ac1e 000d
	  Updated routes:
	    100.64.1.0/24
	Update Message (2), length: 59
	  Origin (1), length: 1, Flags [T]: IGP
	    0x0000:  00
	  AS Path (2), length: 6, Flags [TE]: 65002
	    0x0000:  0201 0000 fdea
	  Next Hop (3), length: 4, Flags [T]: 172.30.0.13
	    0x0000:  ac1e 000d
	  Multi Exit Discriminator (4), length: 4, Flags [O]: 0
	    0x0000:  0000 0000
	  Updated routes:
	    192.0.2.0/24
	    224.0.0.0/24
```

This results in the `rib_in` containing only `100.64.1.0/24`:
```
root@debbie-mgd:~/git/maghemite# ./target/debug/mgadm bgp status imported 47
BGP Routes
=============
[...]

PR #406, opened 2024-11-12
bgp: valid prefixes aren't guilty by association

taspelund:
apply_update() would return early if a single invalid nlri was detected. This results in all nlri in that Update being discarded, even if there were no issues with those other nlri or with the Update itself. Let's not throw the baby out with the bathwater.

Fixes: #405

### DIFF
--- a/bgp/src/session.rs
+++ b/bgp/src/session.rs
@@ -2124,6 +2124,7 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
                     .nlri
                     .iter()
                     .filter(|p| !originated.contains(&p.as_prefix4()))
+                    .filter(|p| self.sanity_check_v4_prefix(&p.as_prefix4()))
                     .map(|n| rdb::Prefix::from(n.as_prefix4()))
                     .collect(),
                 path.clone(),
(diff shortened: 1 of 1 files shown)


bgp/src/session.rs
```diff
@@ -2183,18 +2183,13 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
-        for prefix in &update.nlri {
-            if prefix.length == 0 {
-                continue;
-            }
-            let first = prefix.value[0];
-            // check 127.0.0.0/8, 224.0.0.0/4
-            if (first == 127) || (first & 0xf0 == 224) {
-                return Err(Error::InvalidNlriPrefix(prefix.as_prefix4()));
-            }
+    fn sanity_check_v4_prefix(&self, prefix: &Prefix4) -> bool {
```
rcgoodfellow, line 2186: maybe we can give this a more descriptive name, like `filter_martians`.

rcgoodfellow (commented):
Thanks! Just one small comment on improving the name. I'd like to avoid the use of the word sanity in this way.

rcgoodfellow (approved):

PR #433, opened 2025-02-07
[ddm-admin] Remove now-unused parameters when starting Oximeter server

jgallagher:
Small housekeeping PR: I was doing some unrelated work in sled-agent and noticed these two parameters are no longer needed to start the Oximeter server, so I think we can safely drop them. This should be backwards-compatible, since an old client sending these params is fine (we'll just ignore them), and we only receive these, never send them.

### DIFF
--- a/ddm/src/admin.rs
+++ b/ddm/src/admin.rs
@@ -347,8 +347,6 @@ async fn sync(
 
 #[derive(Clone, Debug, Serialize, Deserialize, JsonSchema, PartialEq)]
 pub struct EnableStatsRequest {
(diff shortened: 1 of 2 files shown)


rcgoodfellow:
I think this will need a corresponding update in Omicron to use the new API when enabling stats?

jgallagher:
Yes, the PR in which omicron pulls in the new OpenAPI spec will need a small tweak to sled-agent to stop passing these parameters.

rcgoodfellow (approved):

PR #436, opened 2025-02-15
update omicron and dendrite

sunshowers:
Drop the oxnet git dependency.


Cargo.toml
```diff
@@ -109,6 +109,3 @@ package = "dpd-client"
 git = "https://github.com/oxidecomputer/dendrite"
 branch = "main"
 package = "common"
-
```
ahl, line 112: this is bonkers.

ahl (approved):

PR #458, opened 2025-04-02
Bump dendrite

Nieuwejaar:
Use the exported `types::PortId` rather than dendite's internal `common::ports::PortId`.
Update the dendrite commit used for testing, and adapt to the new dendrite/smf interaction.


rcgoodfellow (approved):
Thanks! Beat me to the punch on coralling the SMF issues.

PR #466, opened 2025-04-10
Add buildomat job for interop testing

taspelund:


.github/buildomat/jobs/test-interop.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+#:
+#: name = "test-interop"
+#: variety = "basic"
+#: target = "helios-2.0"
```
rcgoodfellow, line 5: This needs to be something like `lab-2.0-opte' to get a physical lab machine, see [this](https://github.com/oxidecomputer/propolis/blob/master/.github/buildomat/jobs/phd-run.sh#L5) as an example.

.github/buildomat/jobs/test-interop.sh
```diff
@@ -0,0 +1,48 @@
...
+#:
+
+set -x
+set -e
+
+cargo --version
+rustc --version
+
+cargo install cargo-nextest
+pfexec pkg install protobuf git
```
jclulow, line 25: Note that there is no **pkg** and no **git** in the **lab-2.0-\*** target environment.  Generally I would recommend that you build anything that you need in a regular **helios-2.0** job, put the needed bits and pieces in an archive as an output rule, and then unpack those things in this job.  We copy any outputs from a job you depend on into `/input` for this job; some more detail is under the documentation for the **dependencies** configuration property [in the buildomat README](https://github.com/oxidecomputer/buildomat?tab=readme-ov-file#configuration).

jclulow:
I pushed a few changes that hopefully inch a little closer to success.  I'm not sure if **cargo-nextest** is a requirement, or if you could just run the test binaries directly?  If you can do the latter, you might look at this older version of what **propolis** used to do for testing: https://github.com/oxidecomputer/propolis/tree/fc09089e2d8a238637449a59752bf134f52631bb/.github/buildomat/jobs

The **tests-build.sh** and **test-run.sh** jobs have long since been replaced with something else, I believe, but show a basic way to build a bunch of tests with cargo and then ship them and run them on another system.

If **cargo-nextest** is actually necessary for some reason, you might want to look at the "archive" facility it has: https://nexte.st/docs/ci-features/archiving/

taspelund:
The latest item I found is that `baseline.rs` in the interop crate was compiling an absolute filepath to the .falcon directory into the binary, within `$CARGO_MANIFEST_DIR`.

I submitted a PR to the testbed repo to address this: https://github.com/oxidecomputer/testbed/pull/130
However, if that isn't an acceptable/useful change then I can update the buildomat scripts to make sure we build and execute the test binary from the same filepath to make this a non-issue here

taspelund:
@jclulow is there a way for me to manually kick off the entire CI suite again?  I'd like to re-run it for this PR a handful of times before taking it out of draft status

jclulow:
I believe you can click around in the interface and find a re-run button

taspelund:
@rcgoodfellow @jclulow I think we've got things at a pretty stable point... CI passed the last 3 times since moving to the new gimlet buildomat target + disabling ssh host validation.

When you get some time, would you guys mind reviewing this and seeing if you're both satisfied with it?

Issue #464, opened 2025-04-09
mgadm should sort routes numerically


taspelund:
When mgadm returns a RIB, the prefixes are sorted lexicographically, not numerically.
e.g.
```
root@oxz_switch0:~# mgadm bgp status imported 47
Static Routes
=============
Prefix     Nexthop       RIB Priority
0.0.0.0/0  198.51.103.1  1
0.0.0.0/0  198.51.103.9  1
BGP Routes
=============
Prefix        Nexthop       RIB Priority  Local Pref  Origin AS  Peer ID       MED   AS Path  Stale
0.0.0.0/0     198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.1.0/24    198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.10.0/24   198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.100.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.101.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.102.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.103.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.104.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.105.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.106.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.107.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.108.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.109.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.11.0/24   198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.110.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.111.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
2.0.112.0/24  198.51.103.9  20            Some(20)    52         198.51.102.5  None  [52]     None
[...]

PR #484, opened 2025-05-06
mgadm: perform natural sorting of prefixes

taspelund:
mgadm: perform natural sorting of prefixes

before:
```
Static Routes
=============
Prefix          Nexthop  RIB Priority
100.64.0.0/24   4.4.4.4  1
                4.4.4.5  1
100.64.1.0/24   4.4.4.5  1
100.64.10.0/24  4.4.4.5  1
100.64.2.0/24   4.4.4.5  1
100.64.3.0/24   4.4.4.5  1
```

after:
```
Static Routes
=============
Prefix          Nexthop  RIB Priority
100.64.0.0/24   4.4.4.4  1
                4.4.4.5  1
100.64.1.0/24   4.4.4.5  1
100.64.2.0/24   4.4.4.5  1
100.64.3.0/24   4.4.4.5  1
100.64.10.0/24  4.4.4.5  1
```

Fixes: #464

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2796,6 +2796,7 @@ dependencies = [
  "humantime",
  "mg-admin-client 0.1.0",
  "mg-common",
+ "natord",
  "rdb",
  "serde_json",
  "slog",
@@ -2887,6 +2888,12 @@ dependencies = [
  "spin 0.9.8",
  "version_check",
 ]
+
+[[package]]
+name = "natord"
+version = "1.0.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "308d96db8debc727c3fd9744aac51751243420e46edf401010908da7f8d5e57c"
 
 [[package]]
 name = "newline-converter"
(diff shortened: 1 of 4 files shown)


rcgoodfellow (approved):
May want to parse the prefix keys and preform an ordered sort over the integer representation of the subnet then the prefix length, as it's not clear to me that natord will always produce the expected result. But this is definitely an improvement.

taspelund:
I was worried it would be a bit expensive to convert to/from strings a bunch of times.  mgd already converts `BTreeMap<Prefix, _>` to `BTreeMap<String, _>` before shipping the response out the door, and now we'd be essentially converting each String back to Prefix for sorting purposes.

I originally thought the "right" way to do this would be to stop converting from Prefix -> String in mgd and make mgadm aware of the `Prefix` type, but I ran into some issues where it compiled but progenitor didn't seem to update anything in the openapi spec to indicate that the BTreeMap uses a different type for its keys.

Typing this out makes me think I should spend more time understanding why the dropshot API didn't seem to update the way I expected

PR #491, opened 2025-05-15
Fix #490

### LINKED ISSUE #490: BGP: removing an entire peer group through the apply endpoint does not remove peers within group.
In the code below, we compute what peers have been removed from a peer group and delete them. However, if an entire peer group has been removed, nothing is done with the peers within that group.

https://github.com/oxidecomputer/maghemite/blob/18b418ea20fd6451fa2984428088b0cbc34856c3/mgd/src/bgp_admin.rs#L521-L549

rcgoodfellow:
Fixes #490

- Test for bug in fa95b6b1d5f7cde049f4b52d393b7b570dea0f39
- Bug fixed in 9b2fc7bc22b45bf8f68fbe363e056944c0223ec2


rcgoodfellow:
The bug is captured here. This test should pass but is failing due to the code identified in #490 

https://github.com/oxidecomputer/maghemite/blob/aeddf21f5614d776d2d3684ebbd167028f3e3b4c/mgd/src/bgp_admin.rs#L1093-L1171

mgd/src/bgp_admin.rs
```diff
@@ -495,8 +495,14 @@ pub async fn bgp_apply(
     ctx: RequestContext<Arc<HandlerContext>>,
     request: TypedBody<ApplyRequest>,
 ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
-    let log = ctx.context().log.clone();
-    let rq = request.into_inner();
+    do_bgp_apply(ctx.context(), request.into_inner()).await
```
rcgoodfellow, line 498: This is just moving the implementation of bgp_apply into a function that can be tested without having to construct a RequestContext.

mgd/src/bgp_admin.rs
```diff
@@ -1065,3 +1087,131 @@ pub(crate) mod helpers {
...
+        do_bgp_apply(&ctx, req.clone())
+            .await
+            .expect("bgp apply request");
+
+        assert_eq!(
+            ctx.db.get_bgp_neighbors().expect("get bgp neighbors").len(),
+            2,
+        );
+
+        peers.clear();
```
mkeeter, line 1205: Nit: I don't think this line does anything

rcgoodfellow: Ah yeah, that was leftover from a previous version of the test.

rcgoodfellow: Fixed in 4a069cfd0cfb635b3d68276fae038919e6ec290c

mgd/src/bgp_admin.rs
```diff
@@ -1065,3 +1087,130 @@ pub(crate) mod helpers {
...
+            data_dir: tmpdir.to_owned(),
+            db: Db::new(dbdir.as_str(), log.clone()).unwrap(),
+            mg_lower_stats: Arc::new(MgLowerStats::default()),
+            stats_server_running: Mutex::new(false),
+            oximeter_port: 0,
+        });
+
+        let mut peers = HashMap::new();
+        peers.insert(
+            String::from("qsfp0"),
```
internet-diglett, line 1140: Will we ever group peers by some criteria other than interface name?

rcgoodfellow: I believe the maghemite perspective on this is fairly agnostic. From omicron's perspective it's interface name, but it could be something else eventually like "cloud-peers", "isp-peers", etc. Currently omicron uses groups by interface name to ease interface level management iirc.

internet-diglett (approved):

Issue #517, opened 2025-06-30
test-flake: interop test failing intermittently


rcgoodfellow:
A run on main recently failed. Seemingly due to key management issues.

https://buildomat.eng.oxide.computer/wg/0/details/01JYZZVZ0RKM01H89Q5DXVJPQ3/ueDC1lnTVOaHCanTP8ZCwCkEM69O3AKdXkU4cQk9HR9I9djY/01JYZZW91GN7V9C22Y3R2HP8NG

rcgoodfellow:
I think we can just close this for now and if it shows up again we'll reopen.

PR #518, opened 2025-06-30
buildomat: disable key validation for scp

taspelund:
This is already done for ssh, so let's make sure it works for scp too.


taspelund:
@rcgoodfellow I believe this will address the flake you're seeing in #517

PR #523, opened 2025-07-16
Use `oxide-tokio-rt` rather than `#[tokio::main]`

hawkw:
As described in [RFD 579], we are now recommending certain non-default Tokio runtime configurations in all software we deploy in production. The [`oxide-tokio-rt`] crate provides a way to easily get these common recommended configurations by using its functions to initialize the runtime rather than `#[tokio::main]`

Presently, `oxide-tokio-rt` does the following:

- Enables DTrace probes for Tokio runtime events using [`tokio-dtrace`]
- Disables the [LIFO slot optimization][lifo], which was the root cause of omicron#8334

But, we anticipate that if there are additional Tokio configurations which we want to recommend in the future, we'll add them here as well.

This commit updates `maghemite` to initialize the Tokio runtime using `oxide-tokio-rt`. I've added a Clippy config for warning on uses of `#[tokio::main]`, as well, to avoid forgetting to use `oxide-tokio-rt` if new binaries are added. All the configs we set currently in `oxide-tokio-rt` requires the `tokio_unstable` RUSTFLAGS cfg, but it looks like that's already being set, so that shouldn't be a problem.

I've changed every binary to use `oxide-tokio-rt` with the exception of `mg-package` --- it didn't seem necessary to use the production-like config there and maybe avoiding an additional dependency to fetch would make it build slightly faster, although I'm unconvinced. I did change the lab binaries to use `oxide-tokio-rt`, as I figured that the DTrace probes could come in handy there.

[RFD 579]: https://rfd.shared.oxide.computer/rfd/0579 
[`oxide-tokio-rt`]: https://github.com/oxidecomputer/oxide-tokio-rt 
[`tokio-dtrace`]: https://github.com/oxidecomputer/tokio-dtrace
[lifo]:
    https://rfd.shared.oxide.computer/rfd/0579#_disabling_the_lifo_slot_optimization

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -605,7 +605,7 @@ version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fde0e0ec90c9dfb3b4b1a0891a7dcd0e2bffde2f7efed5fe7c9bb00e5bfb915e"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -932,6 +932,7 @@ dependencies = [
  "ddm",
  "ddm-admin-client",
  "mg-common",
+ "oxide-tokio-rt",
  "oxnet",
  "slog",
  "slog-async",
@@ -953,6 +954,7 @@ dependencies = [
  "hostname 0.3.1",
  "libnet",
  "mg-common",
+ "oxide-tokio-rt",
  "slog",
  "slog-async",
  "slog-bunyan",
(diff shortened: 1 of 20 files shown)


hawkw:
[CI failure](https://github.com/oxidecomputer/maghemite/pull/523/checks?check_run_id=46135084768) looks unrelated: [we timed out installing a package](https://buildomat.eng.oxide.computer/wg/0/details/01K0AN6HJ3GGDPX68YYJJ16QRM/aTZ4ySimz9AgH8YsZajxXg7eTFg3NhdIAc3393pdxNZZAG9t/01K0AN7FMKDA0YS0Q1FD659GPA#S1605). I'm restarting that one.

rcgoodfellow (approved):

Issue #400, opened 2024-10-25
BGP: make BESTPATH_FANOUT configurable


taspelund:
Currently BESTPATH_FANOUT is a const defined as 1.
This means that update_rib() will only ever copy a single path from rib_in -> loc_rib, and thus in practice the RIB will not use ECMP.

This ticket tracks the work to test/support the ability to make this configurable.

PR #524, opened 2025-07-21
Add bestpath fanout config capabilities

taspelund:
Adds an API endpoint for /bestpath/fanout and methods to get/set the fanout value.  This defines the maximum number of ECMP paths that can be selected by the RIB during bestpath calculation.

Fixes: #400

### DIFF
--- a/mgadm/src/bestpath.rs
+++ b/mgadm/src/bestpath.rs
@@ -1,0 +1,59 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::num::NonZeroU8;
+
+use anyhow::Result;
+use clap::{Args, Subcommand};
+use mg_admin_client::types;
+use mg_admin_client::Client;
+
+#[derive(Subcommand, Debug)]
+pub enum Commands {
+    /// Fanout configuration commands.
+    Fanout(FanoutCommand),
+}
+
+#[derive(Debug, Args)]
+pub struct FanoutCommand {
+    #[command(subcommand)]
+    command: FanoutCmd,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum FanoutCmd {
+    /// Read the current fanout setting.
+    Read,
+
+    /// Update the fanout setting.
+    Update {
+        /// Maximum number of equal-cost paths for ECMP forwarding
+        fanout: NonZeroU8,
+    },
+}
+
+pub async fn commands(command: Commands, c: Client) -> Result<()> {
+    match command {
+        Commands::Fanout(fanout_cmd) => match fanout_cmd.command {
+            FanoutCmd::Read => read_bestpath_fanout(c).await?,
+            FanoutCmd::Update { fanout } => {
+                update_bestpath_fanout(fanout, c).await?
+            }
+        },
+    }
+    Ok(())
+}
+
+async fn read_bestpath_fanout(c: Client) -> Result<()> {
+    let result = c.read_bestpath_fanout().await?;
+    println!("{}", result.into_inner().fanout);
+    Ok(())
+}
+
+async fn update_bestpath_fanout(fanout: NonZeroU8, c: Client) -> Result<()> {
+    c.update_bestpath_fanout(&types::BestpathFanoutRequest { fanout })
+        .await?;
+    println!("Updated bestpath fanout to: {}", fanout);
+    Ok(())
+}
--- a/mgd/src/bgp_param.rs
+++ b/mgd/src/bgp_param.rs
@@ -8,6 +8,7 @@ use rdb::{ImportExportPolicy, Path, PolicyAction, Prefix4};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeSet, HashMap};
+use std::num::NonZeroU8;
 use std::time::Duration;
 use std::{
     collections::BTreeMap,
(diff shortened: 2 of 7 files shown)


mgd/src/bgp_param.rs
```diff
@@ -215,6 +216,18 @@ pub struct GetOriginated4Request {
     pub asn: u32,
 }
 
+#[derive(Debug, Deserialize, Serialize, JsonSchema)]
+pub struct BestpathFanoutRequest {
+    /// Maximum number of equal-cost paths for ECMP forwarding
+    pub fanout: NonZeroU8,
```
rcgoodfellow, line 222: Same comment for this and the `NonZeroU8` from `mgadm` on practical limits for fanout.

rdb/src/db.rs
```diff
@@ -616,6 +629,31 @@ impl Db {
...
+        Ok(match NonZeroU8::new(fan) {
+            None => NonZeroU8::new(1).unwrap(),
+            Some(fanout) => fanout,
+        })
+    }
+
+    pub fn set_bestpath_fanout(&self, fanout: NonZeroU8) -> Result<(), Error> {
+        let tree = self.persistent.open_tree(SETTINGS)?;
+        let value = fanout.get().to_be_bytes();
+        tree.insert(BESTPATH_FANOUT_KEY, &value)?;
```
rcgoodfellow, line 653: This can be simplified to

```rust
tree.insert(BESTPATH_FANOUT_KEY, &[fanout.get()])?;
```

mgadm/src/bestpath.rs
```diff
@@ -0,0 +1,59 @@
...
+
+#[derive(Subcommand, Debug)]
+pub enum FanoutCmd {
+    /// Read the current fanout setting.
+    Read,
+
+    /// Update the fanout setting.
+    Update {
+        /// Maximum number of equal-cost paths for ECMP forwarding
+        fanout: NonZeroU8,
```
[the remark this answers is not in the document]
taspelund: My thought behind a u8 was that it would be a large enough value to give us a lot of headroom for a while. Unless our switches will be doing 32 uplinks * 4 VLANs per uplink, we wouldn't realistically hit the u8 max.

I agree that the Oxide system should impose limits based on what the switch can handle.
Does that limit belong in maghemite, dpd, or in Nexus?

I think it would make sense for a hard limit to live in dpd (just as a defensive practice) and for Nexus to restrict the upper bounds to dpd's limit at the API level.

This is not a hill I'm interested in dying on though, so if you feel strongly about having a check in maghemite as well then I'm happy to add one

rdb/src/db.rs
```diff
@@ -616,6 +629,31 @@ impl Db {
         tree.flush()?;
         Ok(())
     }
+    pub fn get_bestpath_fanout(&self) -> Result<NonZeroU8, Error> {
+        let tree = self.persistent.open_tree(SETTINGS)?;
+        let result = tree.get(BESTPATH_FANOUT_KEY)?;
+        let fan = match result {
+            Some(value) => {
+                u8::from_be_bytes((*value).try_into().map_err(|_| {
+                    Error::DbKey("invalid bestpath_fanout value in db".into())
```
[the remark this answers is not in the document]
taspelund: Will update in another commit

rdb/src/db.rs
taspelund: Sure, will clean this up

taspelund:
I realized during this last round of testing that the handler for the "update" API should trigger a re-run of bestpath.
Otherwise the RIB will not be updated with the new amount of allowed paths (regardless of whether the new value has increased or decreased.
e.g.
```
treyaspelund@Tallon-IV 07:45:24 PM | ~/git/maghemite  trey/bgp_rib_knobs
‣ ./target/debug/mgadm bestpath fanout read
2

treyaspelund@Tallon-IV 07:46:33 PM | ~/git/maghemite  trey/bgp_rib_knobs
‣ ./target/debug/mgadm bgp status selected 65100
Static Routes
=============
Prefix          Nexthop  RIB Priority
100.64.10.0/22  4.4.4.3  1
                4.4.4.4  1
                4.4.4.5  1

treyaspelund@Tallon-IV 07:46:35 PM | ~/git/maghemite  trey/bgp_rib_knobs
‣ ./target/debug/mgadm static add-v4-route 100.64.10.0/22 4.4.4.6

treyaspelund@Tallon-IV 07:46:38 PM | ~/git/maghemite  trey/bgp_rib_knobs
‣ ./target/debug/mgadm bgp status selected 65100
Static Routes
=============
Prefix          Nexthop  RIB Priority
100.64.10.0/22  4.4.4.3  1
                4.4.4.4  1

treyaspelund@Tallon-IV 07:46:40 PM | ~/git/maghemite  trey/bgp_rib_knobs
‣ ./target/debug/mgadm bestpath fanout update 4
Updated bestpath fanout to: 4

treyaspelund@Tallon-IV 07:47:00 PM | ~/git/maghemite  trey/bgp_rib_knobs
‣ ./target/debug/mgadm bgp status selected 65100
Static Routes
=============
Prefix          Nexthop  RIB Priority
100.64.10.0/22  4.4.4.3  1
                4.4.4.4  1
```

I'll write a method that walks the RIB and triggers a re-run of bestpath for each route, perhaps taking a closure to allow the caller to optimize which routes need to be kicked.

rdb/src/db.rs
```diff
@@ -617,6 +648,29 @@ impl Db {
         Ok(())
     }
 
+    pub fn get_bestpath_fanout(&self) -> Result<NonZeroU8, Error> {
+        let tree = self.persistent.open_tree(SETTINGS)?;
+        let fan = match tree.get(BESTPATH_FANOUT)? {
+            // fanout was not in db
+            None => DEFAULT_BESTPATH_FANOUT,
+            Some(value) => value[0],
```
rcgoodfellow, line 656: One of the properties of using 

```rust
let value: [u8; 1] = (*value).try_into().map_err(...)
```

here is that the attempt [will fail if there are too many bytes in the value](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d31320ac48f1db235079302cb2f3227b). If we just take the first element, we could be masking an error that leads to undefined program behavior. And if for some reason there is an empty value we'll crash.

mgadm/src/bestpath.rs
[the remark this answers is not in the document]
rcgoodfellow: Yeah, this particular instance is just the CLI which is not in the critical path so that's fine.

PR #526, opened 2025-07-22
Exit from bestpath early if there's only 1 path to evaluate

taspelund:


taspelund:
`test-interop` job is failing now because `jq` installed or in `$PATH` on juniper/arista VMs... not sure why this worked before but is broken now, but I'll look into it

taspelund:
I am able to reproduce the "missing `jq`" symptom when running interop on my workstation.
It would appear the issue is due to an issue going on with Debian's `bullseye-backports` repo causing it to return a 404 when apt attempts to look for its `Release` file: https://github.com/oxidecomputer/testbed/issues/158

rcgoodfellow (approved):
LGTM. Thanks @taspelund

taspelund:
Pretty sure this latest CI failure is because I updated the init.sh scripts for debian/arista nodes in a4x2 but forgot to update it for arista/juniper in interop.  Will submit a testbed PR to address it, which should let this pass.

Issue #529, opened 2025-08-03
host bits should be zeroed for RIB entries


taspelund:
Today, there is nothing stopping multiple routes w/ the same prefix but different host bits from being accepted into the RIB as separate entries.
```
treyaspelund@Tallon-IV 12:03:56 PM | ~/git/maghemite  main
‣ ./target/debug/mgadm static add-v4-route 1.1.1.1/24 4.4.4.4

treyaspelund@Tallon-IV 12:04:02 PM | ~/git/maghemite  main
‣ ./target/debug/mgadm static add-v4-route 1.1.1.0/24 4.4.4.4

treyaspelund@Tallon-IV 12:04:07 PM | ~/git/maghemite  main
‣ ./target/debug/mgadm static get-v4-routes
{
    "1.1.1.1/24": [
        Path {
            bgp: None,
            nexthop: 4.4.4.4,
            rib_priority: 1,
            shutdown: false,
            vlan_id: None,
        },
    ],
    "1.1.1.0/24": [
        Path {
            bgp: None,
            nexthop: 4.4.4.4,
            rib_priority: 1,
            shutdown: false,
            vlan_id: None,
        },
    ],
}
```

In the above example, both `1.1.1.1/24` and `1.1.1.0/24` map to the same prefix: `1.1.1.0/24`
However, there's nothing in place today to ensure these inputs reference the same prefix.

We should either zero out the host bits before prefix creation or we should enforce this requirement well before RIB insertion of the prefix (e.g. during BGP NLRI validation, in the mgd API handlers, and during sled::Db insertion/parsing).

The current behavior would undoubtedly be confusing for any users who end up in this state, as it's left up to the implementation details of mg-lower, dendrite, and tofino as to if/how these conflicts would be handled.

PR #533, opened 2025-08-08
Initial IPv6 Support

taspelund:
This PR adds initial support for IPv6 static routing in maghemite, with the very beginnings of BGP support (really just originated IPv6 prefixes).
This addresses a few bugs found along the way (e.g. #529), adds unit tests for old/new functionality, and does a good bit of code cleanup, e.g.
 - mostly standardizing on oxnet::IpNet in generic spots and rdb::Prefix in db-specific spots, rather than several unique types
 - moving common functionality into mg-common where it can be used by all other crates w/o creating circular dependencies

### DIFF
--- a/bgp/src/messages.rs
+++ b/bgp/src/messages.rs
@@ -740,15 +740,39 @@ impl Prefix {
             return false;
         }
 
+        // Check address family compatibility
+        let self_is_ipv4 = self.value.len() <= 4;
+        let x_is_ipv4 = x.value.len() <= 4;
+
+        // Cross-family comparisons should return false
+        if self_is_ipv4 != x_is_ipv4 {
+            return false;
+        }
+
         let mut a = self.value.clone();
         a.resize(16, 0);
-        let mut a = u128::from_le_bytes(a.try_into().unwrap());
+        let mut a = u128::from_be_bytes(a.try_into().unwrap());
 
         let mut b = x.value.clone();
         b.resize(16, 0);
-        let mut b = u128::from_le_bytes(b.try_into().unwrap());
+        let mut b = u128::from_be_bytes(b.try_into().unwrap());
 
-        let mask = (1u128 << x.length) - 1;
+        if x.length == 128 {
+            return a == b; // For /128, no masking needed
+        }
+
+        // Handle /0 case - everything is within the default route
+        if x.length == 0 {
+            return true;
+        }
+
+        // Prevent overflow when shift amount is >= 128
+        let shift_amount = 128 - x.length;
+        if shift_amount >= 128 {
+            return false; // Invalid case
+        }
+
+        let mask = !0u128 << shift_amount;
         a &= mask;
         b &= mask;
 
@@ -2456,6 +2480,55 @@ mod tests {
     use super::*;
     use pretty_assertions::assert_eq;
     use pretty_hex::*;
+    use std::net::{Ipv4Addr, Ipv6Addr};
+    use std::str::FromStr;
+
+    #[derive(Debug)]
+    enum AddressFamily {
+        IPv4,
+        IPv6,
+    }
+
+    #[derive(Debug)]
+    struct PrefixConversionTestCase {
+        description: &'static str,
+        address_family: AddressFamily,
+        prefix_length: u8,
+        input_bytes: Vec<u8>,
+        expected_address: &'static str,
+    }
+
+    impl PrefixConversionTestCase {
+        fn new_ipv4(
+            description: &'static str,
+            prefix_length: u8,
+            input_addr: Ipv4Addr,
+            expected_address: &'static str,
+        ) -> Self {
+            Self {
+                description,
+                address_family: AddressFamily::IPv4,
+                prefix_length,
+                input_bytes: input_addr.octets().to_vec(),
+                expected_address,
+            }
+        }
+
+        fn new_ipv6(
+            description: &'static str,
+            prefix_length: u8,
+            input_addr: Ipv6Addr,
+            expected_address: &'static str,
+        ) -> Self {
+            Self {
+                description,
+                address_family: AddressFamily::IPv6,
+                prefix_length,
+                input_bytes: input_addr.octets().to_vec(),
+                expected_address,
+            }
+        }
+    }
 
     #[test]
     fn header_round_trip() {
@@ -2532,31 +2605,240 @@ mod tests {
 
     #[test]
     fn prefix_within() {
-        let prefixes: &[Prefix] = &[
+        // Test IPv4 prefix containment
+        let ipv4_prefixes: &[Prefix] = &[
             "10.10.10.10/32".parse().unwrap(),
             "10.10.10.0/24".parse().unwrap(),
             "10.10.0.0/16".parse().unwrap(),
             "10.0.0.0/8".parse().unwrap(),
+            "0.0.0.0/0".parse().unwrap(),
         ];
 
-        for i in 0..prefixes.len() {
-            for j in i..prefixes.len() {
+        for i in 0..ipv4_prefixes.len() {
+            for j in i..ipv4_prefixes.len() {
                 // shorter prefixes contain longer or equal
-                assert!(prefixes[i].within(&prefixes[j]));
+                assert!(ipv4_prefixes[i].within(&ipv4_prefixes[j]));
                 if i != j {
                     // longer prefixes should not contain shorter
-                    assert!(!prefixes[j].within(&prefixes[i]))
+                    assert!(!ipv4_prefixes[j].within(&ipv4_prefixes[i]))
                 }
             }
         }
 
+        // Test IPv6 prefix containment
+        let ipv6_prefixes: &[Prefix] = &[
+            "2001:db8:1:1::1/128".parse().unwrap(),
+            "2001:db8:1:1::/64".parse().unwrap(),
+            "2001:db8:1::/48".parse().unwrap(),
+            "2001:db8::/32".parse().unwrap(),
+            "::/0".parse().unwrap(),
+        ];
+
+        for i in 0..ipv6_prefixes.len() {
+            for j in i..ipv6_prefixes.len() {
+                // shorter prefixes contain longer or equal
+                assert!(ipv6_prefixes[i].within(&ipv6_prefixes[j]));
+                if i != j {
+                    // longer prefixes should not contain shorter
+                    assert!(!ipv6_prefixes[j].within(&ipv6_prefixes[i]))
+                }
+            }
+        }
+
+        // Test non-overlapping prefixes
         let a: Prefix = "10.10.0.0/16".parse().unwrap();
         let b: Prefix = "10.20.0.0/16".parse().unwrap();
         assert!(!a.within(&b));
         let a: Prefix = "10.10.0.0/24".parse().unwrap();
         assert!(!a.within(&b));
 
-        let b: Prefix = "0.0.0.0/0".parse().unwrap();
-        assert!(a.within(&b));
+        let a: Prefix = "2001:db8:1::/48".parse().unwrap();
+        let b: Prefix = "2001:db8:2::/48".parse().unwrap();
+        assert!(!a.within(&b));
+
+        // Test default routes contain same-family prefixes
+        let ipv4_default: Prefix = "0.0.0.0/0".parse().unwrap();
+        let ipv6_default: Prefix = "::/0".parse().unwrap();
+
+        let any_ipv4: Prefix = "192.168.1.0/24".parse().unwrap();
+        let any_ipv6: Prefix = "2001:db8::/48".parse().unwrap();
+
+        assert!(any_ipv4.within(&ipv4_default));
+        assert!(any_ipv6.within(&ipv6_default));
+
+        // Test cross-family default route edge cases
+        // IPv4 prefixes should NOT be within IPv6 default route
+        assert!(!any_ipv4.within(&ipv6_default));
+        assert!(!ipv4_default.within(&ipv6_default));
+
+        // IPv6 prefixes should NOT be within IPv4 default route
+        assert!(!any_ipv6.within(&ipv4_default));
+        assert!(!ipv6_default.within(&ipv4_default));
+    }
+
+    #[test]
+    fn prefix_conversion() {
+        // Test both IPv4 and IPv6 prefix conversions including edge cases and host bit zeroing
+        let test_cases = vec![
+            // IPv4 test cases
+            // Input: 0.0.0.0 (default route)
+            PrefixConversionTestCase::new_ipv4(
+                "IPv4 default route",
+                0,
+                "0.0.0.0".parse().unwrap(),
+                "0.0.0.0",
+            ),
+            // Input: 10.255.255.255/8 -> 10.0.0.0/8 (host bits zeroed)
+            PrefixConversionTestCase::new_ipv4(
+                "IPv4 Class A with host bits",
+                8,
+                "10.255.255.255".parse().unwrap(),
+                "10.0.0.0",
+            ),
+            // Input: 172.31.255.255/12 -> 172.16.0.0/12 (host bits zeroed)
+            PrefixConversionTestCase::new_ipv4(
+                "IPv4 large private network with host bits",
+                12,
+                "172.31.255.255".parse().unwrap(),
+                "172.16.0.0",
+            ),
+            // Input: 172.16.255.255/16 -> 172.16.0.0/16 (host bits zeroed)
+            PrefixConversionTestCase::new_ipv4(
+                "IPv4 common allocation with host bits",
+                16,
+                "172.16.255.255".parse().unwrap(),
+                "172.16.0.0",
+            ),
+            // Input: 203.0.113.255/20 -> 203.0.112.0/20 (host bits zeroed)
+            PrefixConversionTestCase::new_ipv4(
+                "IPv4 prefix with host bits in last 12 bits",
+                20,
+                "203.0.113.255".parse().unwrap(),
+                "203.0.112.0",
+            ),
+            // Input: 192.168.1.123/24 -> 192.168.1.0/24 (host bits zeroed)
+            PrefixConversionTestCase::new_ipv4(
+                "IPv4 common subnet with host bits",
+                24,
+                "192.168.1.123".parse().unwrap(),
+                "192.168.1.0",
+            ),
+            // Input: 198.51.100.7/30 -> 198.51.100.4/30 (host bits zeroed)
+            PrefixConversionTestCase::new_ipv4(
+                "IPv4 point-to-point link with host bits",
+                30,
+                "198.51.100.7".parse().unwrap(),
+                "198.51.100.4",
+            ),
+            // Input: 10.0.0.1/32 -> 10.0.0.1/32 (no host bits to zero)
+            PrefixConversionTestCase::new_ipv4(
+                "IPv4 host route - no host bits to zero",
+                32,
+                "10.0.0.1".parse().unwrap(),
+                "10.0.0.1",
+            ),
+            // IPv6 test cases
+            // Input: :: (all zeros, default route)
+            PrefixConversionTestCase::new_ipv6(
+                "IPv6 default route",
+                0,
+                "::".parse().unwrap(),
+                "::",
+            ),
+            // Input: fd00:ffff:ffff:ffff:ffff:ffff:ffff:ffff/8 -> fd00::/8 (host bits zeroed)
+            PrefixConversionTestCase::new_ipv6(
+                "IPv6 unique local address prefix with host bits",
+                8,
+                "fd00:ffff:ffff:ffff:ffff:ffff:ffff:ffff".parse().unwrap(),
+                "fd00::",
+            ),
+            // Input: 2001:db8:1234:5678:9abc:def0:1122:3344/32 -> 2001:db8::/32 (host bits zeroed)
+            PrefixConversionTestCase::new_ipv6(
+                "IPv6 common allocation size with host bits",
+                32,
+                "2001:db8:1234:5678:9abc:def0:1122:3344".parse().unwrap(),
+                "2001:db8::",
+            ),
+            // Input: 2001:db8:1234:ffff:ffff:ffff:ffff:ffff/48 -> 2001:db8:1234::/48 (host bits zeroed)
+            PrefixConversionTestCase::new_ipv6(
+                "IPv6 site prefix with host bits in last 80 bits",
+                48,
+                "2001:db8:1234:ffff:ffff:ffff:ffff:ffff".parse().unwrap(),
+                "2001:db8:1234::",
+            ),
+            // Input: 2001:db8::1234:5678:9abc:def0/64 -> 2001:db8::/64 (host bits zeroed)
+            PrefixConversionTestCase::new_ipv6(
+                "IPv6 common prefix length with host bits",
+                64,
+                "2001:db8::1234:5678:9abc:def0".parse().unwrap(),
+                "2001:db8::",
+            ),
+            // Input: 2001:db8::ff/120 -> 2001:db8::/120 (host bits zeroed)
+            PrefixConversionTestCase::new_ipv6(
+                "IPv6 leaves only 8 host bits",
+                120,
+                "2001:db8::ff".parse().unwrap(),
+                "2001:db8::",
+            ),
+            // Input: 2001:db8::1/128 -> 2001:db8::1/128 (no host bits to zero)
+            PrefixConversionTestCase::new_ipv6(
+                "IPv6 host route - no host bits to zero",
+                128,
+                "2001:db8::1".parse().unwrap(),
+                "2001:db8::1",
+            ),
+        ];
+
+        for test_case in test_cases {
+            let prefix = Prefix {
+                length: test_case.prefix_length,
+                value: test_case.input_bytes,
+            };
+
+            match test_case.address_family {
+                AddressFamily::IPv4 => {
+                    let rdb_prefix4 = prefix.as_prefix4();
+                    assert_eq!(
+                        rdb_prefix4.length, test_case.prefix_length,
+                        "IPv4 length mismatch for {}",
+                        test_case.description
+                    );
+                    assert_eq!(
+                        rdb_prefix4.value,
+                        Ipv4Addr::from_str(test_case.expected_address).unwrap(),
+                        "IPv4 address mismatch for {}: expected {}, got {}",
+                        test_case.description,
+                        test_case.expected_address,
+                        rdb_prefix4.value
+                    );
+                    assert!(
+                        rdb_prefix4.host_bits_are_unset(),
+                        "IPv4 host bits not properly zeroed for {}",
+                        test_case.description
+                    );
+                }
+                AddressFamily::IPv6 => {
+                    let rdb_prefix6 = prefix.as_prefix6();
+                    assert_eq!(
+                        rdb_prefix6.length, test_case.prefix_length,
+                        "IPv6 length mismatch for {}",
+                        test_case.description
+                    );
+                    assert_eq!(
+                        rdb_prefix6.value,
+                        Ipv6Addr::from_str(test_case.expected_address).unwrap(),
+                        "IPv6 address mismatch for {}: expected {}, got {}",
+                        test_case.description,
+                        test_case.expected_address,
+                        rdb_prefix6.value
+                    );
+                    assert!(
+                        rdb_prefix6.host_bits_are_unset(),
+                        "IPv6 host bits not properly zeroed for {}",
+                        test_case.description
+                    );
+                }
+            }
+        }
     }
 }
--- a/bgp/src/messages.rs
+++ b/bgp/src/messages.rs
@@ -813,6 +813,15 @@ impl From<rdb::Prefix4> for Prefix {
     }
 }
 
+impl From<rdb::Prefix6> for Prefix {
+    fn from(p: rdb::Prefix6) -> Self {
+        Self {
+            value: p.value.octets().into(),
+            length: p.length,
+        }
+    }
+}
+
 /// A self-describing BGP path attribute
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct PathAttribute {
--- a/bgp/src/messages.rs
+++ b/bgp/src/messages.rs
@@ -2478,6 +2478,7 @@ pub enum Safi {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use mg_common::{cidr, ip, parse};
     use pretty_assertions::assert_eq;
     use pretty_hex::*;
     use std::net::{Ipv4Addr, Ipv6Addr};
@@ -2607,11 +2608,11 @@ mod tests {
     fn prefix_within() {
         // Test IPv4 prefix containment
         let ipv4_prefixes: &[Prefix] = &[
-            "10.10.10.10/32".parse().unwrap(),
-            "10.10.10.0/24".parse().unwrap(),
-            "10.10.0.0/16".parse().unwrap(),
-            "10.0.0.0/8".parse().unwrap(),
-            "0.0.0.0/0".parse().unwrap(),
+            cidr!("10.10.10.10/32"),
+            cidr!("10.10.10.0/24"),
+            cidr!("10.10.0.0/16"),
+            cidr!("10.0.0.0/8"),
+            cidr!("0.0.0.0/0"),
         ];
 
         for i in 0..ipv4_prefixes.len() {
@@ -2627,11 +2628,11 @@ mod tests {
 
         // Test IPv6 prefix containment
         let ipv6_prefixes: &[Prefix] = &[
-            "2001:db8:1:1::1/128".parse().unwrap(),
-            "2001:db8:1:1::/64".parse().unwrap(),
-            "2001:db8:1::/48".parse().unwrap(),
-            "2001:db8::/32".parse().unwrap(),
-            "::/0".parse().unwrap(),
+            cidr!("2001:db8:1:1::1/128"),
+            cidr!("2001:db8:1:1::/64"),
+            cidr!("2001:db8:1::/48"),
+            cidr!("2001:db8::/32"),
+            cidr!("::/0"),
         ];
 
         for i in 0..ipv6_prefixes.len() {
@@ -2646,22 +2647,22 @@ mod tests {
         }
 
         // Test non-overlapping prefixes
-        let a: Prefix = "10.10.0.0/16".parse().unwrap();
-        let b: Prefix = "10.20.0.0/16".parse().unwrap();
+        let a: Prefix = cidr!("10.10.0.0/16");
+        let b: Prefix = cidr!("10.20.0.0/16");
         assert!(!a.within(&b));
-        let a: Prefix = "10.10.0.0/24".parse().unwrap();
+        let a: Prefix = cidr!("10.10.0.0/24");
         assert!(!a.within(&b));
 
-        let a: Prefix = "2001:db8:1::/48".parse().unwrap();
-        let b: Prefix = "2001:db8:2::/48".parse().unwrap();
+        let a: Prefix = cidr!("2001:db8:1::/48");
+        let b: Prefix = cidr!("2001:db8:2::/48");
         assert!(!a.within(&b));
 
         // Test default routes contain same-family prefixes
-        let ipv4_default: Prefix = "0.0.0.0/0".parse().unwrap();
-        let ipv6_default: Prefix = "::/0".parse().unwrap();
+        let ipv4_default: Prefix = cidr!("0.0.0.0/0");
+        let ipv6_default: Prefix = cidr!("::/0");
 
-        let any_ipv4: Prefix = "192.168.1.0/24".parse().unwrap();
-        let any_ipv6: Prefix = "2001:db8::/48".parse().unwrap();
+        let any_ipv4: Prefix = cidr!("192.168.1.0/24");
+        let any_ipv6: Prefix = cidr!("2001:db8::/48");
 
         assert!(any_ipv4.within(&ipv4_default));
         assert!(any_ipv6.within(&ipv6_default));
@@ -2685,56 +2686,56 @@ mod tests {
             PrefixConversionTestCase::new_ipv4(
                 "IPv4 default route",
                 0,
-                "0.0.0.0".parse().unwrap(),
+                ip!("0.0.0.0"),
                 "0.0.0.0",
             ),
             // Input: 10.255.255.255/8 -> 10.0.0.0/8 (host bits zeroed)
             PrefixConversionTestCase::new_ipv4(
                 "IPv4 Class A with host bits",
                 8,
-                "10.255.255.255".parse().unwrap(),
+                ip!("10.255.255.255"),
                 "10.0.0.0",
             ),
             // Input: 172.31.255.255/12 -> 172.16.0.0/12 (host bits zeroed)
             PrefixConversionTestCase::new_ipv4(
                 "IPv4 large private network with host bits",
                 12,
-                "172.31.255.255".parse().unwrap(),
+                ip!("172.31.255.255"),
                 "172.16.0.0",
             ),
             // Input: 172.16.255.255/16 -> 172.16.0.0/16 (host bits zeroed)
             PrefixConversionTestCase::new_ipv4(
                 "IPv4 common allocation with host bits",
                 16,
-                "172.16.255.255".parse().unwrap(),
+                ip!("172.16.255.255"),
                 "172.16.0.0",
             ),
             // Input: 203.0.113.255/20 -> 203.0.112.0/20 (host bits zeroed)
             PrefixConversionTestCase::new_ipv4(
                 "IPv4 prefix with host bits in last 12 bits",
                 20,
-                "203.0.113.255".parse().unwrap(),
+                ip!("203.0.113.255"),
                 "203.0.112.0",
             ),
             // Input: 192.168.1.123/24 -> 192.168.1.0/24 (host bits zeroed)
             PrefixConversionTestCase::new_ipv4(
                 "IPv4 common subnet with host bits",
                 24,
-                "192.168.1.123".parse().unwrap(),
+                ip!("192.168.1.123"),
                 "192.168.1.0",
             ),
             // Input: 198.51.100.7/30 -> 198.51.100.4/30 (host bits zeroed)
             PrefixConversionTestCase::new_ipv4(
                 "IPv4 point-to-point link with host bits",
                 30,
-                "198.51.100.7".parse().unwrap(),
+                ip!("198.51.100.7"),
                 "198.51.100.4",
             ),
             // Input: 10.0.0.1/32 -> 10.0.0.1/32 (no host bits to zero)
             PrefixConversionTestCase::new_ipv4(
                 "IPv4 host route - no host bits to zero",
                 32,
-                "10.0.0.1".parse().unwrap(),
+                ip!("10.0.0.1"),
                 "10.0.0.1",
             ),
             // IPv6 test cases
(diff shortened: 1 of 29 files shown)


bgp/src/test.rs
```diff
@@ -83,23 +42,28 @@ fn test_basic_peering() {
     // Ensure that r1's peer session to r2 has gone back to connect.
     r2.shutdown();
     d2.shutdown();
-    wait_for_eq!(r1_session.state(), FsmStateKind::Connect);
```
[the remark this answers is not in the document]
taspelund: The main reason I modified this is because I believed the FSM state observed is timing-dependent, not consistent.

It seemed to me like the goal of this test case is to confirm that "shutdown" on one side brings the session down.
If so, then looking for "!established" is a less timing-dependent way to confirm that things have been completely torn down.

Otherwise, the state on the active side could be "connect", "opensent" (since the listener isn't torn down on the shutdown side, the transport can still come up) or "idle".
I was hoping to avoid a scenario where the test fails because the active peer is in "idle" or "opensent" and we were only looking for "connect".

I don't have super strong opinions about this though, so if reverting that change means the test passes consistently again, then I'm fine with that.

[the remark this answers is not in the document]
taspelund: I'm not sure at what point this test became unreliable.  I'll go back through and figure out which of my changes borked things.  I don't suspect the issue is with the logic change of "state != established", but I could be wrong

bgp/src/clock.rs
```diff
@@ -2,77 +2,235 @@
...
+        let v = lock!(self.value);
+        v.enabled && v.remaining.is_zero()
+    }
+
+    /// Display time remaining on this timer
+    pub fn remaining(&self) -> Duration {
+        lock!(self.value).remaining
+    }
+
+    /// Reset the value of a timer to the timer's interval.
```
rcgoodfellow: I'm not sure this is what I would call jitter, as to me jitter means the deviation from the target interval varies from cycle to cycle. This is more like a phase offset? I guess if `reset` is called often, then this is more jitter like.

bgp/src/connection.rs
```diff
@@ -21,63 +26,189 @@ pub const MAX_MD5SIG_KEYLEN: usize = 80;
...
+        self.local
+    }
+
+    /// Get the remote socket address
+    pub fn remote(&self) -> SocketAddr {
+        self.remote
+    }
+}
+
+impl PartialOrd for ConnectionId {
```
rcgoodfellow: Just use`#[derive(PartialOrd)]`?

bgp/src/connection.rs
```diff
@@ -21,63 +26,189 @@ pub const MAX_MD5SIG_KEYLEN: usize = 80;
...
+    fn apply_policy(
+        conn: &Cnx,
+        min_ttl: Option<u8>,
+        md5_key: Option<String>,
+    ) -> Result<(), Error>;
+}
 
-    /// Try to connect to a peer. On success messages from the peer a
-    /// propagated through `event_tx`.
+/// Implementors of this trait initiate outbound BGP connections to peers.
```
rcgoodfellow: This comment seems like it's more in reference to an implementation of this trait rather than the trait itself? Similar for the comment on `connect` below.

bgp/src/connection_tcp.rs
```diff
@@ -97,348 +93,675 @@ impl BgpListener<BgpConnectionTcp> for BgpListenerTcp {
...
-            Mutex<BTreeMap<IpAddr, Sender<FsmEvent<BgpConnectionTcp>>>>,
+            Mutex<BTreeMap<IpAddr, SessionEndpoint<BgpConnectionTcp>>>,
         >,
-        _timeout: Duration, //TODO implement
+        timeout: Duration,
     ) -> Result<BgpConnectionTcp, Error> {
-        let (conn, mut peer) = self.listener.accept()?;
+        let (conn, mut peer) = self.listener.accept().map_err(|e| {
+            if e.kind() == std::io::ErrorKind::WouldBlock {
+                Error::Timeout
```
rcgoodfellow: Going immediately to a timeout error on `WouldBlock` does not seem right. Shouldn't we be retrying until for a `timeout` duration?

bgp/src/connection_tcp.rs
```diff
@@ -97,348 +93,675 @@ impl BgpListener<BgpConnectionTcp> for BgpListenerTcp {
...
+            };
 
-        let sa: socket2::SockAddr = self.peer.into();
-        match s.connect_timeout(&sa, timeout) {
-            Ok(()) => {
-                let new_conn: TcpStream = s.into();
-                lock!(self.conn).replace(new_conn.try_clone()?);
-                if let Some(ttl) = min_ttl {
-                    self.set_min_ttl(ttl)?;
+            // Start SA tracking and keepalive for Illumos MD5
```
rcgoodfellow: Why are we separating this from initialization and creating this mostly duplicate logic? Initialization and kicking off the keepalive thread happened all in one spot.

bgp/src/connection_tcp.rs
```diff
@@ -97,348 +93,675 @@ impl BgpListener<BgpConnectionTcp> for BgpListenerTcp {
...
+            let peer = self.peer;
+            let event_tx = params.event_tx;
+            let timeout = params.timeout;
+            let dropped = self.dropped.clone();
+            let log = self.log.clone();
+            let creator = self.creator;
+            let conn_id = self.id;
+
+            // Clone the TcpStream for the recv thread
+            if let Ok(conn) = lock!(self.conn).try_clone() {
```
rcgoodfellow: Handle the failure case?

bgp/src/connection_tcp.rs
```diff
@@ -97,348 +93,675 @@ impl BgpListener<BgpConnectionTcp> for BgpListenerTcp {
...
+                std::sync::atomic::Ordering::SeqCst,
+                std::sync::atomic::Ordering::Acquire,
+            )
+            .is_err()
+        {
+            // Already started, nothing to do
+            return;
+        }
+
+        // Take the params (they should exist since we haven't started yet)
```
rcgoodfellow: I think using the typestate pattern would help tighten this up. If for some reason this expectation fails, there is no logging here and we'll have no idea.

bgp/src/connection_tcp.rs
```diff
@@ -547,10 +891,18 @@ impl BgpConnectionTcp {
 
     fn send_msg(
         stream: &mut TcpStream,
-        log: &Logger,
+        _log: &Logger,
+        _creator: &str,
         msg: Message,
     ) -> Result<(), Error> {
-        trace!(log, "sending {:#?}", msg);
+        // connection_log_lite!(log,
```
rcgoodfellow: remove commented out code or un-comment out?

bgp/src/connection_tcp.rs
```diff
@@ -666,34 +1020,47 @@ impl BgpConnectionTcp {
...
 
         // Get one run in before returning, this helps the SAs to
         // get set up before setting up the socket.
-        Self::do_sa_keepalive(&self.sas, &self.log);
+        Self::do_sa_keepalive(&self.sas, &self.log, self.conn());
 
-        debug!(self.log, "spawning security association keepalive loop");
+        connection_log!(self,
+            debug,
+            // space after loop is needed... because macros?
```
rcgoodfellow: what was the issue here?

bgp/src/connection_tcp.rs
```diff
@@ -829,3 +1206,86 @@ fn sa_set(src: SocketAddr, dst: SocketAddr) -> [(SocketAddr, SocketAddr); 4] {
...
+fn init_md5_associations(
+    fd: i32,
+    key: &str,
+    locals: Vec<SocketAddr>,
+    peer: SocketAddr,
+) -> Result<(), Error> {
+    // Set MD5 security associations for Illumos
+    for local in locals.iter() {
+        for (a, b) in sa_set(*local, peer) {
+            if let Err(e) = libnet::pf_key::tcp_md5_key_add(
```
rcgoodfellow: Is it safe to assume that this will always be an add? What if the connection rolled over and we are back here inside the keepalive time? In that case we would need to handle the update case.

bgp/src/log.rs
```diff
@@ -0,0 +1,346 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+macro_rules! session_log {
+    ($self:expr, $level:ident, $conn:expr, $msg:expr; $($key:expr => $value:expr),*) => {
+        slog::$level!($self.log,
+            $msg;
+            "component" => crate::COMPONENT_BGP,
+            "module" => crate::MOD_NEIGHBOR,
```
rcgoodfellow: Every log here has both

```rust
"component" => crate::COMPONENT_BGP
```

and

```rust
"module" => crate::MOD_NEIGHBOR
```

so what is the latter telling us?

bgp/src/messages.rs
```diff
@@ -425,6 +506,20 @@ impl OpenMessage {
     }
 }
 
+impl Display for OpenMessage {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        let mut param_string = String::new();
```
rcgoodfellow: this can probably be streamlined to

```rust
self.parameters
  .iter()
  .map(ToString::to_string)
  .collect::<Vec<_>>()
  .join(",")
```

bgp/src/messages.rs
```diff
@@ -2575,11 +2911,179 @@ pub enum Safi {
...
+                    }
+                } else {
+                    // First byte is length, rest are the address bytes
+                    let length = wire_bytes[0];
+                    let value = wire_bytes[1..].to_vec();
+                    Self { length, value }
+                }
+            }
+            Err(_) => {
+                // Fallback for wire conversion errors
```
rcgoodfellow: Maybe this should be `TryFrom` instead. What are the potential ramifications if we fail to convert between versions and default to 0.0.0.0/0 below?

bgp/src/session.rs
```diff
@@ -203,185 +475,214 @@ pub enum FsmEvent<Cnx: BgpConnection> {
...
+            md5_auth_key: None,
+            multi_exit_discriminator: None,
+            communities: BTreeSet::new(),
+            local_pref: None,
+            enforce_first_as: false,
+            allow_import: ImportExportPolicy::default(),
+            allow_export: ImportExportPolicy::default(),
+            vlan_id: None,
+            // Default timer values based on BGP standards
+            connect_retry_time: Duration::from_secs(120),
```
rcgoodfellow: This feels a bit dangerous. I guess keepalive and hold will get negotiated, but implicit defaults for the others still concern me. The only place I see `default() being used in `bgp_admin` these values are available in the `Neighbor` request object. I'm also wondering why timers are now a part of the session info object.

bgp/src/router.rs
```diff
@@ -170,14 +193,33 @@ impl<Cnx: BgpConnection + 'static> Router<Cnx> {
...
+        mut a2s: MutexGuard<BTreeMap<IpAddr, SessionEndpoint<Cnx>>>,
         peer: PeerConfig,
-        bind_addr: SocketAddr,
+        bind_addr: Option<SocketAddr>,
         event_tx: Sender<FsmEvent<Cnx>>,
         event_rx: Receiver<FsmEvent<Cnx>>,
         info: SessionInfo,
     ) -> Result<Arc<SessionRunner<Cnx>>, Error> {
-        a2s.insert(peer.host.ip(), event_tx.clone());
+        // Update the SessionInfo with timer values from peer config
```
rcgoodfellow: It seems strange to do this here. If we're going to merge this information into `SessionInfo` can we just construct `SessionInfo` with complete data from the outset?

bgp/src/session.rs
```diff
@@ -569,19 +1082,180 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                    return;
+                }
+
+                // Thread finished, join to check for panics
+                self.join_connector_thread(old_handle, "checking previous");
+            }
+            // Lock is released here at end of scope
+        }
+
+        // Spawn new connector with NO locks held (critical for avoiding deadlock)
```
rcgoodfellow: What deadlock is being avoided? I see `self.connector_handle` being held above, but that's not available to `Cnx::Connector::connect`. `self.connector_handle` is locked again at 1178, but it seems like it could just be held the whole time? Am I missing something?

bgp/src/session.rs
```diff
@@ -569,19 +1082,180 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+        let mut connections = lock!(self.connections);
+        // Disable timers before dropping to ensure timers stop immediately
+        for conn in connections.values() {
+            conn.clock().disable_all();
+        }
+        connections.clear();
+        *lock!(self.primary) = None;
+    }
+
+    /// This is the BGP peer state machine entry point. This function only
```
rcgoodfellow: @rcgoodfellow you are here.

rcgoodfellow (commented):
Thanks for all the work in putting this together @taspelund! I'm through BGP up to the FSM. Will pick back up there tomorrow.

bgp/src/connection.rs
```diff
@@ -21,63 +26,189 @@ pub const MAX_MD5SIG_KEYLEN: usize = 80;
...
+impl ConnectionCreator {
+    /// Get the string representation of the creator
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            ConnectionCreator::Dispatcher => "dispatcher",
+            ConnectionCreator::Connector => "connector",
+        }
+    }
+
+    pub fn direction(&self) -> &'static str {
```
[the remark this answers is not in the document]
taspelund: I thought about this recently too.  I would honestly rather just rip out ConnectionCreator and replace it with ConnectionDirection, as I think that has more meaning for anyone who isn't already familiar with the codebase.

bgp/src/connection_tcp.rs
```diff
@@ -97,348 +93,675 @@ impl BgpListener<BgpConnectionTcp> for BgpListenerTcp {
...
+            // Establish the connection (THIS IS THE BLOCKING CALL)
+            let sa: socket2::SockAddr = peer.into();
+            let new_conn: TcpStream = match s.connect_timeout(&sa, timeout) {
+                Ok(()) => s.into(),
                 Err(e) => {
-                    error!(
-                        self.log,
-                        "source address selection for {}: {e}", self.peer
+                    connection_log_lite!(log,
+                        debug,
```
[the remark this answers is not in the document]
taspelund: I would argue it should be `notice` rather than warning, since this would be normal and require no action from the operator if the peer was simply shutdown.

Slog doesn't have a macro for notice, so we either do info or warn. I'd lean towards info, personally, but I don't have strong feelings about it.

bgp/src/session.rs
```diff
@@ -569,19 +1084,188 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+    fn initiate_connection(&self, timeout: Duration) {
+        // Clone session info first, before acquiring any locks
+        let session_info = lock!(self.session).clone();
+
+        // First critical section: check and join old connector thread
+        {
+            let mut handle_guard = lock!(self.connector_handle);
+
+            if let Some(old_handle) = handle_guard.take() {
+                // Check if thread is still running
```
rcgoodfellow: Since the connect timeout is 1/3 of the connect retry timer, it seems we should not be able to get here through a connect retry. Is this handling a case where there is a failure that short-circuits us here in a lap through idle where idle hold time is zero. E.g. what does the path look like to land us here with a connect thread still running.

bgp/src/session.rs
```diff
@@ -1915,46 +6289,244 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+    ) {
+        match reason {
+            StopReason::Reset => {
+                if let Some(c1) = conn1 {
+                    self.send_admin_reset_notification(c1);
+                }
+                if let Some(c2) = conn2 {
+                    self.send_admin_reset_notification(c2);
+                }
+                self.counters
```
rcgoodfellow: Can you explain the logic behind the counter updates for the various stop reasons?

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                            session_timer!(self, connect_retry).stop();
+                            self.connect_retry_counter
+                                .fetch_add(1, Ordering::Relaxed);
+                            session_log_lite!(
+                                self,
+                                warn,
+                                "{} event not allowed in this state, fsm transition to idle",
+                                session_event.title()
+                            );
+                            return FsmState::Idle;
```
rcgoodfellow: Do we really want to transition to idle here - or try to correct the situation by just stopping the idle hold timer and carrying on in connect?

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                }
+
+                Err(RecvTimeoutError::Timeout) => continue,
+
                 Err(e) => {
-                    err!(self; "on connect event rx: {e}");
+                    session_log_lite!(self, error, "event rx error: {e}";
+                        "error" => format!("{e}")
+                    );
+                    // TODO: Possible death loop. Should we just panic here?
```
rcgoodfellow: So the only other error condition here is `RecvTimeoutError::Disconnected` which means that [all of the senders have been disconnected/dropped](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ca1dfe68943344d5445ee0f797802894).  That does mean that this will be a death loop with `continue` if we hit that error.

The channel is created all the way up in `bgp_admin::add_neighbor`, there the tx side is cloned and delved out. Following the setup path and all the tx clones that are made, one of the clones is held as a member of the `SessionRunner` itself, so it really should be impossible for this to happen.

I would probably put an `unreachable!()` statement here with a comment that explains why this should be an unreachable path.

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                    SessionEvent::TcpConnectionAcked(new)
+                    | SessionEvent::TcpConnectionConfirmed(new) => {
+                        let new_direction = new.direction();
+                        if new_direction == conn.direction() {
+                            collision_log!(
+                                self,
+                                error,
+                                new,
+                                conn,
+                                "rejected new {new_direction} connection ({}): multiple {new_direction} connections not allowed",
```
rcgoodfellow: very long line. per rustfmt.toml we're trying to keep things within 80 cols. it seems rustfmt does not catch this for macros. There are quite a few other macro lines beyond 80 cols.

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                        );
+                    }
+                },
+
+                FsmEvent::Connection(connection_event) => {
+                    match connection_event {
+                        ConnectionEvent::Message { msg, ref conn_id } => {
+                            match self.get_conn(conn_id) {
+                                Some(connection) => {
+                                    if connection.id() != conn.id() {
```
rcgoodfellow: The only place I see connections being added to `self.connections` is in `register_conn` where connections are inserted with a key that is equal to their connection `id()`. When would we encounter a connection entry where the key is not equal to the `id()` of the value?

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                                        );
+                                        self.stop(
+                                            Some(&connection),
+                                            None,
+                                            StopReason::FsmError,
+                                        );
+                                        continue;
+                                    }
+                                }
+                                None => {
```
rcgoodfellow: Can you add a comment on what gets us here?

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
-                err!(self; "on active event rx: {e}");
-                return FsmState::Active(conn);
+                session_log!(
+                    self,
+                    error,
+                    pc.conn,
+                    "event rx error: {e}";
+                    "error" => format!("{e}")
+                );
+                //TODO possible death loop. Should we just panic here?
```
rcgoodfellow: [Ditto](https://github.com/oxidecomputer/maghemite/pull/533/files#r2483088217)

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+        conn_pair: CollisionPair<Cnx>,
+    ) -> FsmState<Cnx> {
+        match conn_pair {
+            CollisionPair::OpenConfirm(exist, new) => {
+                collision_log!(
+                    self,
+                    info,
+                    new,
+                    exist.conn,
+                    "collision detected: new connection [{:?}, conn_id: {}], existing connection [{:?}, conn_id: {}]",
```
rcgoodfellow: Should this log and the next indicate whether we coming from OpenConfirm versus OpenSent?

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                            exist.conn,
+                            "{title} event not allowed in this state, fsm transition to idle";
+                            "event" => title
+                        );
+                        self.stop(
+                            Some(&new),
+                            Some(&exist.conn),
+                            StopReason::FsmError,
+                        );
                         return FsmState::Idle;
```
rcgoodfellow: [Ditto](https://github.com/oxidecomputer/maghemite/pull/533/files#r2483047453).

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
```
rcgoodfellow: [Ditto](https://github.com/oxidecomputer/maghemite/pull/533/files#r2483047453).

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
-            FsmEvent::Message(Message::Open(_)) => {
-                self.counters
-                    .unexpected_open_message
-                    .fetch_add(1, Ordering::Relaxed);
-                wrn!(self; "unexpected open message in open confirm");
-                FsmState::Idle
+        }
+    }
+
+    /// Handler for collisions when existing connection was in OpenSent state
```
rcgoodfellow: @rcgoodfellow you are here

rcgoodfellow (commented):
Next round of comments follows.

bgp/src/connection_tcp.rs
```diff
@@ -596,37 +956,31 @@ impl BgpConnectionTcp {
...
     #[cfg(target_os = "illumos")]
-    fn set_md5_sig_fd(
+    fn manage_md5_associations(
         &self,
         fd: i32,
         key: &str,
         locals: Vec<SocketAddr>,
         peer: SocketAddr,
     ) -> Result<(), Error> {
+        // First set up the socket using the standalone function
```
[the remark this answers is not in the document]
taspelund: PR opened to include md5 password in interop here: https://github.com/oxidecomputer/testbed/pull/208

I do think we want more thorough testing for this than just confirming that interop comes up and establishes (like you mentioned in a different comment, making sure key rotation works on illumos is a key (no pun intended) item to ensure coverage for)

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
-            other => {
-                wrn!(
-                    self;
-                    "event {:?} not expected in open confirm, ignoring", other
-                );
-                FsmState::OpenConfirm(pc)
+        }
+    }
+
+    /// Collision Resolution logic.
```
rcgoodfellow: This comment does not really line up with what's going on in this function. This function is associating the connection id for an incoming message with either a new or existing connection in a collision context.

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                                                .fetch_add(1, Ordering::Relaxed);
+                                            self.counters
+                                                .connection_retries
+                                                .fetch_add(1, Ordering::Relaxed);
+                                            self.stop(Some(&exist), None, StopReason::FsmError);
+                                            self.set_primary_conn(Some(ConnectionKind::Partial(new.clone())));
+                                            return FsmState::OpenSent(new);
+                                        }
+
+                                        // Resolve collision on first Open - don't wait for both
```
rcgoodfellow: IIUC this means that if new wins, we ditch the successful open for a connection that may later have a failed open?

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                        }
+
+                        // Event 11
+                        ConnectionEvent::KeepaliveTimerExpires(
+                            conn_id,
+                        )
+                        // Event 12
+                        | ConnectionEvent::DelayOpenTimerExpires(
+                            conn_id,
+                        ) => {
```
rcgoodfellow: Similar to the comments on dropping to idle on unexpected timer events. This feels like a bug. Should we be amplifying the impact of the bug by stopping connections in response and letting the bug influence the collision result like this?

bgp/src/session.rs
```diff
@@ -1290,249 +4912,908 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                    }
+                    session_log!(
+                        self,
+                        info,
+                        pc.conn,
+                        "rx {}, fsm transition to idle",
+                        session_event.title()
+                    );
+                    self.stop(Some(&pc.conn), None, StopReason::FsmError);
+                    self.exit_established(pc)
```
rcgoodfellow: [Ditto](https://github.com/oxidecomputer/maghemite/pull/533/files#r2483047453).

bgp/src/session.rs
```diff
@@ -1290,249 +4912,908 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                        return FsmState::Established(pc);
+                    }
+                    session_log!(
+                        self,
+                        info,
+                        pc.conn,
+                        "rx delay open timer expires, fsm transition to idle"
+                    );
+                    self.stop(Some(&pc.conn), None, StopReason::FsmError);
+                    self.exit_established(pc)
```
rcgoodfellow: [Ditto](https://github.com/oxidecomputer/maghemite/pull/533/files#r2483047453).

bgp/src/session.rs
```diff
@@ -1290,249 +4912,908 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                            "rx {} for unexpected known connection (conn_id: {}). closing..",
+                            connection_event.title(),
+                            conn_id.short();
+                            "event" => connection_event.title()
+                        );
+                        if let Some(conn) = self
+                            .get_conn(conn_id)
+                            .filter(|c| conn_timer!(c, keepalive).enabled())
+                        {
+                            self.stop(Some(&conn), None, StopReason::FsmError);
```
rcgoodfellow: Stopping the connection feels unrelated to the fact that its keepalive timer expired? The only timer that is stopped in `stop` for `FsmError` is the connect retry timer.

bgp/src/session.rs
```diff
@@ -1290,249 +4912,908 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                            return FsmState::Established(pc);
+                        }
+                        session_log!(
+                            self,
+                            info,
+                            pc.conn,
+                            "rx delay open timer expires, fsm transition to idle"
+                        );
+                        self.stop(Some(&pc.conn), None, StopReason::FsmError);
                         self.exit_established(pc)
```
rcgoodfellow: [Ditto](https://github.com/oxidecomputer/maghemite/pull/533/files#r2483047453).

bgp/src/session.rs
```diff
@@ -1290,249 +4912,908 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                            "rx {} for unexpected known connection (conn_id: {}). closing..",
+                            connection_event.title(),
+                            conn_id.short();
+                            "event" => connection_event.title()
+                        );
+                        if let Some(conn) = self
+                            .get_conn(conn_id)
+                            .filter(|c| conn_timer!(c, delay_open).enabled())
+                        {
+                            self.stop(Some(&conn), None, StopReason::FsmError);
```
rcgoodfellow: Stopping the connection feels unrelated to the fact that its delay open timer expired? The only timer that is stopped in `stop` for `FsmError` is the connect retry timer.

bgp/src/session.rs
```diff
@@ -1290,249 +4912,908 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                                "message" => "open",
+                                "message_contents" => format!("{om}")
+                            );
+                            self.bump_msg_counter(msg_kind, true);
+                            self.stop(
+                                Some(&pc.conn),
+                                None,
+                                StopReason::FsmError,
+                            );
+                            self.exit_established(pc)
```
rcgoodfellow: Previously we just logged this as a warning and carried on. I can see this happening if we got an active connection quickly and an open message got sent from a peer's active FSM path before it has a chance to halt it's own active FSM path.

bgp/src/session.rs
```diff
@@ -1290,249 +4912,908 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                            // to reduce the surface area of this non-standard
+                            // change.
+                            if matches!(m.error_subcode,
+                                ErrorSubcode::Cease(CeaseErrorSubcode::ConnectionCollisionResolution)) {
+
+                                session_log!(
+                                    self,
+                                    info,
+                                    pc.conn,
+                                    "rx collision resolution notification (conn_id: {}), fsm transition to active",
```
rcgoodfellow: We are actually transitioning to Idle below.

bgp/src/session.rs
```diff
@@ -1698,34 +6055,23 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
         let mut msg = match self.asn {
             Asn::FourOctet(asn) => OpenMessage::new4(
                 asn,
-                self.clock
-                    .timers
-                    .hold_configured_interval
-                    .lock()
-                    .unwrap()
-                    .as_secs() as u16,
+                conn_timer!(conn, hold).interval.as_secs() as u16,
```
rcgoodfellow: For an open message, shouldn't this be the configured interval and not the negotiated one? Same for the two octet ASN case below.

bgp/src/session.rs
```diff
@@ -1915,46 +6289,244 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+    fn exit_established(&self, pc: PeerConnection<Cnx>) -> FsmState<Cnx> {
+        self.cleanup_established(&pc);
         FsmState::Idle
     }
 
+    fn bump_msg_counter(&self, msg: MessageKind, unexpected: bool) {
+        match msg {
+            MessageKind::Open => {
+                self.counters.opens_received.fetch_add(1, Ordering::Relaxed);
+                self.counters
```
rcgoodfellow: All these unexpected counter bumps should only be in the `if unexpected` branch below?

bgp/src/session.rs
```diff
@@ -2135,20 +6738,24 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
-        {
+        if info.enforce_first_as {
             self.enforce_first_as(update, peer_as)?;
         }
         Ok(())
     }
 
     fn apply_static_update_policy(&self, update: &mut UpdateMessage) {
-        if self.is_ebgp() {
+        if let Some(ebgp) = self.is_ebgp()
```
rcgoodfellow: Consider `self.is_ebgp().unwrap_or(false)`.

rcgoodfellow (commented):
Comments for the rest of session.rs.

mg-api/src/lib.rs
```diff
@@ -168,28 +172,74 @@ pub trait MgAdminApi {
         request: Query<AsnSelector>,
     ) -> Result<HttpResponseDeleted, HttpError>;
 
+    #[endpoint { method = PUT, path = "/bgp/config/origin6", versions = VERSION_IPV6_BASIC.. }]
```
rcgoodfellow: Note that we only really need to version bump API endpoints that are directly consumed by omicron. Endpoints whose only consumer is `mgadm` don't need multiple versions.

rdb/src/db.rs
```diff
@@ -125,38 +149,184 @@ impl Db {
...
+                    self.static_rib6()
+                        .into_iter()
+                        .map(|(p6, paths)| (Prefix::from(p6), paths)),
+                );
+                rib
+            }
+        }
+    }
+
+    #[allow(dead_code)]
```
rcgoodfellow: Was it intentional to leave these unused functions here?

rdb/src/db.rs
```diff
@@ -464,48 +842,84 @@ impl Db {
...
+                            error,
+                            "error fetching static route entry {item:?}: {e}";
+                            "unit" => UNIT_PERSISTENT
                         );
                         return None;
                     }
                 };
 
                 let key = String::from_utf8_lossy(&key);
+                // XXX: figure out how to handle removal of old static routes
```
rcgoodfellow: See my earlier comment on non-zero host bits.

rdb/src/types.rs
```diff
@@ -519,3 +697,179 @@ impl From<Prefix6> for PrefixChangeNotification {
...
+#[cfg_attr(feature = "clap", derive(ValueEnum))]
+pub enum AddressFamily {
+    /// Internet Protocol Version 4 (IPv4)
+    Ipv4,
+    /// Internet Protocol Version 6 (IPv6)
+    Ipv6,
+}
+
+/// Trait for encoding/decoding prefixes to/from BGP wire format
+pub trait BgpWireFormat: Sized {
```
rcgoodfellow: Feels like this should be in the `bgp` crate? Maybe in `bgp::messages`?

rdb/src/types.rs
```diff
@@ -519,3 +697,179 @@ impl From<Prefix6> for PrefixChangeNotification {
...
+    type Error = Error;
+
+    fn to_wire(&self) -> Result<Vec<u8>, Self::Error> {
+        let mut buf = vec![self.length];
+        let n = (self.length as usize).div_ceil(8);
+        buf.extend_from_slice(&self.value.octets()[..n]);
+        Ok(buf)
+    }
+
+    fn from_wire(input: &[u8]) -> Result<(&[u8], Self), Self::Error> {
```
rcgoodfellow: If this were in `bgp::messages` you could probably reuse some of the nom parser combinators for this.

rcgoodfellow (commented):
Thanks @taspelund! This set of comments concludes my initial review.

This is a ton of work. Thanks for bringing this all together. The BGP FSM bits in particular are a significant improvement, both in the collision handling and the overall expression of the state machine.

Looking forward to getting this across the line.

bgp/src/session.rs
taspelund: I agree with 80 cols, but in the case of string literals I think an exception should be made.
It's very difficult to grep for log strings if they are split/escaped across multiple lines.

bgp/src/session.rs
```diff
@@ -177,16 +232,233 @@ pub enum FsmEvent<Cnx: BgpConnection> {
...
+    HoldTimerExpires(ConnectionId),
+
+    /// Fires when the connection's keepalive timer expires.
+    KeepaliveTimerExpires(ConnectionId),
+
+    /// Fires when the connection's delay open timer expires.
+    DelayOpenTimerExpires(ConnectionId),
+}
+
+impl fmt::Debug for ConnectionEvent {
```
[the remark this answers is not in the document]
taspelund: Good point. I've moved to deriving debug for all enum *Event types

bgp/src/session.rs
```diff
@@ -177,16 +232,233 @@ pub enum FsmEvent<Cnx: BgpConnection> {
...
+    ConnectRetryTimerExpires,
+
+    /// Fires when the session's idle hold timer expires.
+    IdleHoldTimerExpires,
+
+    /// Fires when the local systems tcp-syn recieved a syn-ack from the remote
+    /// peer and the local system has sent an ack.
+    /// i.e.
+    /// We have ACKed the peer's connection.
+    /// We use this event to indicate an inbound connection has completed.
```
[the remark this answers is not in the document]
taspelund: I think that comment may have been re-used from its original definition ;-)

I'll fix it

bgp/src/session.rs
taspelund: I don't recall the exact flow where this was relevant, as it was something I found and fixed during some fairly quick code iteration.
I want to say the deadlock was happening when passing `lock!(self.session)` as an argument to Connector::connect(), but I ended up restructuring it so that wouldn't be an issue.

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                        );
+                        // unregister_conn() call is not needed, since we
+                        // haven't registered `extra`
+                        continue;
+                    }
+                },
+
+                FsmEvent::Connection(connection_event) => {
+                    match connection_event {
+                        // XXX: Make sure we always log the message type + conn_id
```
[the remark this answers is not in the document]
taspelund: This was a note to self.  Will confirm it's been addressed and will yank it out.

bgp/src/session.rs
taspelund: I'm sure I missed updating some comments when I refactored collision resolution, but I'm not sure how this comment ended up on this method at all.
Thanks for spotting it, I'll clean it up

bgp/src/session.rs
taspelund: One of the major motivations for the ConnectionId design was to avoid issues like that, which is accomplished by a couple different ideas working together:
1. Every BgpConnection has a unique ID, and all events specific to that connection include the ID as part of the connection
2. The FSM validates the event's ConnectionId before honoring the event.
3. The recv loop for a BgpConnection is started _by the FSM_ after passing semantic validation (e.g. are we in an active collision? are we already established?). This ensures no Messages are admitted into the event queue unless the FSM accepts the connection that message is tied to.

This specific codepath is handling an Open received in Established that isn't tied to a ConnectionId owned by this FSM state (pc.conn) or at least known in the registry (e.g. a connection surviving ConnectionCollision state per the long-winded comment above).
Any time we receive an Open that isn't one of the above circumstances is a legitimate FSM error perpetrated by the peer, and I think we should treat it as such.

bgp/src/session.rs
taspelund: If we're sending an Open, we have not performed negotiation yet.  This is true so long as DelayOpen is not implemented.
Currently we do the negotiation in handle_open(), which could probably be refactored into its own method so it's not strictly tied to receipt of an Open.

I definitely think it would be better to pull this from config rather than relying on the implementation details of "we haven't negotiated yet," so I'll push a change for this.

bgp/src/session.rs
```diff
@@ -1855,7 +6215,9 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
                 .push(PathAttributeValue::MultiExitDisc(med).into());
         }
 
-        if self.is_ibgp() {
+        if let Some(ibgp) = self.is_ibgp()
```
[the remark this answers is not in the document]
taspelund: The reason I have that method setup to return `Option<bool>` is in case the caller makes a query while the peer is down (or before we know if they're internal/external).
Maybe that doesn't make the most sense ergonomically, but it felt like the idea of answering "is this session iBGP?" with `false` when the session is down seemed semantically wrong to me.

What are your thoughts on how to approach that?

Also, do you think the is_ibgp()/is_ebgp() methods should account for the optional configured ASN? `self.session.remote_asn`

bgp/src/session.rs
taspelund: Yeah, and I should probably add in a block comment that explains it too.

The idea here is to have a single helper method to clean up connections that need to be stopped.
When looking at the FSM description in the RFC, there are a lot of commonalities as to what items need to be updated upon a given error condition, regardless of what FSM state that error condition was encountered in.
This method is intended to centralize much of these situations into a single method which uses the `StopReason` enum to indicate which set of items need to be updated.

I also found that pretty much anytime stop() was being called, a connection was being unregistered as well.  So I added the connection registry deletion to this method as well; again, as a way to centralize that handling.

rdb/src/db.rs
taspelund: Doesn't look like they're worth keeping -- the static_rib variants above either.  Going to remove them.

Thanks for pointing them out!

rdb/src/types.rs
```diff
@@ -200,6 +215,12 @@ impl Ord for Prefix4 {
 }
 
 impl Prefix4 {
+    pub fn new(ip: Ipv4Addr, length: u8) -> Self {
+        let mut new = Self { value: ip, length };
+        new.unset_host_bits();
```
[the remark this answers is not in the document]
taspelund: Do you mean a docstring? or an update to public docs?

bgp/src/connection_tcp.rs
taspelund: From a functional standpoint, it appears that always calling `add` works just fine.
After leaving the interop topology running for 4+ days, the kernel only returns 8 SAs when checking the output of `tcpkey {dump|save}`, which is correct given we expect the SA configuration to setup 4 per peer.
e.g.
For a connection between us (169.254.20.1) and a peer (169.254.20.2), I see these SAs:
1. src=169.254.20.1, sport=0, dst=169.254.20.2, dport=179
2. src=169.254.20.1, sport 179, dst=169.254.20.2, dport=0
3. src=169.254.20.2, sport=0, dst=169.254.20.1, dport=179
4. src=169.254.20.2, sport 179, dst=169.254.20.1, dport=0

If there was a leak, I would suspect there to be FAR more than 8 SAs given we use a lifetime of 2 minutes.

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
-        };
 
-        match event {
-            FsmEvent::ManualStart => {
-                if lock!(self.session).passive_tcp_establishment {
-                    let conn = Cnx::new(
-                        self.bind_addr,
-                        self.neighbor.host,
-                        self.log.clone(),
+            let event = match self.event_rx.recv_timeout(IO_TIMEOUT) {
```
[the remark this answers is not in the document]
rcgoodfellow: Something that encapsulates the next event extraction logic and makes that code reusable.

bgp/src/session.rs
```diff
@@ -686,568 +1393,3477 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                         *
+                         *   - completes BGP initialization
+                         *
+                         *   - sends an OPEN message to its peer,
+                         *
+                         *   - sets the HoldTimer to a large value, and
+                         *
+                         *   - changes its state to OpenSent.
+                         *
+                         * A HoldTimer value of 4 minutes is suggested.
```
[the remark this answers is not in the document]
rcgoodfellow: My comment is about this specific 4 minute guidance that should be removed. It's not generally valid guidance.

bgp/src/session.rs
rcgoodfellow: My observation here is that if we view the established connection and the new connection as both working toward getting to established, in this case we may be preferring a connection that is less far along over one that has gotten further. And it seems odd to resolve a collision between two connections that have not reached equal footing in terms of the their position in the state machine yet.

bgp/src/session.rs
taspelund: I was referring to using grep on the source, not a log file.  Making the string multi-line via escaped newlines makes it harder to grep for the logging call.

taspelund:
Peers resetting the session in a loop: ✅ 
Local system resetting sessions in a loop: ✅ 
Route refresh (tested bidirectionally): ✅ 
GR: upon further inspection of our code, GR helper has not been implemented.  Routes are only marked stale when we send a Route Refresh (with the assumption that either the peer re-advertises us routes and un-marks them stale, or they get removed by the db reaper)
MD5 mismatch: ✅ 
Changing MD5 password: ✅ 
Toggling MD5 password: ✅ 

The manual testing of resetting sessions in a loop did a few thread leaks (issues in the shutdown flag logic) which I've fixed. I've also updated the MD5 SA Keepalive thread logic to have shorter sleeps so we aren't stuck waiting 60 seconds for the shutdown flag to get noticed (manual testing showed that flapping the session ~1/s resulted in ~60 threads lingering per peer, which makes sense given the old sleep was 60s).

With that all done, I would say everything is looking good and is ready for final review.

PR #554, opened 2025-09-26
move to rust 1.90 and 2024 edition

rcgoodfellow:
Move to Rust 1.90 / edition 2024 and deal with clippy issues.


rcgoodfellow:
Interop failure is due to buildomat image compatability issue that is being sorted out currently.

PR #555, opened 2025-09-27
update ddm-admin version to drop the leading v

sunshowers:
Found this because the Dropshot API manager isn't able to parse the ddm-admin document currently on main.

Generated a new `ddm-admin.json` by running `ddm-apigen`.


rcgoodfellow (approved):

rcgoodfellow:
The interop test failure is due to a buildomat image compatibility issue, this is fine to merge.

PR #558, opened 2025-10-02
make mg-admin and ddm-admin APIs versioned

sunshowers:
Last part of https://github.com/oxidecomputer/omicron/issues/8922 for Maghemite: make the admin APIs versioned. Also remove the `openapi` buildomat job, since it's a remnant from when maghemite was a private repo (Omicron stopped using it in https://github.com/oxidecomputer/omicron/pull/5271).

See the readme within https://github.com/oxidecomputer/dropshot-api-manager for instructions.

### DIFF
--- a/mgd/src/admin.rs
+++ b/mgd/src/admin.rs
@@ -15,7 +15,7 @@ use mg_api::*;
 use mg_common::stats::MgLowerStats;
 use rdb::{BfdPeerConfig, Db, Prefix};
 use slog::o;
-use slog::{Logger, error, info, warn};
+use slog::{Logger, error, info};
 use std::collections::HashMap;
 use std::net::{IpAddr, Ipv6Addr, SocketAddr};
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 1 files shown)


sunshowers:
@rcgoodfellow would need to remove `openapi` as a required workflow from GHA:

image

mgd/src/admin.rs
```diff
@@ -49,13 +49,19 @@ pub fn start_server(
...
+                omicron_common::api::VERSION_HEADER,
+                mg_api::latest_version(),
+            ),
+        )));
 
     info!(log, "admin: listening on {}", sa);
 
     Ok(tokio::spawn(async move {
-        match server.start().await {
+        match server.start() {
```
rcgoodfellow, line 64: This also needs to be something like

```rust
match server.start().expect("failed to start API server").await
```

sunshowers: ah yeah, whoops

PR #584, opened 2025-12-15
Add support for MP-BGP and BGP unnumbered

taspelund:
Initial implementation of MP-BGP, supporting IPv4 Unicast and IPv6 Unicast address-families.

Adds MP_REACH_NLRI and MP_UNREACH_NLRI types and handling, capabilities negotiation and handling, updated message parsing framework, enhanced UPDATE error handling (RFC 7606), lots of tests, and more flexible BGP next-hop encoding support (in preparation for extended next-hop and BGP unnumbered).

Also pulls in bgp unnumbered support.

### DIFF
--- a/rdb/src/db.rs
+++ b/rdb/src/db.rs
@@ -664,7 +664,7 @@ impl Db {
         match rib_in.get(prefix) {
             // rib-in has paths worth evaluating for loc-rib
             Some(paths) => {
-                match bestpaths(paths, fanout.get() as usize) {
+                match bestpaths(paths, fanout.get().into()) {
                     // bestpath found at least 1 path for loc-rib
                     Some(bp) => {
                         rib_loc.insert(*prefix, bp.clone());
@@ -701,7 +701,7 @@ impl Db {
         match rib_in.get(prefix) {
             // rib-in has paths worth evaluating for loc-rib
             Some(paths) => {
-                match bestpaths(paths, fanout.get() as usize) {
+                match bestpaths(paths, fanout.get().into()) {
                     // bestpath found at least 1 path for loc-rib
                     Some(bp) => {
                         rib_loc.insert(*prefix, bp.clone());
@@ -747,7 +747,7 @@ impl Db {
                         prefix,
                         paths,
                         &mut rib4_loc,
-                        fanout.get() as usize,
+                        fanout.get().into(),
                     );
                 }
             }
@@ -763,7 +763,7 @@ impl Db {
                         prefix,
                         paths,
                         &mut rib6_loc,
-                        fanout.get() as usize,
+                        fanout.get().into(),
                     );
                 }
             }
@@ -1031,7 +1031,7 @@ impl Db {
                         p4,
                         paths,
                         &mut rib4_loc,
-                        fanout.get() as usize,
+                        fanout.get().into(),
                     );
                 }
             }
@@ -1058,7 +1058,7 @@ impl Db {
                         p6,
                         paths,
                         &mut rib6_loc,
-                        fanout.get() as usize,
+                        fanout.get().into(),
                     );
                 }
             }


taspelund:
`test-interop` is currently failing because interop hasn't gotten the updates to support the new `mgadm bgp config neighbor create` syntax.  This is addressed in https://github.com/oxidecomputer/testbed/pull/244 and the test should start passing when we pull this in (either by merging the testbed PR into main, or by pinning maghemite to the PR branch).

bgp/src/connection_tcp.rs
```diff
@@ -685,33 +742,51 @@ impl BgpConnectionTcp {
...
                         "error" => format!("{e}")
                     );
 
-                    let subcode = match e {
-                        Error::UnsupportedCapability(_) => {
-                            OpenErrorSubcode::UnsupportedCapability
-                        }
-                        _ => OpenErrorSubcode::Unspecific,
+                    let (subcode, reason) = match &e {
+                        Error::UnsupportedCapability(cap) => (
```
rcgoodfellow: Was trying to sort out the difference between `UnsupportedCapability` and `UnsupportedCapabilityCode` and it looks like the former error condition is only possible on `to_wire`?

bgp/src/messages.rs
```diff
@@ -1499,13 +2521,743 @@ pub enum AsPathType {
...
+#[schemars(
+    description = "A BGP next-hop address in one of three formats: IPv4, IPv6 single, or IPv6 double."
+)]
+pub enum BgpNexthop {
+    Ipv4(Ipv4Addr),
+    Ipv6Single(Ipv6Addr),
+    Ipv6Double(Ipv6DoubleNexthop),
+}
+
+impl BgpNexthop {
```
rcgoodfellow: @rcgoodfellow you are here.

rcgoodfellow (commented):
First round of comments follows. Will pick up the rest in the morning.

bgp/src/messages.rs
```diff
@@ -1499,13 +2521,743 @@ pub enum AsPathType {
...
+            BgpNexthop::Ipv6Double(_) => ((Ipv6Addr::BITS * 2) / 8) as u8,
+        }
+    }
+
+    /// Serialize next-hop to wire format bytes
+    pub fn to_bytes(&self) -> Vec<u8> {
+        match self {
+            BgpNexthop::Ipv4(addr) => addr.octets().to_vec(),
+            BgpNexthop::Ipv6Single(addr) => addr.octets().to_vec(),
+            BgpNexthop::Ipv6Double(addrs) => {
```
rcgoodfellow: For stuff like this `chain` is nice.

```rust
BgpNexthop::Ipv6Double(addrs) => addrs
    .global
    .octets()
    .into_iter()
    .chain(addrs.link_local.octets().into_iter())
    .collect()
```

bgp/src/messages.rs
```diff
@@ -1499,13 +2521,743 @@ pub enum AsPathType {
...
+
+        // Parse Next-hop Length (1 byte)
+        let (input, nh_len) = be_u8::<_, nom::error::Error<&[u8]>>(input)
+            .map_err(|e| UpdateParseErrorReason::AttributeParseError {
+                type_code: Some(PathAttributeTypeCode::MpReachNlri as u8),
+                detail: format!("failed to parse next-hop length: {e}"),
+            })?;
+
+        // Extract next-hop bytes
+        if input.len() < nh_len as usize {
```
rcgoodfellow:
```suggestion
        if input.len() < usize::from(nh_len) {
```

bgp/src/messages.rs
```diff
@@ -1499,13 +2521,743 @@ pub enum AsPathType {
...
+        let nexthop =
+            BgpNexthop::from_bytes(nh_bytes, nh_len, afi).map_err(|_| {
+                let expected = match afi {
+                    Afi::Ipv4 => "4",
+                    Afi::Ipv6 => "16 or 32",
+                };
+                UpdateParseErrorReason::InvalidMpNextHopLength {
+                    afi: afi_raw,
+                    expected,
+                    got: nh_len as usize,
```
rcgoodfellow:
```suggestion
                    got: usize::from(nh_len),
```

bgp/src/messages.rs
```diff
@@ -2997,47 +4778,557 @@ impl From<Capability> for CapabilityCode {
...
+    pub error_code: ErrorCode,
+    pub error_subcode: ErrorSubcode,
+    pub reason: NotificationParseErrorReason,
+}
+
+impl Display for NotificationParseError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{:?}/{:?}: {}",
```
rcgoodfellow: Is using debug formatters intentional here?

bgp/src/messages.rs
```diff
@@ -2997,47 +4778,557 @@ impl From<Capability> for CapabilityCode {
...
+    pub error_code: ErrorCode,
+    pub error_subcode: ErrorSubcode,
+    pub reason: RouteRefreshParseErrorReason,
+}
+
+impl Display for RouteRefreshParseError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{:?}/{:?}: {}",
```
rcgoodfellow: Is using debug formatters intentional here?

bgp/src/session.rs
```diff
@@ -529,10 +831,10 @@ pub struct SessionInfo {
     /// Ensure that routes received from eBGP peers have the peer's ASN as the
     /// first element in the AS path.
     pub enforce_first_as: bool,
-    /// Policy governing imported routes.
-    pub allow_import: ImportExportPolicy,
-    /// Policy governing exported routes.
-    pub allow_export: ImportExportPolicy,
+    /// IPv4 Unicast address family configuration (None = disabled)
+    pub ipv4_unicast: Option<Ipv4UnicastConfig>,
```
rcgoodfellow: just a general note on ipv4 unicast / ipv6 unicast, i think we can dispense with the unicast part. Seems unlikely to me that MBGP is going to make its way into our stack, even when we do get to PIM that will likely be its own thing, and then the question becomes unicast as opposed to what?

bgp/src/session.rs
```diff
@@ -577,18 +879,24 @@ impl SessionInfo {
             communities: BTreeSet::new(),
             local_pref: None,
             enforce_first_as: false,
-            allow_import: ImportExportPolicy::default(),
-            allow_export: ImportExportPolicy::default(),
+            ipv4_unicast: Some(Ipv4UnicastConfig {
```
rcgoodfellow: Why is IPv4 `Some(Default::default())` and IPv6 `None`?

bgp/src/session.rs
```diff
@@ -3502,6 +3903,60 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
                         FsmState::Idle
                     }
                 }
+
+                ConnectionEvent::ParseError {
```
rcgoodfellow: This block is nearly identical to the one at 3946, some logic and logging consolidation would make the state machine easier to read.

bgp/src/session.rs
```diff
@@ -5207,100 +6062,232 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
-                            .map(|x| crate::messages::Prefix::from(*x))
-                            .collect(),
-                        nlri: to_announce
-                            .into_iter()
-                            .map(|x| crate::messages::Prefix::from(*x))
-                            .collect(),
-                    };
+                            FsmState::Established(pc)
+                        }
+                        TypedImportExportPolicy::V6(previous6) => {
```
rcgoodfellow: This code is nearly identical to the v4 code, let's make an issue to consolidate dual stack logic wherever possible and cite this as one of the places to address.

bgp/src/session.rs
```diff
@@ -6037,18 +7061,19 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
             "sending route refresh";
             "message" => "route refresh"
         );
-        if let Err(e) = conn.send(Message::RouteRefresh(RouteRefreshMessage {
-            afi: Afi::Ipv4 as u16,
-            safi: Safi::NlriUnicast as u8,
-        })) {
+        let rr = Message::RouteRefresh(RouteRefreshMessage {
+            afi: af as u16,
+            safi: Safi::Unicast as u8,
```
rcgoodfellow:
```suggestion
            safi: Safi::Unicast.into()
```

bgp/src/session.rs
```diff
@@ -6305,81 +7330,273 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                    BgpNexthop::Ipv4(addr) => addr,
+                    _ => {
+                        return Err(Error::InvalidAddress(
+                            "IPv4 routes require IPv4 next-hop".into(),
+                        ));
+                    }
+                };
+
+                let mut path_attributes = self.router.base_attributes();
+                path_attributes.push(PathAttributeValue::NextHop(nh4).into());
```
rcgoodfellow: Do we need to consider mp-reach ipv4 NLRI here for dual-stack setups?

bgp/src/session.rs
```diff
@@ -6648,95 +7906,307 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+
     fn handle_refresh(
         &self,
         msg: RouteRefreshMessage,
         pc: &PeerConnection<Cnx>,
     ) -> Result<(), Error> {
-        // XXX: Update for IPv6
-        if msg.afi != Afi::Ipv4 as u16 {
-            return Ok(());
+        if msg.safi != Safi::Unicast as u8 {
```
rcgoodfellow:
```suggestion
        if msg.safi != Safi::Unicast.into() {
```

bgp/src/session.rs
```diff
@@ -6827,7 +8250,184 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
 
-        //TODO(IPv6) iterate through MpReachNlri attributes for IPv6
+        // Process MP_REACH_NLRI for IPv4 and IPv6 routes
+        if let Some(reach) = update.mp_reach() {
+            match reach {
+                MpReachNlri::Ipv4Unicast(reach4) => {
+                    let mp_nexthop = match &reach4.nexthop {
+                        BgpNexthop::Ipv4(ip4) => IpAddr::V4(*ip4),
+                        BgpNexthop::Ipv6Single(ip6) => IpAddr::V6(*ip6),
+                        BgpNexthop::Ipv6Double(addrs) => {
```
rcgoodfellow: Same question as above on using the local component of the nexthop vs the global.

bgp/src/session.rs
```diff
@@ -6827,7 +8250,184 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                                as_path,
+                                stale: None,
+                            }),
+                            vlan_id: lock!(self.session).vlan_id,
+                        };
+
+                        self.db.add_bgp_prefixes(&mp_nlri4, path4);
+                    }
+                }
+                MpReachNlri::Ipv6Unicast(reach6) => {
```
rcgoodfellow: This code can also go in the follow up dual-stack consolidation issue I mentioned as it's very similar to the V4 code.

mg-admin-client/Cargo.toml
```diff
@@ -16,5 +16,6 @@ schemars.workspace = true
 chrono.workspace = true
 uuid.workspace = true
 rdb-types.workspace = true
+bgp.workspace = true
```
rcgoodfellow: This means all clients of mgd are compiling our whole BGP implementation into their binaries.

mg-admin-client/src/lib.rs
```diff
@@ -22,6 +22,11 @@ progenitor::generate_api!(
         Prefix = rdb_types::Prefix,
         AddressFamily = rdb_types::AddressFamily,
         ProtocolFilter = rdb_types::ProtocolFilter,
+        JitterRange = bgp::params::JitterRange,
```
rcgoodfellow: We should probably move these types out of `bgp` and into `mg-api` crate to avoid `bgp` compiling into clients.

rdb/src/bestpath.rs
```diff
@@ -26,39 +26,55 @@ use itertools::Itertools;
...
     }
 
+    // Extract references to path values from the HashMap
+    let path_refs: Vec<&Path> = paths.values().collect();
+
     // Partition the choice space on whether routes are shutdown or not. If we
     // only have shutdown routes then use those. Otherwise use active routes
-    let (active, shutdown): (BTreeSet<&Path>, BTreeSet<&Path>) =
-        paths.iter().partition(|x| x.shutdown);
+    let (active, shutdown): (Vec<&Path>, Vec<&Path>) =
```
rcgoodfellow: Why the change to a vec?

bgp/src/messages.rs
```diff
@@ -779,70 +879,538 @@ impl UpdateMessage {
...
             }
-            let (out, pa) = PathAttribute::from_wire(buf)?;
-            result.push(pa);
-            buf = out;
+
+            // ===== FRAMING: Parse attribute header (type + length) =====
+
+            // 1. Parse 2-byte type header (flags + type_code)
+            let (remaining, type_bytes) =
+                match take::<_, _, nom::error::Error<&[u8]>>(2usize)(buf) {
```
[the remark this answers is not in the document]
taspelund: done. will be in the next commit

bgp/src/messages.rs
```diff
@@ -1499,13 +2521,743 @@ pub enum AsPathType {
...
+
+        // AFI (2 bytes)
+        buf.extend_from_slice(&(self.afi() as u16).to_be_bytes());
+
+        // SAFI (1 byte)
+        buf.push(self.safi() as u8);
+
+        // Next-hop
+        let nh_bytes = self.nexthop().to_bytes();
+        buf.push(nh_bytes.len() as u8); // Next-hop length
```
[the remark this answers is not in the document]
taspelund: I updated this to use `BgpNexthop::byte_len()` so that .to_wire() can be infallible

bgp/src/session.rs
```diff
@@ -6827,7 +8250,184 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                        }
+                    };
+
+                    let nexthop6 = match &reach6.nexthop {
+                        BgpNexthop::Ipv6Single(ip6) => IpAddr::V6(*ip6),
+                        BgpNexthop::Ipv6Double(addrs) => {
+                            IpAddr::V6(addrs.global)
+                        }
+                        BgpNexthop::Ipv4(ip4) => {
+                            // IPv4 nexthop for IPv6 routes is unusual but possible
```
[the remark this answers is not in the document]
taspelund: Interesting... I figured the v4-over-v6 implementation would have been fully protocol-version agnostic. Is this something we need to make accommodations for in mgd? e.g. dropping v6 updates when the next-hop is IPv4 (or IPv4-mapped IPv6)

rdb/src/db.rs
```diff
@@ -68,9 +68,9 @@ const BESTPATH_FANOUT: &str = "bestpath_fanout";
 /// Default bestpath fanout value. Maximum number of ECMP paths in RIB.
 const DEFAULT_BESTPATH_FANOUT: u8 = 1;
 
-pub type Rib = BTreeMap<Prefix, BTreeSet<Path>>;
-pub type Rib4 = BTreeMap<Prefix4, BTreeSet<Path>>;
-pub type Rib6 = BTreeMap<Prefix6, BTreeSet<Path>>;
+pub type Rib = BTreeMap<Prefix, HashMap<PathKey, Path>>;
```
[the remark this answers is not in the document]
taspelund: Short description in the git commit:
https://github.com/oxidecomputer/maghemite/pull/584/changes/5ea560d0fe36858e81ced9d360d5100168a8deff

Basically when I added unit tests for the nexthop override functionality, I was finding that the tests failed after receiving an Update message for a previously-known route would not actually have its path replaced in the RIB.
e.g.
1. R1 sends 10.0.0.0/24 via 1.1.1.1 to R2
2. R2 installs 10.0.0.0/24 via 1.1.1.1
3. R1 sends 10.0.0.0/24 via 1.1.1.2 to R2
4. R2 inserts 1.1.1.2 as an additional path for 10.0.0.0/24
5. R2's final state: 10.0.0.0/24 via {1.1.1.1, 1.1.1.2}

This ended up being because the Update RX codepath was calling .replace() when the RIB key (Prefix) already existed in the BTreeMap, but the trait impl that BTreeMap was using to check key equality (I think `Eq` or `PartialEq`, I don't recall off the top of my head) was finding the Prefixes to be not equal... thus resulting in an insertion rather than a replacement.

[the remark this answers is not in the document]
taspelund: I thought I had used a much more thorough git commit that explained this in more depth, but clearly that didn't make it into the branch...

bgp/src/session.rs
taspelund: Sure, makes sense. Is this something you want changed for this PR?

bgp/src/session.rs
taspelund: Some() indicates the address-family is enabled, None indicates it is not.
Or are you asking why we default to enabling ipv4 and not ipv6?

bgp/src/session.rs
```diff
@@ -219,17 +383,129 @@ impl<Cnx: BgpConnection> From<&FsmState<Cnx>> for FsmStateKind {
...
 
     // The shaper for the router has changed. Event contains previous checker.
     // Current shaper is available in the router policy object.
     ShaperChanged(Option<rhai::AST>),
 
-    /// Fires when export policy has changed.
-    ExportPolicyChanged(ImportExportPolicy),
+    /// Fires when an export policy has changed.
+    /// Contains the previous policy for determining routes to re-advertise.
+    ExportPolicyChanged(TypedImportExportPolicy),
```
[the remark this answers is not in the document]
taspelund: I renamed `ImportExportPolicy` to `ImportExportPolicyV1` and named the new enum `ImportExportPolicy`.
So we're back to roughly the same nomenclature we had before.

bgp/src/session.rs
```diff
@@ -6648,95 +7906,307 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
                 );
                 Vec::new()
             }
         };
 
+        let nexthop = match update.nexthop() {
+            Ok(nh) => match nh {
+                BgpNexthop::Ipv4(ip4) => IpAddr::V4(ip4),
+                BgpNexthop::Ipv6Single(ip6) => IpAddr::V6(ip6),
+                BgpNexthop::Ipv6Double(addrs) => IpAddr::V6(addrs.global),
```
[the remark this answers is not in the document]
taspelund: I filed an issue for this so we can follow-up at a later time: https://github.com/oxidecomputer/maghemite/issues/619

rdb/src/bestpath.rs
```diff
@@ -26,39 +26,55 @@ use itertools::Itertools;
...
+        return None;
+    }
+
     // Short-circuit: if there's only 1 candidate, then it is the best
     if paths.len() == 1 {
         return Some(paths.clone());
     }
 
+    // Extract references to path values from the HashMap
+    let path_refs: Vec<&Path> = paths.values().collect();
```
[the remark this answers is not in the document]
taspelund: I could have sworn I had updated this at the same time I had the more thorough git commit for the RIB type changes...  Anyway, will clean this up

bgp/src/session.rs
```diff
@@ -7126,6 +8797,179 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
     }
+
+    fn get_counters(&self) -> PeerCounters {
+        PeerCounters::from(self.counters.as_ref())
+    }
+
+    pub fn get_peer_info(&self) -> PeerInfo {
+        let fsm_state = self.state();
+        let dur = self.current_state_duration().as_millis() % u64::MAX as u128;
+        let fsm_state_duration = dur as u64;
```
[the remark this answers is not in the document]
taspelund: I'm going to update this spot and the existing logic to cap the value at u64::MAX.  For the current API version, I am going to just expose this as a Duration since that's what we're using for the other session timers.

bgp/src/session.rs
taspelund: https://github.com/oxidecomputer/maghemite/issues/618

bgp/src/messages.rs
```diff
@@ -616,6 +653,12 @@ impl OpenMessage {
 
     /// Deserialize an open message from wire format.
     pub fn from_wire(input: &[u8]) -> Result<OpenMessage, Error> {
+        // RFC 4271 §4.2: OPEN minimum 10 bytes body
+        // (1 version + 2 ASN + 2 hold_time + 4 ID + 1 opt_param_len)
+        if input.len() < 10 {
```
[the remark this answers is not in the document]
taspelund: I don't have super strong opinions here, but if you want this changed then we can do that

bgp/src/session.rs
```diff
@@ -6827,7 +8250,184 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                MpReachNlri::Ipv4Unicast(reach4) => {
+                    let mp_nexthop = match &reach4.nexthop {
+                        BgpNexthop::Ipv4(ip4) => IpAddr::V4(*ip4),
+                        BgpNexthop::Ipv6Single(ip6) => IpAddr::V6(*ip6),
+                        BgpNexthop::Ipv6Double(addrs) => {
+                            IpAddr::V6(addrs.global)
+                        }
+                    };
+
+                    let mp_nlri4: Vec<Prefix> = reach4
```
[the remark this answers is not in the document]
taspelund: Agreed. Filed https://github.com/oxidecomputer/maghemite/issues/622 for this.

rdb/src/db.rs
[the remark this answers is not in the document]
rcgoodfellow: Is a simpler solution here the following?

```diff
 impl Ord for Path {
     fn cmp(&self, other: &Self) -> Ordering {
+        if let (Some(a), Some(b)) = (&self.bgp, &other.bgp) {
+            if a.peer == b.peer {
+                return Ordering::Equal;
+            }
+        }
         if self.nexthop != other.nexthop {
             return self.nexthop.cmp(&other.nexthop);
         }
```

taspelund:
CI failures for `falcon` job seem to be due to https://www.illumos.org/issues/17853
It would seem that if our instance of falcon is the only one running, then cleanup of the VNICs would trigger this bug

bgp/src/params.rs
```diff
@@ -265,19 +816,44 @@ pub struct ApplyRequest {
...
+    #[serde(flatten)]
+    pub parameters: BgpPeerParameters,
+}
+
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, PartialEq)]
+pub struct UnnumberedBgpPeerConfig {
+    pub interface: String,
+    pub name: String,
+    pub router_lifetime: u16,
+    #[serde(flatten)]
```
jgallagher: Do we need to `#[serde(flatten)]` this? It causes all of these fields to be duplicated in the progenitor client, which leads to some kinda painful duplication (e.g., https://github.com/oxidecomputer/omicron/blob/8186a1d914d4297136eb0892a3881a38dc9fc68b/sled-agent/src/bootstrap/early_networking.rs#L570-L605 and https://github.com/oxidecomputer/omicron/blob/8186a1d914d4297136eb0892a3881a38dc9fc68b/sled-agent/src/bootstrap/early_networking.rs#L617-L654).

This is certainly not "hold R18" urgent, and if changing this isn't trivial I'm happy to file a followup issue either here (if it makes sense to remove the `flatten`) or in omicron (if we really want to keep the `flatten` and should come up with other ways to reduce the duplication on the client side).

rdb/src/db.rs
```diff
@@ -726,13 +726,29 @@ impl Db {
...
             let mut rib4_loc = lock!(self.rib4_loc);
-            for (prefix, paths) in self.full_rib4().iter() {
+            for (prefix, paths) in rib4_in.iter() {
                 if bestpath_needed(&Prefix::from(*prefix), paths) {
-                    self.update_rib4_loc(&rib4_in, &mut rib4_loc, prefix);
+                    Self::update_rib_loc(
+                        prefix,
+                        paths,
+                        &mut rib4_loc,
+                        fanout.get() as usize,
```
taspelund: `usize::from(fanout.get())`

rdb/src/db.rs
```diff
@@ -976,8 +1024,15 @@ impl Db {
...
-                if let Prefix::V4(p4) = prefix {
-                    self.update_rib4_loc(&rib4_in, &mut rib4_loc, p4);
+                if let Prefix::V4(p4) = prefix
+                    && let Some(paths) = rib4_in.get(p4)
+                {
+                    Self::update_rib_loc(
+                        p4,
+                        paths,
+                        &mut rib4_loc,
+                        fanout.get() as usize,
```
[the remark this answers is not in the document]
rcgoodfellow: fixed (`.into()`)

bgp/src/params.rs
rcgoodfellow: Tracking here

- https://github.com/oxidecomputer/maghemite/issues/635

PR #585, opened 2025-12-18
various ipv6 fixes

rcgoodfellow:
- fix sync of v6 routes to dendrite


mg-lower/src/dendrite.rs
```diff
@@ -433,42 +433,51 @@ pub(crate) fn get_routes_for_prefix(
                 };
             }
 
-            dpd_routes
```
rcgoodfellow: This was doing the same thing as the `result` construction above, but ignoring the mg-lower tag and throwing out the `result`. This change uses `result` instead.

taspelund (approved):

PR #598, opened 2026-01-02
BGP unnumbered

rcgoodfellow:
This PR adds support for BGP unnumbered.

Depends on

- oxidecomputer/maghemite#584
- oxidecomputer/dendrite#181

## The Main Event

Unnumbered peers are added as a new type of peer with their own set of API functions. When an unnumbered peer is added it goes into the pending state. The new peer request is handed over to an unnumbered peer manager where IPv6/NDP router advertisements (RA) are used to discover the address of the peer over the interface provided. When a router is discovered on the interface the pending peer then turns into a real peer. We use the IPv6 link-local address discovered in the RA and create a BGP session using that address. From here the extant BGP machinery works in the same way it currently does.

Neighbor discovery is a one-shot process when the unnumbered peer is created. Once we determine the peer address through an RA, that is what we use. If the peer address changes, the session needs to be deleted and recreated so discovery will happen again. No attempt is made to track neighbor state beyond initial session establishment. This may be revisited in a future release. One shot semantics has the benefit of being simple and stable. I don't believe we want to be overly sensitive to NDP RA dynamics as that would undercut the intent of things like BGP hold time and TCP connection tracking. Having a persistent monitor of NDP dynamics will need to be designed carefully.

Unnumbered support is designed to work in a point to point setting. We do not currently handle situations where there are multiple reachable routers reachable over a broadcast domain that the unnumbered interface is connected to. We simply attempt to establish a session with the first peer we find. This is how things have generally worked for unnumbered in my experience. We could develop something more robust, but this becomes a rather complex state machine where we have to try to establish sessions with every peer we discover. This could lead to doomed BGP sessions iterating through a state machine indefinitely.

## Testing

This PR also adds a new testing substrate in the `falcon-lab` folder. BGP unnumbered requires multicast ICMPv6/NDP to work. This means we need a testing environment that routers connected over a broadcast domain, and some other unnumbered implementation to peer with to ensure we are not deluding ourselves with self-peering. Using illumos zones would have been nice here, but none of the BGP implementations I'm aware of on illumos besides mgd fully support unnumbered. FRR on the other hand has great support for unnumbered. So peering with a linux-based router puts us in falcon territory.

Another really important thing to test for unnumbered is mg-lower interaction. We need test coverage on nexthop interface mapping for data plane route installation. The interop test environment is currently upper-half only. Due to license nonsense, the interop topology machinery is also stuck in purgatory. So in this PR I decided to add a new falcon-based testing substrate that only uses unencumbered peers and the oxide nodes are softnpu vms so we have a full lower half with dendrite, P4 and the whole 9 yards. This also means the testing machinery can actually live in this repository which is a nice win. The best way to get acquainted with this new testing substrate is starting by reading through the [first test](https://github.com/oxidecomputer/maghemite/blob/ry/mgd-v6-router-disco/falcon-lab/src/test.rs).

### DIFF
--- a/mgadm/src/bgp.rs
+++ b/mgadm/src/bgp.rs
@@ -13,7 +13,7 @@ use mg_admin_client::{
         Ipv6UnicastConfig, NeighborResetRequest,
     },
 };
-use rdb::types::{PeerId, PolicyAction, Prefix4, Prefix6};
+use rdb::types::{PeerId, Prefix4, Prefix6};
 use std::{
     fs::read_to_string,
     io::{Write, stdout},
@@ -136,8 +136,8 @@ pub struct HistorySubcommand {
 pub enum HistoryCmd {
     /// Get FSM event history for BGP sessions.
     Fsm {
-        /// Optional: Filter by specific peer address.
-        peer: Option<IpAddr>,
+        /// Optional: Filter by specific peer (IP address for numbered, interface name for unnumbered).
+        peer: Option<String>,
 
         /// Which buffer to display: 'major' (default) or 'all'.
         #[clap(default_value = "major")]
@@ -158,8 +158,8 @@ pub enum HistoryCmd {
 
     /// Get BGP message history for sessions.
     Message {
-        /// Peer address to show history for.
-        peer: IpAddr,
+        /// Peer to show history for (IP address for numbered, interface name for unnumbered).
+        peer: String,
 
         /// BGP Autonomous System number. Can be specified via ASN env var.
         #[clap(env)]
@@ -190,8 +190,8 @@ pub struct ClearSubcommand {
 pub enum ClearCmd {
     /// Clear the state of the selected BGP neighbor.
     Neighbor {
-        /// IP address of the neighbor you want to clear the state of.
-        addr: IpAddr,
+        /// Peer identifier (IP address for numbered, interface name for unnumbered).
+        peer: String,
 
         /// BGP Autonomous System number. Can be a 16-bit or 32-bit unsigned value.
         #[clap(env)]
@@ -496,25 +496,6 @@ pub struct RouterConfig {
     pub graceful_shutdown: bool,
 
     /// Autonomous system number for this router
-    #[clap(env)]
-    pub asn: u32,
-}
-
-#[derive(Args, Debug)]
-pub struct ExportPolicy {
-    /// Address of the peer to apply this policy to.
-    pub addr: IpAddr,
-
-    /// Prefix this policy applies to
-    pub prefix: Prefix4,
-
-    /// Priority of the policy, higher value is higher priority.
-    pub priority: u16,
-
-    /// The policy action to apply.
-    pub action: PolicyAction,
-
-    /// Autonomous system number for the router to add the export policy to.
     #[clap(env)]
     pub asn: u32,
 }
@@ -987,9 +968,9 @@ pub async fn commands(command: Commands, c: Client) -> Result<()> {
         Commands::Clear(cmd) => match cmd.command {
             ClearCmd::Neighbor {
                 asn,
-                addr,
+                peer,
                 operation,
-            } => clear_nbr(asn, addr, operation, c).await?,
+            } => clear_nbr(asn, peer, operation, c).await?,
         },
 
         Commands::Omicron(cmd) => match cmd.command {
@@ -1415,16 +1396,30 @@ async fn delete_nbr(asn: u32, peer: String, c: Client) -> Result<()> {
 
 async fn clear_nbr(
     asn: u32,
-    addr: IpAddr,
+    peer: String,
     operation: NeighborOperation,
     c: Client,
 ) -> Result<()> {
-    c.clear_neighbor_v2(&NeighborResetRequest {
-        asn,
-        addr,
-        op: operation.into(),
-    })
-    .await?;
+    match parse_peer_type(&peer) {
+        PeerType::Numbered(addr) => {
+            c.clear_neighbor_v2(&NeighborResetRequest {
+                asn,
+                addr,
+                op: operation.into(),
+            })
+            .await?;
+        }
+        PeerType::Unnumbered(interface) => {
+            c.clear_unnumbered_neighbor(
+                &types::UnnumberedNeighborResetRequest {
+                    asn,
+                    interface,
+                    op: operation.into(),
+                },
+            )
+            .await?;
+        }
+    }
     Ok(())
 }
 
@@ -1580,7 +1575,7 @@ async fn delete_shp(asn: u32, c: Client) -> Result<()> {
 async fn get_fsm_history(
     c: Client,
     asn: u32,
-    peer: Option<IpAddr>,
+    peer: Option<String>,
     buffer: &str,
     limit_str: &str,
     wide: bool,
@@ -1596,18 +1591,25 @@ async fn get_fsm_history(
         _ => "Major Events",
     };
 
+    // Parse peer filter if provided and convert to API type
+    let peer_id = peer.as_ref().map(|p| {
+        let bgp_peer_id: bgp::session::PeerId =
+            p.parse().expect("PeerId::from_str should always succeed");
+        peer_id_to_api(bgp_peer_id)
+    });
+
     let result = c
         .fsm_history_v2(&types::FsmHistoryRequest {
             asn,
-            peer: peer.map(PeerId::Ip),
+            peer: peer_id,
             buffer: buffer_type,
         })
         .await?
         .into_inner();
 
     if result.by_peer.is_empty() {
-        if let Some(peer_addr) = peer {
-            println!("No FSM history found for peer {}", peer_addr);
+        if let Some(peer_str) = peer {
+            println!("No FSM history found for peer {}", peer_str);
         } else {
             println!("No FSM history found for ASN {}", asn);
         }
@@ -1710,7 +1712,7 @@ async fn get_fsm_history(
 async fn get_message_history(
     c: Client,
     asn: u32,
-    peer: IpAddr,
+    peer: String,
     direction: &str,
     limit_str: &str,
     wide: bool,
@@ -1733,10 +1735,16 @@ async fn get_message_history(
         }
     };
 
+    // Parse peer and convert to API type
+    let bgp_peer_id: bgp::session::PeerId = peer
+        .parse()
+        .expect("PeerId::from_str should always succeed");
+    let peer_id = peer_id_to_api(bgp_peer_id);
+
     let result = c
         .message_history_v3(&types::MessageHistoryRequest {
             asn,
-            peer: Some(PeerId::Ip(peer)),
+            peer: Some(peer_id),
             direction: dir,
         })
         .await?
--- a/mgd/src/admin.rs
+++ b/mgd/src/admin.rs
@@ -13,7 +13,7 @@ use dropshot::{
 };
 use mg_api::*;
 use mg_common::stats::MgLowerStats;
-use rdb::{BfdPeerConfig, Db, Prefix};
+use rdb::{BfdPeerConfig, Db, PeerId, Prefix};
 use slog::{Logger, error, info, o};
 use std::collections::HashMap;
 #[cfg(feature = "mg-lower")]
@@ -378,10 +378,7 @@ impl MgAdminApi for MgAdminApiImpl {
     async fn get_exported_v2(
         ctx: RequestContext<Self::Context>,
         request: TypedBody<ExportedSelector>,
-    ) -> Result<
-        HttpResponseOk<HashMap<bgp::session::PeerId, Vec<Prefix>>>,
-        HttpError,
-    > {
+    ) -> Result<HttpResponseOk<HashMap<PeerId, Vec<Prefix>>>, HttpError> {
         bgp_admin::get_exported_v2(ctx, request).await
     }
 


ndp/src/packet.rs
```diff
@@ -74,7 +74,12 @@ impl Default for Icmp6RouterAdvertisement {
             checksum: 0,
             hop_limit: Self::DEFAULT_HOPLIMIT,
             flags: 0,
-            lifetime: 0, //indicates this is not a default router
+            // XXX arista routers will only peer with neighbors that have a
+            // non-zero lifetime. This is extremely bad behavior as a non-zero
+            // lifetime indicates the router is advertising itself as a default
+            // router. We need to plumb an arista-workaround option for this.
+            lifetime: 1800,
+            //lifetime: 0, //indicates this is not a default router
```
taspelund: Are we wanting to add a config option for this in this PR? if we address it in a later PR, we may have to introduce a new dropshot API version to accommodate it

falcon-lab/src/test.rs
```diff
@@ -0,0 +1,201 @@
...
+async fn frr_setup(r: FrrNode, d: std::sync::Arc<Runner>) -> Result<()> {
+    const BASE_CONFIG: &str = "
+        configure
+        ip forwarding
+        ip route 1.2.3.0/24 null0
+        route-map PERMIT-ALL permit 10
+        router bgp 44
+          timers bgp 2 6
+          neighbor enp0s8 interface remote-as external
+          neighbor enp0s8 capability extended-nexthop
```
taspelund: Explicitly enabling extended-nexthop isn't necessary for bgp unnumbered to work in frr.  ENHE is implicitly enabled for interface-based neighbors peering over v6 link-locals.

Interface-based config also works for ipv4 /30 or /31 subnets, and iirc ENHE is not implicitly enabled then.
Regardless of the v4 addressing on the interface, you can add the `v6-only` keyword to the "neighbor .. interface" command to ensure v6 link-locals are the only thing used for automatic peering.

mg-lower/src/dendrite.rs
```diff
@@ -249,6 +250,31 @@ where
...
+                    dpd_log!(log,
+                        error,
+                        "failed to create route {r:?} {e}";
+                        "error" => format!("{e}")
+                    );
+                    return Err(e.into());
+                }
+            }
             _ => {
                 // XXX: re-evaluate for RFC 8950 (BGP unnumbered) support
```
taspelund: Still needed?

mgadm/src/bgp.rs
```diff
@@ -693,19 +765,95 @@ impl From<Neighbor> for types::Neighbor {
...
+            connect_retry_jitter: n.connect_retry_jitter,
+            idle_hold_jitter: n.idle_hold_jitter,
+            deterministic_collision_resolution: n
+                .deterministic_collision_resolution,
+        }
+    }
+}
+
+impl From<UnnumberedNeighbor> for types::UnnumberedNeighbor {
+    fn from(n: UnnumberedNeighbor) -> types::UnnumberedNeighbor {
```
taspelund: Would we be able to avoid any of this boilerplate by including UnnumberedNeighbor in the dropshot `replace` directive in mg-admin-client/src/lib.rs?

mgd/src/bgp_admin.rs
```diff
@@ -1397,6 +1508,92 @@ pub(crate) mod helpers {
         Ok(())
     }
 
+    pub(crate) fn add_unnumbered_neighbor(
+        ctx: Arc<HandlerContext>,
+        rq: UnnumberedNeighbor,
+        _ensure: bool, //TODO or just always do ensure mode since this is async?
```
taspelund: Would `ensure` for an unnumbered neighbor just mean that the one-shot unnumbered manager request would be delayed until "start"? or do you have a different intent behind `ensure`?

mgd/src/error.rs
```diff
@@ -18,6 +19,12 @@ pub enum Error {
     #[error("bgp error: {0}")]
     Bgp(#[from] bgp::error::Error),
 
+    #[error("error adding an unnumbered error: {0}")]
+    AddUnnumberedNeighbor(#[from] AddNeighborError),
+
+    #[error("error resolving a neighbor: {0}")]
+    ResolveNeighbor(#[from] ResolveNeighborError),
```
taspelund: Should this be `ResolveUnnumberedNeighbor`, or did you want this to be generic enough to work for both numbered and unnumbered?  I don't have strong feelings, just wondering about the intention

mgd/src/error.rs
```diff
@@ -18,6 +19,12 @@ pub enum Error {
...
+
```
taspelund: nit: Do we want these as separate error classes in mgd? Or would it make more sense to tuck these both under an UnnumberedManager variant?

mgd/src/unnumbered_manager.rs
```diff
@@ -0,0 +1,268 @@
...
+            log: log.clone(),
+        });
+
+        {
+            let s = s.clone();
+            let log = log.clone();
+            spawn(move || s.run(log));
+        }
+
+        s
```
taspelund: Can we use the `ManagedThread` type/pattern for this?  That makes it a bit more ergonomic if we ever want to do something with the JoinHandle or if we want to be able to report the state of mgd's threads

mgd/src/unnumbered_manager.rs
```diff
@@ -0,0 +1,268 @@
...
+                interface: ifx,
+            },
+            NbrInfo { session: info, nbr },
+        );
+        Ok(())
+    }
+
+    pub fn remove_neighbor(
+        self: &Arc<Self>,
+        asn: u32,
```
taspelund: Probably not something to fix in this PR, but something to consider:

Do we need the asn at all for neighbor deletions? (unnumbered or otherwise)
We already require the interface/ip of the peer to be unique, so the asn isn't acting as a required discriminator

mgd/src/unnumbered_manager.rs
```diff
@@ -0,0 +1,268 @@
...
+
+use bgp::{
+    connection_tcp::BgpConnectionTcp,
+    params::UnnumberedNeighbor,
+    router::Router,
+    session::{SessionInfo, SessionRunner},
+};
+use ndp::{Ipv6NetworkInterface, NdpManager, NewInterfaceNdpManagerError};
+use rdb::Db;
+use slog::{Logger, error, warn};
```
taspelund: nit: I'd like to see the use of slog KV pairs, e.g. unit, component, etc. but I don't think it's a blocker

ndp/src/manager.rs
```diff
@@ -0,0 +1,285 @@
...
+}
+
+/// The `InterfaceNdpManager` runs router discovery for an individual interface.
+///
+/// Discovery is started on construction and stopped when the interface manager
+/// object is dropped.
+#[derive(Debug)]
+pub struct InterfaceNdpManager {
+    tx_thread: Option<JoinHandle<()>>,
+    rx_thread: Option<JoinHandle<()>>,
```
taspelund: Maybe we can reuse `ManagedThread` here too?

ndp/src/manager.rs
```diff
@@ -0,0 +1,285 @@
...
+impl InterfaceNdpManagerInner {
+    /// Run the interface NDP manager receive loop. Advertisements are used to
+    /// set the current peer address. Advertisements are sent in response to
+    /// solicitations.
+    ///
+    /// A read timeout of 1 second is used. When the time out hits instead of
+    /// receiving a advertisement or solicitation packet, the current neighbor
+    /// (if any) is checked for expiration.
+    pub fn rx_loop(&self, s: Socket) {
+        const INTERVAL: Duration = Duration::from_secs(1);
```
taspelund: nit: why are we defining the two interval const's in two different spots? why not just tuck them both under `impl InterfaceNdpManagerInner`?

ndp/src/manager.rs
```diff
@@ -0,0 +1,285 @@
...
+            let mut buf: [MaybeUninit<u8>; 1024] =
+                unsafe { MaybeUninit::uninit().assume_init() };
+
+            match s.recv_from(&mut buf) {
+                Ok((len, src)) => {
+                    let buf: &[u8] = unsafe {
+                        std::slice::from_raw_parts(buf.as_ptr().cast(), len)
+                    };
+                    let Some(src) = src.as_socket_ipv6().map(|x| *x.ip())
+                    else {
```
taspelund: Do we need to add handling here for an EOF read? i.e. len == 0

ndp/src/util.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+}
+
+pub fn cksum(
+    src: Ipv6Addr,
+    dst: Option<Ipv6Addr>,
+    ulp_len: u32,
+    data: &mut [u8],
+) {
+    const ICMP6_NEXT_HDR: u8 = 58;
```
taspelund: Can we add a comment explaining this is the IP Protocol number for ICMP6?  I just googled a bunch of things trying to see if there was a Next Header field in the ICMP6 header without realizing this is a field in the IP packet.

ndp/src/util.rs
```diff
@@ -0,0 +1,193 @@
...
+    s.set_multicast_if_v6(index).map_err(E::SetMulticastIf)?;
+
+    s.set_multicast_hops_v6(255)
+        .map_err(E::SetMulticastHopsV6)?;
+
+    s.set_multicast_loop_v6(false)
+        .map_err(E::SetMulticastLoop)?;
+
+    s.join_multicast_v6(&ALL_NODES_MCAST, index)
+        .map_err(E::JoinAllNodesMulticast)?;
```
taspelund: We don't also have to join the all routers group?  Not necessarily suggesting a change, just curious

taspelund:
Thanks for all this work, Ry!

I've left individual comments above, but overall I'd say it looks good.
The only places I have some actual concerns are in enabling ENHE capability by default (I want to make sure we don't break anything for numbered peers), and in the HashMap that rdb is using to store its unnumbered next-hops (I think we should map interface -> address rather than address -> interface, that way we can handle something like `fe80::10%qsfp0` and `fe80::10%qsfp1` at the same time).

ndp/src/manager.rs
```diff
@@ -0,0 +1,285 @@
...
+        const INTERVAL: Duration = Duration::from_secs(1);
+
+        loop {
+            if self.stop.load(Ordering::SeqCst) {
+                break;
+            }
+            let _ds = DropSleep(INTERVAL);
+
+            let mut buf: [MaybeUninit<u8>; 1024] =
+                unsafe { MaybeUninit::uninit().assume_init() };
```
taspelund: What's the motivation for `uninit().assume_init()`?
The MaybeUninit [docs](https://doc.rust-lang.org/beta/std/mem/union.MaybeUninit.html) state this as UB in one of the first examples:
```
use std::mem::{self, MaybeUninit};

let x: i32 = unsafe { mem::uninitialized() }; // undefined behavior! ⚠️
// The equivalent code with `MaybeUninit<i32>`:
let x: i32 = unsafe { MaybeUninit::uninit().assume_init() }; // undefined behavior! ⚠️
```

falcon-lab/src/util.rs
```diff
@@ -0,0 +1,24 @@
+#[macro_export]
+macro_rules! wait_for_eq {
```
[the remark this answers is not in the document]
rcgoodfellow: Yes. `wait_for_eq` is materially different in these tests, we use errors instead of assert in place.

falcon-lab/src/test.rs
```diff
@@ -190,3 +191,34 @@ async fn frr_setup(r: FrrNode, d: std::sync::Arc<Runner>) -> Result<()> {
...
     Ok(())
 }
+
+async fn eos_setup(r: EosNode, d: Arc<Runner>) -> Result<()> {
+    const BASE_CONFIG: &str = "
+        enable
+        configure
+        ipv6 unicast-routing
+        ip routing ipv6 interfaces
+        ip routing
```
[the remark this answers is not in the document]
rcgoodfellow: The very sparse documentation I've seen with EOS has them both. I've already had a large fight with EOS getting unnumbered to work at all and I don't want to spend much more time polishing this turd.

ndp/src/packet.rs
rcgoodfellow: This is still a TODO for this PR. We should talk live about what we want to do.

.github/buildomat/jobs/falcon-lab.sh
```diff
@@ -0,0 +1,58 @@
...
+
+if [[ $(curl -s http://catacomb.eng.oxide.computer:12346/trim-me) =~ "true" ]]; then
+		pfexec zpool trim cpool
+		while [[ ! $(zpool status -t cpool) =~ "100%" ]]; do sleep 10; done
+fi
+
+pfexec chown "$UID" /ci
+cd /ci
+export FALCON_DATASET="cpool/falcon"
+
```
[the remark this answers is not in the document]
rcgoodfellow: We could do that on a regular CI job that does not do skip clone, but we don't have a whole lot to work with for these types of jobs. We should really turn this into a rust program and just have the buildomat job be close to a one-liner.

mg-lower/src/dendrite.rs
```diff
@@ -249,6 +250,31 @@ where
...
+                    vlan_id,
+                };
+
+                let update = types::Ipv4OverIpv6RouteUpdate {
+                    cidr: c,
+                    target,
+                    replace: false,
+                };
+                if let Err(e) = rt.block_on(async {
+                    dpd.route_ipv4_over_ipv6_add(&update).await
```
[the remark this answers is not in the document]
rcgoodfellow: I think that's more of a dendrite design question for

- https://github.com/oxidecomputer/dendrite/pull/181

mgd/src/bgp_admin.rs
```diff
@@ -1670,12 +1895,17 @@ mod tests {
         let log =
             mg_common::log::init_file_logger("apply_remove_entire_group.log");
 
+        let db = Db::new(dbdir.as_str(), log.clone()).unwrap();
```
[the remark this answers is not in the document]
rcgoodfellow: Done

mgd/src/unnumbered_manager.rs
rcgoodfellow: The design of the unnumbered manager right now is that its not really managed. It runs for the lifetime of the process and has no shutdown signal semantics. Would you still want to use `ManagedThread` for something like this?

mgd/src/unnumbered_manager.rs
```diff
@@ -0,0 +1,268 @@
...
+        let mut local: Vec<_> = candidates
+            .into_iter()
+            .filter_map(|x| x.addr.map(|addr| (addr, x.index)))
+            .filter_map(|(addr, idx)| match addr.ip() {
+                IpAddr::V6(ip) if ip.is_unicast_link_local() => Some((ip, idx)),
+                _ => None,
+            })
+            .collect();
+
+        let Some((addr, index)) = local.pop() else {
```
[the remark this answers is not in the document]
rcgoodfellow: I wanted to at least have a warning in the logs if this situation occurs as we could land in a very unexpected place in terms of behavior. I don't know what controls are available for this at the OS level, but if such controls are available, we would need to use a more os-focused networking crate like `netadm-sys` to take advantage of that rather than `network-interfaces`.

mgd/src/unnumbered_manager.rs
```diff
@@ -0,0 +1,268 @@
...
+            ip: addr,
+            index,
+        })
+    }
+
+    pub fn get_pending(self: &Arc<Self>) -> HashMap<NbrKey, NbrInfo> {
+        self.pending_sessions.lock().unwrap().clone()
+    }
+
+    fn run(self: Arc<Self>, log: Logger) {
```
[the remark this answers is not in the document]
rcgoodfellow: I've added a log item for the unnumbered manager loop starting. It's not possible to add one for it stopping because it never stops. The compiler recognizes this and fires of an unreachable code warning.

mgd/src/unnumbered_manager.rs
```diff
@@ -0,0 +1,268 @@
...
+    time::Duration,
+};
+
+pub struct UnnumberedNeighborManager {
+    routers: Arc<Mutex<BTreeMap<u32, Arc<Router<BgpConnectionTcp>>>>>,
+    ndp_mgr: Arc<NdpManager>,
+    pending_sessions: Mutex<HashMap<NbrKey, NbrInfo>>,
+    db: Db,
+    log: Logger,
+}
```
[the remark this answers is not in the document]
rcgoodfellow: The intent here was to lay out all the structures up front and then get into impl blocks below to try to make the overall structure easier to read.

ndp/src/manager.rs
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    /// Add an interface to the NDP manager. Discovery starts immediately.
+    pub fn add_interface(
+        &self,
+        ifx: Ipv6NetworkInterface,
+    ) -> Result<(), NewInterfaceNdpManagerError> {
+        self.interfaces
+            .write()
+            .unwrap()
```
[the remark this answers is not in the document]
rcgoodfellow: Done and throughout this file

ndp/src/manager.rs
```diff
@@ -0,0 +1,285 @@
...
+            return false;
+        };
+
+        ifxs_guard.remove(pos);
+        true
+    }
+
+    /// Get a router peer, if any, that has been discovered for the given interface.
+    pub fn get_peer(&self, ifx: &Ipv6NetworkInterface) -> Option<Ipv6Addr> {
+        let ifxs_guard = self.interfaces.read().unwrap();
```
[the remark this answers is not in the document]
rcgoodfellow: Done and throughout this file

ndp/src/manager.rs
rcgoodfellow: I was trying to contain this to the smallest possible scope. To that end i've moved RA_INTERVAL to be a function local constant in tx_loop.

ndp/src/manager.rs
```diff
@@ -0,0 +1,285 @@
...
+                        continue;
+                    }
+                    error!(self.log, "rx: {e}");
+                }
+            }
+        }
+    }
+
+    /// Start the transmit loop, periodically sending out announcements every
+    /// five seconds.
```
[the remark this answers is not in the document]
rcgoodfellow: Agreed. Removed.

ndp/src/util.rs
rcgoodfellow: Done

ndp/src/manager.rs
rcgoodfellow: I've tightened this up a bit.

bgp/src/messages.rs
```diff
@@ -3936,10 +3972,14 @@ pub enum Capability {
     /// (deprecated). Note this capability is not yet implemented.
     MultipleRoutesToDestination {},
 
-    //TODO
     /// Multiple nexthop encoding capability as defined in RFC 8950. Note this
-    /// capability is not yet implemented.
-    ExtendedNextHopEncoding {},
+    /// is for routing v4 over link-local v6 nexthops.
```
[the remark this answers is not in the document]
rcgoodfellow: Done

mgd/src/unnumbered_manager.rs
taspelund: Even if we don't currently do an explicit join, I think it would be better to hold onto a handle for the thread so it's simpler to update later if need be.  At some point I'd like to add a `mgadm threads show` sort of command to return a summary of each thread state (possibly with panic counters and/or cpu runtime stats), and I think ManagedThread would be a simple way to start that process.

I'm also not married to the term "managed" in the type name, so if you don't feel it fits then I'm ok with a more fitting name.

falcon-lab/src/util.rs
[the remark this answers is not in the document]
taspelund: Maybe we can put an err variant of wait_for_eq into mg-common for all crates to use?

bgp/src/messages.rs
```diff
@@ -566,7 +566,17 @@ impl OpenMessage {
...
+                    Capability::FourOctetAs { asn },
+                    // Default to supporting v4 over v6 encoding.
+                    Capability::ExtendedNextHopEncoding {
+                        elements: vec![ExtendedNexthopElement {
+                            afi: Afi::Ipv4.into(),
+                            safi: u8::from(Safi::Unicast).into(),
+                            nh_afi: Afi::Ipv6.into(),
+                        }],
+                    },
+                ],
```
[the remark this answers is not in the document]
rcgoodfellow: For this PR we'll advertise extended nexthop for unnumbered sessions only and revisit broader support later.

bgp/src/messages.rs
```diff
@@ -3900,6 +3909,33 @@ impl Display for AddPathElement {
...
+impl Display for ExtendedNexthopElement {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "safi={}/afi={}/nh_afi={}",
+            self.afi, self.safi, self.nh_afi
+        )
+    }
+}
+
```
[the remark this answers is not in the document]
rcgoodfellow: Done

mgd/src/unnumbered_manager.rs
rcgoodfellow: Add a pointer to
- https://github.com/oxidecomputer/maghemite/issues/609

ndp/src/packet.rs
rcgoodfellow: This is now done.

PR #638, opened 2026-02-11
R18 again

### LINKED ISSUE #634: large update message regression
I've got Maghemite post #584 up in the lab with a juniper router slinging 4k routes at it. This causes the session to fall over due to TCP errors. More details to come later when I unwind some other things going on in the lab.

taspelund:
Couple important fixes for R18 plus a couple mgadm quality of life fixes.

Important fixes:
- add handling for `ErrorKind::WouldBlock` in recv_msg() to avoid tearing down TCP session when a read returns `EWOULDBLOCK`/`EAGAIN` for a non-blocking socket (this fixes #634) -- when moving the TCP socket to non-blocking, this read call was missed when adding error handling for WouldBlock.
- fix RA router lifetime (arista workaround) update path -- NDP manager interface config would be updated, but the tx loop would never restart or re-read the config. Now it stores the lifetime in an AtomicU16 that can be updated without restarting the tx loop.

mgadm:
- replace humantime duration formatting with more readable formatter
- re-order ASN to final positional arg in NDP commands so environ works

Other:
- add validation of rx BGP message length
~- set TCP_NODELAY on BGP sockets~

details for all changes are in the individual commit messages

### DIFF
--- a/bgp/src/connection_tcp.rs
+++ b/bgp/src/connection_tcp.rs
@@ -536,8 +536,6 @@ impl BgpConnectionTcp {
         direction: ConnectionDirection,
         config: &SessionInfo,
     ) -> Result<Self, Error> {
-        conn.set_nodelay(true)?;
-
         let id = ConnectionId::new(source, peer);
 
         let dropped = Arc::new(AtomicBool::new(false));


bgp/src/connection_tcp.rs
```diff
@@ -535,6 +535,8 @@ impl BgpConnectionTcp {
         direction: ConnectionDirection,
         config: &SessionInfo,
     ) -> Result<Self, Error> {
+        conn.set_nodelay(true)?;
```
rcgoodfellow: what prompted this?

Issue #649, opened 2026-03-02
Silent path loss during RIB update for distinct peers with identical path attributes


taspelund:
When multiple BGP unnumbered peers exist and have the same link-local IP, it is possible for insertion of a RIB path to overwrite an existing path instead of being inserted alongside it.  This was noticed at a customer site where 6 BGP unnumbered peers were established and advertising us default routes (v4 and v6), but `mgadm rib status imported` showed only 1 or 2 paths instead of 6.

When digging into this further, I found that the issue is in the `Ord` impl for the rdb Path type.
The Ord impl looks like this:
```
impl Ord for Path {
    fn cmp(&self, other: &Self) -> Ordering {
        // Paths from the same source are considered equal for set membership.
        // This enables BTreeSet::replace() to update paths from the same source.
        //
        // BGP paths: identified by peer IP
        if let (Some(a), Some(b)) = (&self.bgp, &other.bgp)
            && a.peer == b.peer
        {
            return Ordering::Equal;
        }

        // Static paths: identified by (nexthop, nexthop_interface, vlan_id)
        if self.bgp.is_none()
            && other.bgp.is_none()
            && self.nexthop == other.nexthop
            && self.nexthop_interface == other.nexthop_interface
            && self.vlan_id == other.vlan_id
        {
            return Ordering::Equal;
        }
        if self.nexthop != other.nexthop {
            return self.nexthop.cmp(&other.nexthop);
        }
        if self.shutdown != other.shutdown {
            return self.shutdown.cmp(&other.shutdown);
        }
        if self.rib_priority != other.rib_priority {
            return self.rib_priority.cmp(&other.rib_priority);
        }
        self.bgp.cmp(&other.bgp)
    }
}

impl PartialOrd for BgpPathProperties {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(self.cmp(other))
    }
}
impl Ord for BgpPathProperties {
    fn cmp(&self, other: &Self) -> Ordering {
        if self.origin_as != other.origin_as {
            return self.origin_as.cmp(&other.origin_as);
        }
        if self.id != other.id {
            return self.id.cmp(&other.id);
        }
        // MED should *not* be used as a basis for comparison. Paths with
        // distinct MED values are not distinct paths.
        if self.as_path != other.as_path {
            return self.as_path.cmp(&other.as_path);
        }
        self.stale.cmp(&other.stale)
    }
}
```

[...]

PR #650, opened 2026-03-02
rdb: fix silent path loss in Path Ord impl

taspelund:
The Path Ord implementation had a fallback chain that compared fields like nexthop, shutdown, rib_priority, then delegated to BgpPathProperties::Ord (which compared origin_as, id, as_path, stale). When two distinct BGP paths matched on all those fields, Ord returned Equal and BTreeSet silently dropped one — affecting rib_in, bestpath selection, and rib_loc.

For example, two unnumbered BGP sessions to the same router on different interfaces share the same link-local nexthop, router ID, AS, and AS path — all fields the old Ord compared. A router with sessions on eth0 and eth1 both receiving 10.0.0.0/24 from AS 65000 would silently lose one of those paths.

This replaces the fallback chain with a match on path type that cleanly separates identity from attributes:

  - BGP path identity: PeerId only
  - Static path identity: (nexthop, nexthop_interface, vlan_id)
  - Cross-type: BGP sorts after static (Some > None)

All other fields (med, local_pref, origin_as, as_path, shutdown, rib_priority, etc.) are attributes carried by the path, not identity. They get updated via BTreeSet::replace() without affecting set membership.

Remove Ord/PartialOrd from BgpPathProperties since nothing compares it directly — Path::Ord extracts the peer field itself.

Fixes: #649

### DIFF
--- a/rdb/src/db.rs
+++ b/rdb/src/db.rs
@@ -2253,10 +2253,8 @@ mod test {
         let db = get_test_db();
 
         // --- IPv4 static path ---
-        let nexthop4 =
-            IpAddr::V4(Ipv4Addr::from_str("198.51.100.1").unwrap());
-        let prefix4 =
-            Prefix4::new(Ipv4Addr::from_str("10.0.0.0").unwrap(), 24);
+        let nexthop4 = IpAddr::V4(Ipv4Addr::from_str("198.51.100.1").unwrap());
+        let prefix4 = Prefix4::new(Ipv4Addr::from_str("10.0.0.0").unwrap(), 24);
         let static_key4 = StaticRouteKey {
             prefix: Prefix::V4(prefix4),
             nexthop: nexthop4,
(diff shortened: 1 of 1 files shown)


rdb/src/db.rs
rcgoodfellow: I think this should also be

```rust
paths.replace(replacement)
```

rcgoodfellow (commented):
I think this broadly looks good. A few high level comments.

I think this will improve the situation where we get the exact same route from two different peers, so if one withdraws, we still have the same route from the other peer in the imported part of the RIB. Which is what we want, not both routes disappearing in that situation.

It does raise a question about whether we now need to dedup in bestpath. Because we can now get the exact same routes from different peers and they won't collapse through `Ord` should we be eliminating duplicates so they don't skew path diversity for bestpath.

rcgoodfellow (approved):

PR #675, opened 2026-03-25
MRIB: Multicast RIB implementation

zeeshanlakhani:
Implements the Multicast Routing Information Base (MRIB) for multicast support. The MRIB follows a two-table architecture (`mrib_in` → `mrib_loc`) with RPF verification against the unicast RIB when sources are provided.

This PR includes:
  - `rdb/src/mrib/mod.rs`: core MRIB with route storage and change notifications
  - `rdb/src/mrib/rpf.rs`: RPF verification using poptrie for O(1) LPM lookups. A single long-lived worker rebuilds the cache on unicast loc-RIB change(s), coalescing mutation bursts through a bounded pending slot, and the cache is eventually consistent with the RIB
  - `rdb/src/db.rs`: MRIB persistence with sled transactions. RPF rebuild triggers ride the existing change-notification path
  - `rdb/src/types.rs`: validated multicast types (`UnderlayMulticastIpv6`, `Vni`)
  - `mg-api/src/lib.rs`: API v12 (`VERSION_MULTICAST_SUPPORT`) endpoints, following RFD 619 pattern (atop #665)
  - `mg-api-types/versions/src/multicast_support`: versioned API types
  - `mgd/src/mrib_admin.rs`: HTTP handlers bridging API to MRIB
  - `mgadm/src/mrib.rs`: read-only CLI for MRIB inspection (Omicron programs the MRIB via mg-api)

Omicron is the source of truth for multicast overlay/underlay groups/addresses.


rdb/src/error.rs
```diff
@@ -13,6 +15,9 @@ pub enum Error {
     #[error("serialization error {0}")]
     Serialization(#[from] serde_json::Error),
 
+    #[error("io error {0}")]
+    Io(#[from] std::io::Error),
+
```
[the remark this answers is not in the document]
zeeshanlakhani: Good catch, `From` Impl not letting clippy say anything here.

zeeshanlakhani:
@taspelund all the stuff on main has been merged in btw.

mg-api-types/versions/src/multicast_support/mrib.rs
```diff
@@ -0,0 +1,1172 @@
...
+///
+/// Virtual Network Identifiers are constrained to 24-bit values per the
+/// Geneve specification (RFC 8926 Section 3.3).
+pub const MAX_VNI: u32 = 0xFF_FFFF;
+
+/// Default VNI for fleet-wide multicast routing.
+///
+/// A low-numbered VNI chosen to avoid colliding with user VNIs, though it is
+/// not yet within the Oxide-reserved range.
+pub const DEFAULT_MULTICAST_VNI: u32 = 77;
```
taspelund: Why not make these associated consts of the Vni type?

mg-api-types/versions/src/multicast_support/mrib.rs
```diff
@@ -0,0 +1,1172 @@
...
+    /// assert!(Vni::new(MAX_VNI + 1).is_err());
+    /// ```
+    pub fn new(value: u32) -> Result<Self, MulticastError> {
+        if value > MAX_VNI {
+            return Err(MulticastError::Validation(format!(
+                "VNI {value} exceeds the maximum 24-bit value {MAX_VNI}"
+            )));
+        }
+        Ok(Self(value))
+    }
```
taspelund: It seems wrong/confusing to return `MulticastError` when the broken invariant is exceeding a 24-bit value that's wholly unrelated to multicast. Can we maybe add something like a `VniError` type that's better suited for this?

mg-api-types/versions/src/multicast_support/mrib.rs
```diff
@@ -0,0 +1,1172 @@
...
+
+    /// Return the underlying 24-bit value.
+    #[inline]
+    pub const fn as_u32(self) -> u32 {
+        self.0
+    }
+}
```
taspelund, line 116: Maybe worth an `impl Deref`?

mgd/src/mrib_admin.rs
```diff
@@ -0,0 +1,151 @@
...
+fn origin_to_static_only(origin: Option<RouteOriginFilter>) -> Option<bool> {
+    match origin {
+        None => None,
+        Some(RouteOriginFilter::Static) => Some(true),
+        Some(RouteOriginFilter::Dynamic) => Some(false),
+    }
+}
+
+pub async fn get_mrib_imported(
+    rqctx: RequestContext<Arc<HandlerContext>>,
```
taspelund: It seems like this and get_mrib_selected could be collapsed into a single function with a boolean/enum argument

mg-api-types/versions/src/multicast_support/mrib.rs
zeeshanlakhani: Good point.

mg-api-types/versions/src/multicast_support/mrib.rs
zeeshanlakhani: It's not typical to `Deref` here because the [Rust API guidelines](https://rust-lang.github.io/api-guidelines/predictability.html#only-smart-pointers-implement-deref-and-derefmut) reserve `Deref`/`DerefMut` for smart pointers. Explicit escapes already cover conversion: `as_u32()` (const) and `From<Vni> for u32`.

mg-api-types/versions/src/multicast_support/mrib.rs
```diff
@@ -0,0 +1,1175 @@
...
+    PartialOrd,
+    Ord,
+    Hash,
+    Serialize,
+    Deserialize,
+    JsonSchema,
+)]
+#[serde(try_from = "u32", into = "u32")]
+#[schemars(transparent)]
+pub struct Vni(u32);
```
taspelund, line 74: I probably should have clocked this before, but I don't think VNI is specific to multicast or mrib at all. Could this live in a different file/module with a name that is more reflective of the generic nature of a VNI?

zeeshanlakhani: There's a reason: `multicast_support` is the version-named module per the API-versioning convention (types live in the version that introduced them), and the definition is already hoisted to `client-common` ‚Äî this file only re-exports it. We want the oxnet consolidation anyway (TODO in `vni.rs`).

PR #682, opened 2026-03-26
happy maps

taspelund:
Updates some of the map data structures we use in mgd/bgp.  In particular:
1. Moves `UnnumberedManager` trait impls over to using `iddqd::BiHashMap` for the interface name to scope id mapping. This ensures we only have to maintain a single map (instead of 2 like what was used by UnnumberedManagerMock) and that we get consistent O(1) lookups (instead of the O(n) lookups in the UnnumberedManagerNdp impl).
2. Moves `Router.sessions` over to a new SessionMap type which wraps an `iddqd::IdOrdMap`, which ensures we can't ever insert a SessionRunner into a map behind an invalid PeerId key since it is derived from the SessionRunner being inserted. There was not a known issue with the existing impl, but this provides some more type safety to prevent anything bad from happening in the future.
3. Removes the peer_to_session (previously known as addr_to_session prior to bgp unnumbered) entirely. This was providing a way to get the config (SessionInfo) and handle for sending FSM events (event_tx) for a given peer, primarily in service of the Dispatcher. Since the config and FSM handle are just extensions of the SessionRunner itself, this could all be queried using the global SessionMap (item 2 above). This consolidates us down to just 1 mapping of Peer to SessionRunner instead of 2 where one provides a subset of the others' functionality.

### DIFF
--- a/bgp/src/router.rs
+++ b/bgp/src/router.rs
@@ -70,8 +70,13 @@ impl<Cnx: BgpConnection + 'static> SessionMap<Cnx> {
         self.0.get(peer).map(|h| &h.0)
     }
 
-    pub fn insert(&mut self, session: Arc<SessionRunner<Cnx>>) {
-        self.0.insert_overwrite(SessionHandle(session));
+    /// Inserts `session`, overwriting any existing entry for the same
+    /// `PeerId`. Returns the displaced session if one was present.
+    pub fn insert_overwrite(
+        &mut self,
+        session: Arc<SessionRunner<Cnx>>,
+    ) -> Option<Arc<SessionRunner<Cnx>>> {
+        self.0.insert_overwrite(SessionHandle(session)).map(|h| h.0)
     }
 
     pub fn remove(&mut self, peer: &PeerId) -> Option<Arc<SessionRunner<Cnx>>> {
@@ -442,7 +447,7 @@ impl<Cnx: BgpConnection + 'static> Router<Cnx> {
             unnumbered_manager,
         ));
 
-        sessions.insert(runner.clone());
+        sessions.insert_overwrite(runner.clone());
         drop(sessions);
 
         self.spawn_session_thread(runner.clone());
--- a/bgp/src/router.rs
+++ b/bgp/src/router.rs
@@ -19,6 +19,7 @@ use crate::{
     },
     unnumbered::UnnumberedManager,
 };
+use iddqd::{IdOrdItem, IdOrdMap, id_upcast};
 use mg_api_types::rdb::prefix::{Prefix4, Prefix6};
 use mg_common::{lock, read_lock, write_lock};
 use rdb::{Asn, Db};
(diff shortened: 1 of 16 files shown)


taspelund:
I have this in a4x2 currently and BGP unnumbered peers have all come up, but nexus hasn't fully come online. I don't suspect the issues are related to these changes, but I will resume troubleshooting this tomorrow.

taspelund:
Rebased atop main after #724

bgp/src/router.rs
```diff
@@ -35,6 +35,72 @@ use std::{
...
+impl<Cnx: BgpConnection + 'static> SessionMap<Cnx> {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn get(&self, peer: &PeerId) -> Option<&Arc<SessionRunner<Cnx>>> {
+        self.0.get(peer).map(|h| &h.0)
+    }
+
+    pub fn insert(&mut self, session: Arc<SessionRunner<Cnx>>) {
```
rcgoodfellow: Should probably document that this is an overwriting insert and consider returning the optional overwritten entry.

bgp/src/router.rs
```diff
@@ -446,7 +492,6 @@ impl<Cnx: BgpConnection + 'static> Router<Cnx> {
 
     pub fn delete_session(&self, peer: impl Into<PeerId>) {
         let peer_id = peer.into();
-        lock!(self.peer_to_session).remove(&peer_id);
```
rcgoodfellow, line 449: So the idea here is that `peer_to_session` and `self.sessions` are now effectively collapsed into `self.sessions` so the remaining lock!(self.sessions).remove` below is sufficient?

bgp/src/unnumbered.rs
```diff
@@ -2,9 +2,79 @@
...
+    }
+}
+
+impl ScopeMap {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn insert(&mut self, scope_id: u32, interface: String) {
+        self.0.insert_overwrite(ScopeEntry {
```
rcgoodfellow: Similar to the prior comment, making it clear this is an overwriting insert in documentation and consider returning the optional overwritten entry.

mgd/src/oxstats.rs
```diff
@@ -284,7 +284,7 @@ impl Stats {
         for (asn, r) in &*routers {
             let mut session_counters = BTreeMap::new();
             let sessions = lock!(r.sessions);
-            for (key, session) in &*sessions {
+            for (key, session) in sessions.iter() {
```
rcgoodfellow, line 287: nice

mgd/src/unnumbered_manager.rs
```diff
@@ -325,14 +327,8 @@ impl UnnumberedManagerNdp {
...
-        // Clean up scope mapping by searching for interface name
-        let mut scope_map = lock!(self.interface_scope_map);
-        if let Some((&scope_id, _)) = scope_map
-            .iter()
-            .find(|(_, name)| name.as_str() == interface_name)
-        {
-            scope_map.remove(&scope_id);
-        }
+        // Clean up scope mapping
+        lock!(self.interface_scope_map).remove_by_interface(interface_name);
```
rcgoodfellow, line 331: nice simplification for call sites here and below at 426

rcgoodfellow (commented):
Thanks @taspelund! Mostly looks good to me and code is definitely improved with the wrappers around iddqd maps.

bgp/src/router.rs
taspelund: Yes, precisely

taspelund:
@rcgoodfellow your feedback has been integrated. I've added doc comments to the insertion methods and renamed them to be more clear at call sites. Feel free to re-review whenever you're ready

rcgoodfellow (approved):

PR #689, opened 2026-03-30
Add state duration to DDM FSM

### LINKED ISSUE #54: ddm: track peer establishment time

taspelund:
At its heart, this is a re-implementation of #180 which has been updated for the modern codebase and offers some cleanup in addition to timestamps.

First and foremost, this PR adds duration tracking for DDM FSM states. It does so by updating the FSM state enum variants to hold a std::time::Instant representing the timestamp when the FSM entered that state, and calculating a Duration using .elapsed() when the FSM state is queried (i.e. when `ddmadm get-peers` is called).

Part of this work was to cleanup the implementation of peer tracking: namely getting rid of the Db.peers table and instead tracking a peer in each FSM. Since the FSMs are instantiated per interface, this naturally gives us the interface-to-peer mapping structure so we don't actually need to maintain a separate table for the peers. Instead, we update/query the peers directly from the FSMs.

There are also a couple quality of life improvements: putting the state duration in a separate column and sorting the output of `ddmadm get-peers`.

Fixes: #54


taspelund:
rebased atop main after #724

ddm/src/admin.rs
```diff
@@ -113,7 +113,24 @@ impl DdmAdminApi for DdmAdminApiImpl {
...
     ) -> Result<HttpResponseOk<HashMap<u32, PeerInfo>>, HttpError> {
         let ctx = lock!(ctx.context());
-        Ok(HttpResponseOk(ctx.db.peers()))
+        let mut result = HashMap::new();
+        for sm in &ctx.peers {
+            let if_index = *lock!(sm.iface.if_index);
+            let Some(peer) = lock!(sm.iface.peer_identity).clone() else {
+                continue;
+            };
+            let status = sm.iface.peer_status();
```
rcgoodfellow: Consider putting this before the two locking statements above, in case `peer_status()` decides one day it wants to lock those things itself.

ddm/src/oxstats.rs
```diff
@@ -154,7 +157,7 @@ impl Producer for Stats {
                 self.hostname.clone().into(),
                 self.rack_id,
                 self.sled_id,
-                peer.config.if_name.clone().into(),
+                lock!(peer.iface.if_name).clone().into(),
```
rcgoodfellow: Can probably just lock and copy out once, dropping the lock immediately for all these calls so we only have to lock once per peer.

ddm/src/sm.rs
```diff
@@ -262,6 +325,8 @@ impl State for Init {
             self.ctx.config.if_name.clone_from(&info.ifname);
             self.ctx.config.if_index = info.index as u32;
             self.ctx.config.addr = addr;
+            *lock!(self.ctx.iface.if_index) = info.index as u32;
```
rcgoodfellow: Perhaps these two things are worth putting behind a `self.set_if_info()` method.

ddmadm/src/main.rs
```diff
@@ -251,6 +261,19 @@ async fn run() -> Result<()> {
     Ok(())
 }
 
+fn format_duration(d: &std::time::Duration) -> String {
```
rcgoodfellow: I think we can use the `humantime` crate for formatting. As long as we clamp the duration to the smallest timestamp we care to look at, it won't print smaller timescale values, so we don't have to read nanosecond precision timestamps every time.

rcgoodfellow (commented):
Thanks @taspelund comments follow.

Something I'd like to make sure we test with this is `ddmd` restart through `svcadm restart mg-ddm` to make sure that all the peers come back up how we expect. That was the original use case for the peer database. When we started putting interfaces in SMF that may completely obviate the peers part of the database, but I'd like to be sure with ddm-restart testing. The `trio`/`quartet` tests may actually cover this already, but I'd like to be sure.

taspelund:
@rcgoodfellow your feedback has been integrated. I will run this through a4x2 to explicitly test out the smf restart

rcgoodfellow (approved):

taspelund:
Manual testing of svcadm restart in a4x2 was successful:
```
root@oxz_switch:~# ddmadm get-prefixes
Destination            Next Hop                Path
fdb0:a840:2500:1::/64  fe80::aa40:25ff:fe00:1  g0
fd00:17:1:1::/64       fe80::aa40:25ff:fe00:1  g0
fd00:17:1:d01::/64     fe80::aa40:25ff:fe00:1  g0
fdb0:a840:2500:3::/64  fe80::aa40:25ff:fe00:3  g1
fd00:17:1:2::/64       fe80::aa40:25ff:fe00:3  g1
fd00:17:1:d02::/64     fe80::aa40:25ff:fe00:3  g1
fdb0:a840:2500:7::/64  fe80::aa40:25ff:fe00:7  g3
fd00:17:1:d04::/64     fe80::aa40:25ff:fe00:7  g3
fd00:17:1:3::/64       fe80::aa40:25ff:fe00:5  g2
fd00:17:1:d03::/64     fe80::aa40:25ff:fe00:5  g2
fdb0:a840:2500:5::/64  fe80::aa40:25ff:fe00:5  g2
root@oxz_switch:~# svcadm restart mg-ddm
root@oxz_switch:~# ddmadm get-prefixes
Destination            Next Hop                Path
fdb0:a840:2500:5::/64  fe80::aa40:25ff:fe00:5  g2
fd00:17:1:d03::/64     fe80::aa40:25ff:fe00:5  g2
fd00:17:1:3::/64       fe80::aa40:25ff:fe00:5  g2
fd00:17:1:d04::/64     fe80::aa40:25ff:fe00:7  g3
fdb0:a840:2500:7::/64  fe80::aa40:25ff:fe00:7  g3
fd00:17:1:2::/64       fe80::aa40:25ff:fe00:3  g1
fdb0:a840:2500:3::/64  fe80::aa40:25ff:fe00:3  g1
fd00:17:1:d02::/64     fe80::aa40:25ff:fe00:3  g1
fdb0:a840:2500:1::/64  fe80::aa40:25ff:fe00:1  g0
fd00:17:1:1::/64       fe80::aa40:25ff:fe00:1  g0
fd00:17:1:d01::/64     fe80::aa40:25ff:fe00:1  g0
root@oxz_switch:~# ddmadm get-prefixes
Destination            Next Hop                Path
fdb0:a840:2500:7::/64  fe80::aa40:25ff:fe00:7  g3
fd00:17:1:d04::/64     fe80::aa40:25ff:fe00:7  g3
fd00:17:1:d03::/64     fe80::aa40:25ff:fe00:5  g2
fd00:17:1:3::/64       fe80::aa40:25ff:fe00:5  g2
fdb0:a840:2500:5::/64  fe80::aa40:25ff:fe00:5  g2
fdb0:a840:2500:3::/64  fe80::aa40:25ff:fe00:3  g1
fd00:17:1:2::/64       fe80::aa40:25ff:fe00:3  g1
fd00:17:1:d02::/64     fe80::aa40:25ff:fe00:3  g1
fdb0:a840:2500:1::/64  fe80::aa40:25ff:fe00:1  g0
fd00:17:1:d01::/64     fe80::aa40:25ff:fe00:1  g0
fd00:17:1:1::/64       fe80::aa40:25ff:fe00:1  g0
root@oxz_switch:~# ddmadm get-peers
Interface  Host  Address                 Kind    Status    Duration
5          g0    fe80::aa40:25ff:fe00:1  Server  Exchange  4s 590ms
6          g1    fe80::aa40:25ff:fe00:3  Server  Exchange  4s 589ms
7          g2    fe80::aa40:25ff:fe00:5  Server  Exchange  4s 589ms
8          g3    fe80::aa40:25ff:fe00:7  Server  Exchange  4s 588ms
```

PR #696, opened 2026-04-02
[multicast] DDM multicast exchange: V4 protocol, MRIB sync

zeeshanlakhani:
Adds multicast group subscription distribution to the DDM exchange protocol with a V4 version. 

Key changes/additions:
- V4 exchange protocol with multicast support, negotiated via an `MCAST_CAPABLE` discovery flag (RFC 5492 style). The wire version byte stays at the V2 floor, so V2/V3 peers are unaffected and mixed-version topologies interoperate cleanly. 
- Two validated newtypes in the recently merged client-common sub-crate (the cycle-free leaf crate, so the API-types crates Omicron consumes avoid an omicron_common cycle):
  - `OverlayMulticast` (announced group is multicast)
  - `UnderlayMulticastIpv6` (admin-local mapping within ff04::/64)
  Both validate at construction and on deserialization, rejecting invalid origins at every ingress.
- Membership half/lower: ddmd programs DPD replication-group members in-process (`ddm::mcast`) from imported routes plus the peer table via a triggered sweep with periodic reconciliation. This matches how ddmd installs imported unicast underlay routes (`sys::add_underlay_routes`). 
- Periodic exchange resync in the peer state machine to repair drift from missed pulls
- Learned multicast state is exposed via DDM admin API for Omicron consumption
- `update_imported_mcast` is atomic within the Db (single lock for import/delete/diff, which differs from the tunnel work)
- Collapsed `send_update` dispatch, with exchange responses now checked for HTTP success (`require_success`, unicast included)
- Shared pull handler functions (`collect_underlay_tunnel`, `collect_multicast`), and a shared tfport name parser in mg-common (`port_link_from_ifname`), fixing link derivations
- Admin API binds synchronously and consistently on `::`, with an `--api-only` flag (admin API without state machines, analogous to `mgd --no-bgp-dispatcher`) for test fixtures and the Linux build.
- More serde round-trip and validation tests, a version negotiation contract test pinning mixed-version behavior, plus trio coverage of multicast advertise, router restart, and withdraw.
- CI: softnpu/sidecar-lite/dendrite pins bumped to multicast-capable builds and the trio/quartet jobs have been moved to helios-3.0 to match dendrite's build target. The test harness now fails fast if dendrite or tfport lands in SMF maintenance instead of timing out on peering.

## References

Stacked on zl/mrib (MRIB: Multicast RIB implementation, #675).


mg-lower/src/platform.rs
```diff
@@ -405,6 +430,35 @@ impl Ddm for ProductionDdm {
...
+    async fn withdraw_multicast_groups<'a>(
+        &'a self,
+        body: &'a Vec<MulticastOrigin>,
+    ) -> Result<
+        ddm_admin_client::ResponseValue<()>,
+        ddm_admin_client::Error<DdmError>,
+    > {
+        self.client.withdraw_multicast_groups(body).await
+    }
 }
```
taspelund: similar question here about the ref to a vec (although no clippy allowances here? not sure why it would be needed in one trait but not the other)

mg-lower/src/platform.rs
```diff
@@ -766,6 +822,45 @@ pub(crate) mod test {
...
+        async fn withdraw_multicast_groups<'a>(
+            &'a self,
+            body: &'a Vec<MulticastOrigin>,
+        ) -> Result<
+            ddm_admin_client::ResponseValue<()>,
+            ddm_admin_client::Error<DdmError>,
+        > {
+            self.multicast_originated
+                .lock()
+                .unwrap()
```
taspelund: Can we use the `lock!()` macro instead?

mg-lower/src/platform.rs
```diff
@@ -216,6 +216,31 @@ pub trait Ddm {
...
+    >;
+
+    #[allow(clippy::ptr_arg)]
+    async fn withdraw_multicast_groups<'a>(
+        &'a self,
+        body: &'a Vec<MulticastOrigin>,
+    ) -> Result<
+        ddm_admin_client::ResponseValue<()>,
+        ddm_admin_client::Error<DdmError>,
+    >;
```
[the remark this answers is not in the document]
zeeshanlakhani: If you look at the `advertise|withdraw_tunnel|etc...` functions, you'll see the same pattern. `#[allow(clippy::ptr_arg)]` is forced by the progenitor-generated client, as the generated signature uses `&Vec<MulticastOrigin>`, not `&[MulticastOrigin]`. This is done throughout the module. The ddm-api returns `TypedBody<HashSet<<TunnelOrigin>>>`, and progenitor, in-turn, converts that into `&Vec<MulticastOrigin>`. 

We could change the `Ddm` trait to `&[MulticastOrigin]`, but then you'd need `body.to_vec` to satisfy the client, so an extra allocation for no reason.

mg-lower/src/platform.rs
```diff
@@ -766,6 +822,45 @@ pub(crate) mod test {
...
         }
+
+        async fn get_originated_multicast_groups(
+            &self,
+        ) -> Result<
+            ddm_admin_client::ResponseValue<Vec<MulticastOrigin>>,
+            ddm_admin_client::Error<DdmError>,
+        > {
+            Ok(ddm_response_ok!(
+                self.multicast_originated.lock().unwrap().clone()
```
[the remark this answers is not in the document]
zeeshanlakhani: https://github.com/oxidecomputer/maghemite/pull/696#discussion_r3042587818

ddm/src/sys.rs
```diff
@@ -359,6 +359,99 @@ pub fn remove_tunnel_routes(
...
+
+/// Remove OPTE M2P table entries for withdrawn multicast routes.
+#[cfg(target_os = "illumos")]
+pub fn remove_multicast_routes(
+    log: &Logger,
+    ifname: &str,
+    routes: &HashSet<MulticastRoute>,
+) -> Result<(), String> {
+    use oxide_vpc::api::ClearMcast2PhysReq;
+    use oxide_vpc::api::MulticastUnderlay;
```
taspelund: same consolidation nit here

ddm/src/admin.rs
```diff
@@ -333,6 +333,71 @@ impl DdmAdminApi for DdmAdminApiImpl {
...
+        Ok(HttpResponseOk(originated))
+    }
+
+    async fn get_multicast_groups(
+        ctx: RequestContext<Self::Context>,
+    ) -> Result<HttpResponseOk<HashSet<MulticastRoute>>, HttpError> {
+        let ctx = lock!(ctx.context());
+        let imported = ctx.db.imported_mcast();
+        Ok(HttpResponseOk(imported))
+    }
```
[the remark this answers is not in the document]
zeeshanlakhani: It's not as straightforward due to needing access to `ctx.db` which lives in `ddm::Db`, and ddm depends on `ddm-api`. But, I can abstract this and then default it in `mg_common`.

ddm/src/exchange.rs
```diff
@@ -794,26 +923,87 @@ async fn pull_handler(
...
+            path: vec![hop],
+        });
+    }
+
+    Ok(multicast)
+}
+
+fn opt<T>(s: HashSet<T>) -> Option<HashSet<T>> {
+    if s.is_empty() { None } else { Some(s) }
+}
```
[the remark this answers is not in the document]
zeeshanlakhani: I moved things around to make this more generic, but I think it's cleaner.

mg-common/src/net.rs
```diff
@@ -93,3 +185,131 @@ pub enum IpPrefix {
...
+}
+
+impl PartialEq for MulticastOrigin {
+    fn eq(&self, other: &Self) -> bool {
+        self.overlay_group == other.overlay_group
+            && self.underlay_group == other.underlay_group
+            && self.vni == other.vni
+            && self.source == other.source
+    }
+}
```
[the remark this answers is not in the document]
zeeshanlakhani: Used on Hashset comparisons, but now commented.

zeeshanlakhani:
Now in line with the parent.

client-common/src/vni.rs
```diff
@@ -0,0 +1,139 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Geneve Virtual Network Identifier (VNI).
+//!
+//! Lives in the cycle-free leaf crate so the API-types crates consumed by
+//! Omicron can share a single definition without depending on
+//! `omicron_common`, which would form a dependency cycle.
```
taspelund, line 9: I would REALLY love to see this in oxnet so it can be reused by dendrite, omicron, etc.

client-common/src/multicast.rs
```diff
@@ -0,0 +1,236 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Validated underlay multicast address shared across the routing suite.
+//!
+//! Lives in the cycle-free leaf crate so the API-types crates consumed by
+//! Omicron can share a single definition without depending on
+//! `omicron_common`, which would form a dependency cycle.
```
taspelund, line 9: Would these types be good candidates for oxnet? The crate is meant to hold networking types for Oxide use, not strictly general-purpose networking types.

If this type or an analogue to it exists in omicron, opte, or dendrite, then I think we should consider it

zeeshanlakhani: Oh for sure to this and [above](https://github.com/oxidecomputer/maghemite/pull/696#discussion_r3877470511). There are TODOs on this front already (about moving all of this to oxnet). However, oxnet is used by so many repos, so we should do it in one shot for everything (after these are reviewed/merged in), and, at least, it's additive.

zeeshanlakhani: I did add TODOs.

PR #654, opened 2026-03-04
Levon/make bgp dispatch listen configurable

internet-diglett:

PR #698, opened 2026-04-03
re-implement #654 onto latest main

internet-diglett:
Re-implements #654 because it was kinda hairy solving the merge conflicts

### DIFF
--- a/mgd/src/bgp_admin.rs
+++ b/mgd/src/bgp_admin.rs
@@ -2052,6 +2052,10 @@ pub(crate) mod helpers {
             "params" => format!("{rq:#?}")
         );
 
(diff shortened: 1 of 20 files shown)


mgd/src/bgp_admin.rs
```diff
@@ -2048,6 +2052,10 @@ pub(crate) mod helpers {
             "params" => format!("{rq:#?}")
         );
 
+        // Validate that at least one AF is enabled
+        rq.validate_address_families()
+            .map_err(Error::InvalidRequest)?;
+
```
taspelund: LGTM, thanks!

taspelund (approved):
Thanks for the last few changes levon, ship it!

PR #717, opened 2026-04-22
BFD and falcon-lab cleanup [Spring Cleaning 1/N]

taspelund:
First round of splits from #648 

## BFD daemon cleanup

- Switch BFD threads to `ManagedThread` + named `Builder::spawn`; propagate spawn `Result`s up to admin callers.
- Group peer config into `AddPeerRequest`; add `AddPeerError::PeerExists` so mgd returns 409 instead of 500 on duplicates.
- Convert `egress()` from blocking `rx.recv()` to `recv_timeout(1s)` with a kill-switch poll, avoiding the silent-channel deadlock.
- Extend bfd unit tests with IPv6 peers.

## New falcon-lab test: `trio-bfd-static-routing`

- Dual-stack BFD + static routing over the existing trio topology (FRR + cEOS peers), with shared bootstrap extracted into `boot_trio()` / `BootedTrio`.
- Five-phase assertion matrix exercises BFD-gated next-hop shutdown and the "all shutdown → all reinstated" fallback. Each phase validates mgd BFD state, peer-side BFD state on live peers, and the dpd target sets for both v4 and v6 prefixes.
- Failure injection via `pkill -STOP bfdd` on FRR and `docker pause ceos` on EOS — reversible, daemon-state-preserving.
- Plumbing needed for numbered peers to work: per-link `link_ipv{4,6}_create` + `link_ipv6_enabled_set`, run ddmd (mg-lower depends on it), bump `bestpath_fanout` to 2, pre-seed each tfport with an `addrconf` link-local before static v6, and tolerate "already"-style `ipadm` errors for `--no-cleanup` re-runs.

## Trio-unnumbered assertion tightening

- Per-neighbor FSM check.
- Assert specific prefixes and path counts in imported RIB, selected RIB, dpd targets, and each peer's imported BGP routes, replacing count-only assertions.
- Bump `bestpath_fanout` to 2 so ECMP is actually exercised.

### DIFF
--- a/falcon-lab/src/test.rs
+++ b/falcon-lab/src/test.rs
@@ -759,15 +759,15 @@ async fn expect_route(
 ) -> Result<()> {
     // Push cr1 before cr2 so the list is already in the sorted order that
     // dpd_v*_targets returns.
-    let mut want_v4 = Vec::new();
-    let mut want_v6 = Vec::new();
+    let mut want_v4: Vec<IpAddr> = Vec::new();
+    let mut want_v6: Vec<IpAddr> = Vec::new();
     if cr1_in {
-        want_v4.push(CR1_V4);
-        want_v6.push(CR1_V6);
+        want_v4.push(IpAddr::V4(CR1_V4));
+        want_v6.push(IpAddr::V6(CR1_V6));
     }
     if cr2_in {
-        want_v4.push(CR2_V4);
-        want_v6.push(CR2_V6);
+        want_v4.push(IpAddr::V4(CR2_V4));
+        want_v6.push(IpAddr::V6(CR2_V6));
     }
 
     let desc_v4 = format!("{phase} v4");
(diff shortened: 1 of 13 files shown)


falcon-lab/src/test.rs
```diff
@@ -320,3 +420,457 @@ async fn eos_setup(r: EosNode, d: Arc<Runner>) -> Result<()> {
...
     Ok(())
 }
+
+pub async fn cleanup_bfd_static_test() -> Result<()> {
+    // dropping this without persistent set will destroy the topo
+    let _topo = trio(TRIO_BFD_STATIC_TOPO_NAME)?;
+    Ok(())
+}
+
+pub async fn run_trio_bfd_static_test(
```
rcgoodfellow, line 430: This is awesome. Really stoked to see this e2e test come together.

rcgoodfellow (approved):

PR #720, opened 2026-04-25
Light Cleanup [Spring Cleaning 2/N]

taspelund:
Second round of splits from #648 

Minor changes:
- set TCP_NODELAY on BGP connections
- LoopbackIpManager bug fixes
- EPIPE handling in mgadm/ddmadm
- Fix a small conversion bug in BGP RHAI policy
- Cleanup a couple instances of shared state in BGP tests (so things will work under `cargo test` in addition to `cargo nextest`)

See each commit for specifics

### DIFF
--- a/mg-common/src/lib.rs
+++ b/mg-common/src/lib.rs
@@ -39,6 +39,115 @@ pub fn format_duration_human(d: Duration) -> String {
     }
 }
 
+/// Like `println!`, but silently exits on broken pipe (EPIPE) instead of
+/// panicking. Other I/O errors still panic.
+#[macro_export]
+macro_rules! println_nopipe {
+    () => {
+        {
+            use std::io::Write;
+            let r = writeln!(std::io::stdout());
+            match r {
+                Ok(_) => {},
+                Err(e) if e.kind() == std::io::ErrorKind::BrokenPipe => {
+                    std::process::exit(0);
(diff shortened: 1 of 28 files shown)


taspelund:
Just pushed an update to this branch that rebases it atop the major code reorganization in #724

mg-common/src/lib.rs
```diff
@@ -39,6 +39,115 @@ pub fn format_duration_human(d: Duration) -> String {
     }
 }
 
+/// Like `println!`, but silently exits on broken pipe (EPIPE) instead of
+/// panicking. Other I/O errors still panic.
+#[macro_export]
```
rcgoodfellow, line 44: seems like we might want a crate for this that can be shared by our various CLIs. Very similar code [here](https://github.com/oxidecomputer/oxide.rs/blob/main/cli/src/print.rs)

rcgoodfellow (approved):
LGTM. Just one comment on the nopipe CLI stuff

mg-common/src/lib.rs
taspelund: Agreed, that's where I stole the idea from! 😄

Would you prefer to solve this prior to merging this PR, or is it cool if I follow-up after I've had some time to think about what a clean impl would look like?

PR #724, opened 2026-04-28
Wrap up RFD 619 migration

taspelund:
Many types were not fully migrated to the RFD 619 conventions in #611 and #665. This PR seeks to remedy that.

As a whole, this PR is simply a re-organization of existing code rather than a functional change.
It does so by splitting some crates apart (introducing `$CRATE-types` and `$CRATE-types-versions` for it) and attempting to bring some order to this chaos. In some cases, types were moved into a more fitting home (e.g. BFD types moving out of rdb/mgd crates and into bfd-types).

The exception to no functional changes is that the operation_id in the OpenAPI spec does change for a handful of methods.
These are all broken out into a new dropshot API version for mg-api: `ENDPOINT_RENAME`.
This results in clients seeing a different method name when importing the latest API version via mg-admin-client.
Aside from the method/operation_id name change, the API remains byte-compatible.

I had claude help with the grunt work on this one, and the commit messages describe the implementation plan it followed (including some course corrections).


ddm-api/Cargo.toml
```diff
@@ -4,10 +4,10 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
-ddm-types-versions.workspace = true
+ddm-api-types-versions.workspace = true
 dropshot.workspace = true
 dropshot-api-manager-types.workspace = true
-mg-common.workspace = true
+ddm-api-types.workspace = true
```
sunshowers: Do we still need to depend on ddm-api-types here?

mg-api/Cargo.toml
```diff
@@ -5,9 +5,10 @@ edition = "2024"
 
 [dependencies]
 bgp.workspace = true
+mg-api-types.workspace = true
 dropshot.workspace = true
 dropshot-api-manager-types.workspace = true
-mg-types-versions.workspace = true
+mg-api-types-versions.workspace = true
```
sunshowers: Same q -- do we still need to depend on mg-api-types here?

mg-api/src/lib.rs
sunshowers, line 67: `request` should be `latest::bfd::BfdPeerConfig`. (There's a few other examples in this file of this pattern.)

mg-api-types/versions/src/impls/policy.rs
```diff
@@ -0,0 +1,53 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Cross-version helpers for `crate::v{1,4}::bgp::policy::ImportExportPolicy*`.
+
+use crate::v1::bgp::policy::ImportExportPolicy;
+use crate::v1::rdb::prefix::Prefix;
+use crate::v4::bgp::policy::{ImportExportPolicy4, ImportExportPolicy6};
```
sunshowers: `impls` should only refer to latest identifiers (so that as a newer version is added, the impls are carried forward). There's a few examples of this.

mg-api-types/versions/src/impls/rib.rs
```diff
@@ -0,0 +1,35 @@
...
+            value
+                .into_iter()
+                .map(|(k, v)| {
+                    let paths_v1: BTreeSet<v1::rdb::path::Path> =
+                        v.into_iter().map(v1::rdb::path::Path::from).collect();
+                    (k.to_string(), paths_v1)
+                })
+                .collect(),
+        )
+    }
```
sunshowers: This should definitely live in the version module that `latest::rib::Rib` re-exports from.

mg-api-types/versions/src/latest.rs
```diff
@@ -0,0 +1,161 @@
...
+
+                    if !filtered_paths.is_empty() {
+                        filtered.insert(prefix, filtered_paths);
+                    }
+                }
+
+                filtered
+            }
+        }
+    }
```
sunshowers: Hmm, not the right place for this.

mg-api-types/versions/src/latest.rs
```diff
@@ -0,0 +1,161 @@
...
+    pub use crate::v5::bgp::{
+        ExportedSelector, FsmHistoryRequest, FsmHistoryResponse,
+        MessageHistoryRequest, MessageHistoryResponse, NeighborSelector,
+        PeerInfo, PeerTimers, UnnumberedNeighborResetRequest,
+        UnnumberedNeighborSelector,
+    };
+    pub use crate::v8::bgp::{
+        ApplyRequest, BgpPeerConfig, BgpPeerParameters, Neighbor,
+        UnnumberedBgpPeerConfig, UnnumberedNeighbor,
+    };
```
sunshowers: These and the rest of the file should list each type out on a separate line, grouped by version, separated by whitespace. (See [section 4.9](https://rfd.shared.oxide.computer/rfd/0619#determinations-reexports)).

The reason we prefer this pattern is that when a type changes here we'd like to preserve blame.

ddm-api-types/versions/src/latest.rs
```diff
@@ -21,3 +21,7 @@ pub mod exchange {
     pub use crate::v1::exchange::PathVector;
     pub use crate::v1::exchange::PathVectorV2;
 }
+
+pub mod net {
+    pub use crate::v1::net::{IpPrefix, Ipv4Prefix, Ipv6Prefix, TunnelOrigin};
+}
```
sunshowers, line 30: These should be listed out on separate lines.

mg-api-types/versions/src/lib.rs
```diff
@@ -29,8 +29,10 @@
 //! [detailed guide]: https://github.com/oxidecomputer/dropshot-api-manager/blob/main/guides/new-version.md
 //! [RFD 619]: https://rfd.shared.oxide.computer/rfd/619
 
+pub mod error;
 mod impls;
 pub mod latest;
+pub mod parse;
```
sunshowers: `error` and `parse` should either live inside `impls`, or be subsumed into their corresponding modules (for example, maybe the error type can live inside `impls::bgp`)?

mg-api-types/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+// facade). Consumers reach `Prefix4`, `AddressFamily`, etc. via
+// `mg_api_types::*`; the progenitor `replace = {}` block in
+// `mg-admin-client` depends on these top-level paths.
+pub use mg_api_types_versions::latest::rdb::neighbor::*;
+pub use mg_api_types_versions::latest::rdb::path::*;
+pub use mg_api_types_versions::latest::rdb::prefix::*;
+pub use mg_api_types_versions::latest::rdb::router::*;
+pub use mg_api_types_versions::latest::rdb::{
+    AddressFamily, ProtocolFilter, Rib, Rib4, Rib6, filter_rib_by_protocol,
+};
```
sunshowers: I would put these in an `rdb` module to match the rest of the crate. (Also can this just be a wildcard re-export, similar to the rest of the files?)

mgd/src/bgp_admin.rs
```diff
@@ -12,38 +12,40 @@ use crate::validation::{
...
+    ShaperSource, UnnumberedNeighbor, UnnumberedNeighborResetRequest,
+    UnnumberedNeighborSelector,
 };
-use mg_types::ndp::{
+use mg_api_types::ndp::{
     NdpInterface, NdpInterfaceSelector, NdpManagerState, NdpPeer,
     NdpPendingInterface, NdpThreadState,
 };
-use mg_types_versions::{v1, v2, v4, v5};
+use mg_api_types_versions::v1::bgp::policy::ImportExportPolicy;
```
sunshowers: Can the use of `ImportExportPolicy` be switched to being referred to via the `v1::` path? That will prevent a future latest `ImportExportPolicy` export from colliding with this one.

sunshowers (commented):
Thanks for doing this! This is quite large so I had Claude Code review it using this prompt:

```
Fetch https://raw.githubusercontent.com/oxidecomputer/dropshot-api-manager/refs/heads/main/guides/migrating-to-versions-crate.md
  using curl (do not summarize) and review this branch (compare to merge
  base with main) for compliance with RFD 619. Remember to also fetch the
  RFD itself as directed in the document.
```

This is a selection of what it pointed out, plus a couple things I spotted myself. You may want to also run this locally yourself to catch all the examples of drift from the RFD.

taspelund:
@sunshowers thank you for reviewing!  I believe all your feedback has been addressed.  I took your advice and used your prompt locally, which ended up finding a handful of other items that are now addressed too

mg-api-types/versions/src/rib_exported_string_key/mod.rs
```diff
@@ -0,0 +1,3 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
```
sunshowers, line 3: I'd just drop this module.

mg-api-types/versions/src/operation_id_cleanup/mod.rs
```diff
@@ -1,7 +1,3 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
```
sunshowers, line 3: I'd also drop this module.

mg-api-types/versions/src/impls/bgp/mod.rs
```diff
@@ -0,0 +1,944 @@
...
+            asn,
+            name,
+            host,
+            group,
+            parameters,
+        }
+    }
+}
+
+impl v1::bgp::config::Neighbor {
```
sunshowers: This should live in the `v1` module since it isn't related to the latest types. (Alternatively, can we switch to using the latest method defined below and then use the `From` impl to convert to v1?

mg-api-types/versions/src/impls/bgp/mod.rs
```diff
@@ -0,0 +1,944 @@
...
+        deterministic_collision_resolution: false,
+        src_addr,
+        src_port,
+    }
+}
+
+/// Build a v1 `BgpPeerParameters` from runtime `BgpNeighborParameters`.
+fn v1_params_from_rdb(
+    params: BgpNeighborParameters,
+) -> v1::bgp::config::BgpPeerParameters {
```
sunshowers: Same here -- v1-related implementation code should live in the v1 module.

mg-api-types/versions/src/impls/bgp/mod.rs
```diff
@@ -0,0 +1,944 @@
...
+        })
+    } else {
+        None
+    }
+}
+
+// ----- v2 (ipv6_basic) <-> v1 (initial) downgrades for PeerInfo / FsmStateKind -----
+
+impl From<crate::v2::bgp::session::FsmStateKind>
+    for v1::bgp::config::FsmStateKind
```
sunshowers: This should live in v2.

mg-api-types/versions/src/impls/bgp/mod.rs
```diff
@@ -0,0 +1,944 @@
...
+            asn,
+            duration_millis,
+            timers,
+        }
+    }
+}
+
+// ----- v1 (initial, frozen) -> latest upgrades -----
+
+impl From<v1::bgp::config::BgpPeerConfig> for latest::bgp::BgpPeerConfig {
```
sunshowers: `From` impls should live next to wherever the `latest` types are defined -- the conversion is always going to be fixed. `impls` is just for things that have to move forward as a new version of a type is introduced. (Sorry, I know it's a lot, but it's all for very good reasons.)

mg-api-types/versions/src/impls/bgp/mod.rs
```diff
@@ -0,0 +1,944 @@
...
+                idle_hold_jitter: None,
+                deterministic_collision_resolution: false,
+                src_addr: None,
+                src_port: None,
+            },
+        }
+    }
+}
+
+impl From<v1::bgp::config::ApplyRequest> for latest::bgp::ApplyRequest {
```
sunshowers: Same here -- should live where `latest::bgp::ApplyRequest` is defined.

mg-api-types/versions/src/impls/bgp/mod.rs
```diff
@@ -0,0 +1,944 @@
...
+                .collect(),
+            unnumbered_peers: HashMap::default(),
+        }
+    }
+}
+
+// ----- v4 (mp_bgp, frozen) <-> latest (v8) BgpPeerParameters -----
+
+impl From<latest::bgp::BgpPeerParameters>
+    for v4::bgp::config::BgpPeerParameters
```
sunshowers: Same here, and in the rest of the file.

mg-admin-client/src/lib.rs
```diff
@@ -17,18 +17,59 @@ progenitor::generate_api!(
...
+        JitterRange = mg_api_types::bgp::JitterRange,
+        MessageDirection = mg_api_types::bgp::MessageDirection,
+        NeighborResetOp = mg_api_types::bgp::NeighborResetOp,
+        NeighborResetRequest = mg_api_types::bgp::NeighborResetRequest,
+        Origin4 = mg_api_types::bgp::Origin4,
+        Origin6 = mg_api_types::bgp::Origin6,
+        PeerInfo = mg_api_types::bgp::PeerInfo,
+        Router = mg_api_types::bgp::Router,
+        ShaperSource = mg_api_types::bgp::ShaperSource,
+        UnnumberedNeighborResetRequest = mg_api_types::bgp::UnnumberedNeighborResetRequest,
```
sunshowers: can these and the rest of the re-exports use `mg_api_types_versions::latest` instead? We prefer that because the versions crates are where published types live (the types crates can have unpublished types).

sunshowers (commented):
Just a few more comments, thanks.

taspelund:
Alrighty, the latest rounds of comments have been addressed. Claude believes we are fully compliant with RFD 619 😁

mg-api-types/versions/src/ipv6_basic/bgp/session.rs
```diff
@@ -0,0 +1,159 @@
...
+
+use std::collections::VecDeque;
+use std::net::SocketAddr;
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+use crate::v1;
+use crate::v4::bgp::messages::Message;
```
sunshowers: This looks like the v2 module which refers to `Message` in the v4 module. Does `Message` need to be moved to v2?

mg-api-types/versions/src/ipv6_basic/bgp/history.rs
```diff
@@ -65,14 +71,32 @@ pub struct FsmHistoryResponse {
...
+
+    /// Set of prefixes to originate.
+    pub prefixes: Vec<Prefix6>,
+}
+
 #[derive(Default, Debug, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct MessageHistory {
     pub received: VecDeque<MessageHistoryEntry>,
     pub sent: VecDeque<MessageHistoryEntry>,
 }
```
sunshowers, line 96: Hmm, it looks like there are now two copies of `MessageHistory` in the v2/ipv6_basic module: here and in `session.rs`.

mg-api-types/versions/src/ipv6_basic/bgp/history.rs
```diff
@@ -115,18 +139,20 @@ pub enum Message {
     RouteRefresh(RouteRefreshMessage),
 }
 
-impl From<bgp::messages::Message> for Message {
-    fn from(msg: bgp::messages::Message) -> Self {
+impl From<v4::bgp::messages::Message> for Message {
+    fn from(msg: v4::bgp::messages::Message) -> Self {
```
sunshowers: v2 <-> v4 conversions should live in v4.

mg-api-types/versions/src/impls/rib.rs
```diff
@@ -0,0 +1,25 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Cross-version downgrade from the latest published RIB shape to v1.
+
+use std::collections::BTreeSet;
+
+use crate::{latest, v1};
+
+impl From<latest::rib::Rib> for v1::rib::Rib {
```
sunshowers: This should live next to where `latest::rib::Rib` is defined.

mg-api-types/versions/src/impls/session.rs
```diff
@@ -0,0 +1,185 @@
...
+                .into_iter()
+                .map(v1::bgp::session::MessageHistoryEntry::from)
+                .collect(),
+            sent: sent
+                .into_iter()
+                .map(v1::bgp::session::MessageHistoryEntry::from)
+                .collect(),
+        }
+    }
+}
```
sunshowers: These should live in the v2 module.

sunshowers (commented):
Almost! This is all I could find.

taspelund:
Alrighty, latest is up for review.  There's only one thing that was flagged by claude in the latest run, and it's not entirely clear what the "right" move is.  Do you have any opinions?

> While auditing this branch against RFD 619, I noticed latest.rs mixes two re-export patterns per area: most types are hoisted to the area root (crate::v1::bgp::config::Router → latest::bgp::Router) while specialized families stay subnamespaced (crate::v4::bgp::messages::Message → latest::bgp::messages::Message). RFD 619's prose doesn't mandate strict mirroring, but every worked example assumes it. Moving to strict per-domain mirror (no hoists, facade path always matches disk) would re-namespace 69 types in latest.rs and update ~33 business-logic files plus the progenitor replace block; no OpenAPI changes. It's all-or-nothing, propagates to every future API version, and trades shorter import paths for stricter alignment. My lean is to stick with hybrid and document the convention, but happy to flip if you'd rather have strict mirror — wanted an explicit call before locking it into the docs.

sunshowers:
> latest.rs mixes two re-export patterns per area: most types are hoisted to the area root (crate::v1::bgp::config::Router → latest::bgp::Router) while specialized families stay subnamespaced (crate::v4::bgp::messages::Message → latest::bgp::messages::Message)

Ideally these would be mirrored and consistent. i.e. for every identifier, replacing `vN` with `latest` without changing anything else about the path leads to the same result. So I'd pick anything that works for you and is consistent across the fixed and floating identifiers.

taspelund:
>Ideally these would be mirrored and consistent. i.e. for every identifier, replacing vN with latest without changing anything else about the path leads to the same result

Ok, this is now in place

sunshowers (approved):
Thank you! I know this has been a lot

taspelund:
Thank you for all the rounds of review!

mg-api-types/versions/src/initial/bgp/config.rs
```diff
@@ -0,0 +1,231 @@
...
+/// V1 API NeighborSelector (numbered peers only).
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+pub struct NeighborSelector {
+    pub asn: u32,
+    pub addr: IpAddr,
+}
+
+/// V1 API neighbor reset operations (backwards compatibility)
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+#[schemars(rename = "NeighborResetOp")]
```
ahl, line 30: this is a no-op and I see this pattern a LOT in this PR. Just a heads up @taspelund that this is not needed. I'm going to make a PR to remove all this.

PR #729, opened 2026-05-07
[ddmd] add --api-only flag for test fixtures and Linux build

zeeshanlakhani:
Omicron's https://github.com/oxidecomputer/omicron/pull/10381 introduces a stubbed `ddmd` admin endpoint because spawning a real `ddmd` in a generic test toolchain is not viable: the routing state machine (discovery, exchange, route synchronization) depends on illumos networking facilities the toolchain does not provide. Consumers of the stub, e.g., Nexus RPW (multicast members), sled-agent's DDM reconciler, and anything that resolves the DDM internal-DNS service name, cannot exercise the real admin surface from Omicron's test harness.

This work adds an opt-in `--api-only` flag to `ddmd` that runs only the admin API server and skips the state machine entirely, allowing the fixture to spawn the real binary. This is analogous to `mgd --no-bgp-dispatcher`, which Omicron's `MgdInstance` already uses for the same purpose.

To make the fixture path usable on Linux, `ddmd` itself must build on Linux. The previous code pulled the illumos-only crates `libnet`, `dpd-client`, `opte-ioctl`, and `oxide-vpc` unconditionally through `ddm`, which failed to link on Linux (`-lzfs`, `-ldlpi`). This change introduces a `backend` feature in both `ddm` and `ddmd` (default-on, mirroring `mgd`'s `mg-lower` pattern) that marks those four crates optional. The buildomat `linux.sh` job now builds `ddmd` and `ddmadm`, with `ddmd` invoked as `cargo build --bin ddmd --no-default-features`.

The illumos-only halves of `ddm` are isolated by the feature gate:

- The routing state machine implementation moves from `sm.rs` into `sm/state.rs`.
- The exchange runtime (HTTP push/pull and route programming) moves from `exchange.rs` into `exchange/runtime.rs`.
- The discovery runtime (UDPv6 solicitation/advertisement loops) moves from `discovery.rs` into `discovery/runtime.rs`.

Each parent `mod.rs` keeps the platform-agnostic types and re-exports the runtime surface so existing call sites resolve unchanged on illumos. The runtime submodules are gated as a unit by `#[cfg(all(feature = "backend", target_os = "illumos"))]`. We also remove the single-function `ddm/src/util.rs`, inlining the function into `discovery/runtime.rs`, where its sole caller lives.

The SIGTERM cleanup handler is installed regardless of the flag, so Ctrl-C still exits cleanly in `--api-only` mode. The imported route sets are empty in that mode, so the cleanup itself is a noop. `--api-only` and `--addr` are mutually exclusive at the clap level (`conflicts_with`), so passing them together is rejected at parse time.


ddmd/src/main.rs
jgallagher: I _think_ you could put everything from here up until the `exit()` inside a cfg-gated block like

```rust
#[cfg(...)]
{
    // all the real shutdown work
}
```

inside this function and not need to define a fully separate `termination_handler()` fn.

ddmd/src/main.rs
```diff
@@ -237,6 +201,128 @@ async fn run() {
...
+    dpd: &Option<DpdConfig>,
+    hostname: &str,
+    rt: &Arc<tokio::runtime::Handle>,
+    log: &Logger,
+) -> (
+    Vec<StateMachine>,
+    Vec<std::sync::mpsc::Sender<ddm::sm::Event>>,
+) {
+    if arg.no_state_machine {
+        if !arg.addresses.is_empty() {
```
jgallagher: Should `clap` reject invocations that have `--no-state-machine` with one or more addresses?

ddm/Cargo.toml
```diff
@@ -35,3 +31,15 @@ oxnet.workspace = true
...
+# illumos-only deps used by the routing state machine and platform sys layer.
+# Gated by the `illumos` feature so non-illumos builds (e.g. Linux test
+# fixtures running ddmd with `--no-state-machine`) link cleanly.
+libnet = { workspace = true, optional = true }
+dpd-client = { workspace = true, optional = true }
+opte-ioctl = { workspace = true, optional = true }
+oxide-vpc = { workspace = true, optional = true }
+
+[features]
+default = ["illumos"]
```
jgallagher: I think I'd name this feature something else to avoid confusion between the feature name and the target OS. This is especially true if we want to be able to build a stub `ddmd` _on illumos_, since in that case you'd want "not the `illumos` feature but yes the `illumos` target OS". Not sure what to suggest for a name, though; could either be something implying "not a stub" (not my favorite but maybe the most sensible) or something describing the actual functionality that's only present if it's enabled (`apply-routes` or something like that).

jgallagher (commented):
Just a few notes on the mechanics of the split; I'll defer to folks who know maghemite better for the code organization.

ddmd/src/main.rs
zeeshanlakhani: Yeah, I had this version earlier, and went back and forth on it.

jgallagher: Yeah I don't feel strongly about it either way. I generally prefer "shorter" here, but can see an argument for it being clearer to be separate functions entirely.

taspelund:
Code seems ok to me overall, but I haven't spent enough time in the ddm side of this repo to have strong opinions on the organization.

The feature and CLI flag naming could be a little more intuitive, as I don't think it's immediately obvious what effect "no state machine" would have to someone who isn't familiar with ddm or the feature itself.

Maybe the CLI flag could be something like --api-only and the cargo feature could be "backend"?

zeeshanlakhani:
@taspelund updated.

ddm/src/admin.rs
```diff
@@ -68,11 +68,11 @@ pub fn handler(
         ..Default::default()
     };
 
-    let ds_log = ConfigLogging::StderrTerminal {
-        level: ConfigLoggingLevel::Error,
-    }
-    .to_logger("admin")
-    .map_err(|e| e.to_string())?;
+    let ds_log = log.new(o!(
```
jgallagher: This is a change to the logging _level_ too, right? Previously we constructed a `StderrTerminal` at the `Error` level, but now we're inheriting whatever level `log` has - presumably `Info`? Is that okay?

ddm/src/exchange/runtime.rs
```diff
@@ -588,11 +297,11 @@ pub fn handler(
         ..Default::default()
     };
 
-    let ds_log = ConfigLogging::StderrTerminal {
-        level: ConfigLoggingLevel::Error,
-    }
-    .to_logger("exchange")
-    .map_err(|e| e.to_string())?;
+    let ds_log = log.new(o!(
```
jgallagher: Same kind of logging-level change here.

jgallagher (approved):
I'm not sure who _should_ approve this - @taspelund and I both deferred. I'll slap my approval on with a couple comments about code that changed during the move - other than that this looks like just rearranging organization to support the new flags. But big 👍 if you want to find and wait for whoever really owns this to take a look.

ddm/src/admin.rs
zeeshanlakhani: Yeah, good discussion point. Structured-logging aside (which is good here), the move to `Info` matches what `mgd` does vs what `ddm` had ~ `mgd/src/admin.rs` has been constructing its dropshot logger as `log.new(o!("component" => COMPONENT_MGD, "module" => MOD_ADMIN, "unit" => UNIT_API_SERVER))`, doing what the the parent level runs at (typically `Info`). The Error cap was a ddm-only.  

So the question is do we want both daemons quiet normally vs on error? I think, I'll keep what ddm did previously and file an issue to look at logging consistency.

ddm/src/exchange/runtime.rs
zeeshanlakhani: Same response as above. On-par commit incoming.

ddm/src/admin.rs
taspelund: I'm going to tag in @rcgoodfellow in to help give some historical context -- I just don't have the history with ddm to know why we'd filter the logs down to just >= error.

I would think info should generally not be so cluttered that we need to avoid them in the logger.  If the reason for the filter was truly just noise, then we should consider converting noisy logging calls to debug instead of leaving them at info... although I'm ok if that means filing a follow-up issue to address it later, rather than holding up this PR

zeeshanlakhani: #740 opened to do the logging level change separately.

zeeshanlakhani:
I discussed this with @taspelund, and he was good with it. We'll tackle the logging level separately as per the issue.

PR #730, opened 2026-05-07
[mg-admin-api] fix v2/v3 MessageHistory and related types

sunshowers:
This type was changed in v4, but the change was accidentally also applied to v2 and v3. The nontrivial change was not detected by drift due to https://github.com/oxidecomputer/drift/pull/22, which is now fixed in drift 0.1.4.

Fix this up by introducing types and conversions for v2. This is unfortunately complicated by the fact that a lot of the BGP types are not in mg-admin-types-versions, partly because untangling this is quite difficult. I've opted to just do a relatively minimal fix for now where the v2 types are defined in mg-admin-types-versions, but the newest types continue to live in bgp. I'll try untangling this further in the future.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1617,9 +1617,9 @@ dependencies = [
 
 [[package]]
 name = "drift"
-version = "0.1.3"
+version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "44cc7f4e6337c9c0d0402e2dccaeaea63fc38467a3e5ab3f465372dafce2cd89"
+checksum = "d46ae7066bba53a265e8412309dc64739d32c597d688f1938a44c01bd30629e7"
 dependencies = [
  "anyhow",
  "indexmap 2.14.0",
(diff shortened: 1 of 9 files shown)


sunshowers:
With drift 0.1.4 but without this fix, we would see this error: https://gist.github.com/sunshowers/cc4b0ce3702acef9cad27ac7d0e36136

taspelund:
@sunshowers I recently opened https://github.com/oxidecomputer/maghemite/pull/724 that I think tries to do the same thing as this PR, but applies it to the whole repo. If you have time to give it a look, I'd love to get your feedback on it

sunshowers:
> @sunshowers I recently opened https://github.com/oxidecomputer/maghemite/pull/724 that I think tries to do the same thing as this PR, but applies it to the whole repo. If you have time to give it a look, I'd love to get your feedback on it

Oh neat. I actually started on this myself, can definitely have a look at that later.

Would it be okay if we got this in? This is a correctness fix. I'm happy to help merge main into your branch afterwards.

taspelund:
Yeah, totally. This looks great, thank you for the fix :-)

taspelund (approved):

PR #743, opened 2026-05-20
bump to opte version 41, add v4 over v6 static routes

rcgoodfellow:
## OPTE Version Bump

OPTE version bump to go along with:

- oxidecomputer/opte#984

## Fix Omicron Dependency Loop

PR #720 introduced a dependency from `mg-admin-client` to `mg-common` which depends on a large number of omicron crates. Since omicron imports `mg-admin-client` this creates a dependency mess in omicron. So I'm cutting off the dependency here and adding a CI test that ensures none of the client or API type crates can depend on omicron.

Also adds a CI script to ensure that an omicron dependency loop does not creep back in.

## Support for Static IPv4 Routes Over IPv6 Nexthops

This was an oversight in the initial IPv4 over IPv6 routing support.


taspelund:
Looks good to me overall. I asked claude to do a quick spotcheck and it had a couple bits of feedback on the CI script.
- No set -euo pipefail. Without pipefail, cargo tree | grep omicron masks cargo tree failures — if cargo tree errors out (lockfile drift, network) and prints nothing to stdout, grep returns 1 and the check "passes". Add set -euo pipefail and run cargo tree to a file (or cargo tree --quiet --prefix none) before grepping.
- cargo tree without --all-features / per-feature sweep can miss feature-gated omicron deps. Consider cargo tree -e normal --all-features or iterating features.
- pushd $1 unquoted; doesn't verify the dir exists. Minor.
- The check covers mg-admin-client, ddm-admin-client, ddm-api-types, mg-api-types. Worth confirming mg-api-types-versions (now a workspace dep of mg-admin-client) is also clean, or add it to the list.

mg-api-types/versions/src/lib.rs
```diff
@@ -33,6 +33,8 @@ mod impls;
 pub mod latest;
 #[path = "initial/mod.rs"]
 pub mod v1;
+#[path = "v4_over_v6_static_routes/mod.rs"]
+pub mod v10;
```
taspelund, line 37: Is cargo fmt sorting this in lexicographical order instead of numerical? or can we adjust to use numerical order for readability?

taspelund (approved):
v4 over v6 work looks clean to me. only one nit about module sorting order in mg-api-types, but I'm not feeling overly motivated to fight rustfmt/cargo fmt if it's not a simple change

PR #746, opened 2026-05-21
Add shellcheck CI job

taspelund:
Adds a CI job to shellcheck CI jobs. Also fixes up some existing shellcheck errors in the scripts.

### DIFF
--- a/.github/buildomat/jobs/shellcheck.sh
+++ b/.github/buildomat/jobs/shellcheck.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+#:
+#: name = "linux"
+#: variety = "basic"
+#: target = "ubuntu-22.04"
+#: rust_toolchain = "stable"
+#: output_rules = [
+#:   "/work/debug/*",
+#:   "/work/release/*",
+#: ]
+#:
+
+banner "packages"
+sudo apt update -y
+sudo apt install -y shellcheck
+
+banner "shellcheck"
+failed=0
+for script in .github/buildomat/jobs/*.sh; do
+    if ! shellcheck -x "$script"; then
+        echo "$script failed shellcheck" >&2
+        failed=1
+    fi
+done
+(( failed == 0 )) || exit 1
--- a/.github/buildomat/jobs/shellcheck.sh
+++ b/.github/buildomat/jobs/shellcheck.sh
@@ -5,7 +5,7 @@
 #: target = "ubuntu-24.04"
 
 banner "packages"
-sudo apt update -y
+sudo apt update
 sudo apt install -y shellcheck
 
 banner "shellcheck"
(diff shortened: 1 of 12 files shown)


.github/buildomat/jobs/shellcheck.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+#:
+#: name = "shellcheck"
+#: variety = "basic"
+#: target = "ubuntu-22.04"
```
rcgoodfellow: Nit we should generally use the latest LTS for linux runners.

.github/buildomat/jobs/shellcheck.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+#:
+#: name = "shellcheck"
+#: variety = "basic"
+#: target = "ubuntu-24.04"
+
+banner "packages"
+sudo apt update -y
```
zeeshanlakhani: I'm not the best at shellfu, but is "-y" needed for update?

.github/buildomat/jobs/shellcheck.sh
```diff
@@ -0,0 +1,19 @@
...
+sudo apt install -y shellcheck
+
+banner "shellcheck"
+failed=0
+for script in .github/buildomat/jobs/*.sh; do
+    if ! shellcheck -x -o all "${script}"; then
```
zeeshanlakhani, line 14: double checking, we're good with all optional checks on here? i.e. "-o"?

.github/buildomat/jobs/shellcheck.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+#:
+#: name = "shellcheck"
+#: variety = "basic"
+#: target = "ubuntu-24.04"
+
```
zeeshanlakhani, line 6: isn't `set -o pipefail` needed here?

.github/buildomat/jobs/shellcheck.sh
taspelund: Yeah, I deliberately enabled it so we could have the tightest linting

.github/buildomat/jobs/shellcheck.sh
taspelund: It's not in this script because it doesn't use pipes.

My understanding of pipefail is that it modifies the return code of a pipeline so that it propagates the final/right-most non-zero rc, rather than the final/right-most rc (zero or non-zero)
i.e.
It makes sure that a non-zero rc in the middle of a pipeline doesn't get overridden by a successful rc later in the pipeline.

So I don't think it's relevant for this particular script

.github/buildomat/jobs/shellcheck.sh
taspelund: Doesn't seem to be needed for update, although I know it's needed for install. I can yank it

zeeshanlakhani (approved):
There may be better shellfu reviewers, but thanks for my answering my Q's.

PR #747, opened 2026-05-21
falcon-lab: bump ox to helios-3.0, add diag collection

taspelund:


falcon-lab/src/eos.rs
```diff
@@ -64,6 +64,31 @@ impl EosNode {
         LinuxNode(self.0)
     }
 
+    /// Capture BGP / BFD / routing state via the Arista CLI.
+    pub async fn collect_diagnostics(&self, d: &Runner, topo: &str) {
+        let name = self.name(d);
+        // `Cli -c` takes a single newline-separated script.
+        const SCRIPT: &str = "enable\n\
```
rcgoodfellow: I don't think these literal newlines are needed. See 

https://github.com/oxidecomputer/maghemite/blob/2aa749a2e4184b69f4e7362f90b1e7d183ee8c57/falcon-lab/src/test.rs#L487-L517

falcon-lab/src/diagnostics.rs
```diff
@@ -0,0 +1,44 @@
...
+    topo: &str,
+    label: &str,
+    source: Option<&str>,
+    contents: &str,
+) {
+    let header = match source {
+        Some(s) => format!("=== {topo} {label} ({s}) ==="),
+        None => format!("=== {topo} {label} ==="),
+    };
+    warn!(d.log, "{header}\n{contents}");
```
zeeshanlakhani: should we be `warn!`ing here when there may not be an error?

falcon-lab/src/dendrite.rs
```diff
@@ -10,9 +10,24 @@ use dpd_client::{
 };
 use libfalcon::{NodeRef, Runner};
 use slog::{Logger, debug, info};
-use std::{net::IpAddr, sync::Arc, time::Duration};
+use std::{error::Error as StdError, net::IpAddr, sync::Arc, time::Duration};
 use tokio::time::{Instant, sleep};
 
+/// Format an error along with its full `source()` chain. The default `Display`
```
zeeshanlakhani: nit: the doc comment references error types that are motivating examples, but the fn signature accepts any `&(dyn StdError + 'static)`. It's probably worth tightening the comment to make that explicit.

falcon-lab/src/eos.rs
taspelund: cool, that's a lot cleaner. thanks!

falcon-lab/src/test.rs
```diff
@@ -114,7 +136,58 @@ where
...
+            topo_name: topo_name.to_string(),
+        }),
+        Err(e) => {
+            collect_diagnostics(&ad, ox, cr1, cr2, topo_name).await;
+            Err(e)
+        }
+    }
+}
+
+#[allow(clippy::too_many_arguments)]
```
zeeshanlakhani: With the clippy allowance override, we may want to consider extracting a `TrioBuild` struct for the three `*_commit` args. Both `boot_trio` and  `boot_trio_inner` (and the `run_*` callers) pass the same cluster, so there's some sharing that can take place. ... And we could remove the clippy override.

zeeshanlakhani (approved):

rcgoodfellow (approved):

zeeshanlakhani:
@taspelund I pushed up the lab-3.0-gimlet buildomat target we'll need, but it needs to completed upstream first. Once that happens, a re-run will make this pass.

PR #758, opened 2026-05-29
DSCP for BGP/BFD, Default TTL of 255 for BGP, Graceful config updates for DSCP/TTL in BGP [Spring Cleaning 3/N]

taspelund:
Sets outbound TTL to 255 on BGP sessions by default (no min_ttl).

Adds support for DSCP/Traffic Class marking of IPv4/IPv6 packets carrying BGP/BFD.
QoS mark is configurable per peer and defaults to CS6 (Network Control) as recommended by RFC 4594 for routing protocols and related traffic.

Adds graceful config updates for min_ttl, DSCP and enforce-first-as in BGP.

Adds integration tests for DSCP/TTL updates in BGP.


taspelund:
Confirmed via manual testing that QoS markings and default outbound TTL are correctly applied, and that updates to the policy don't trigger the session to drop immediately (e.g. setting min_ttl=1 is always safe for inbound packets and the single-hop session didn't drop).  Moving this out of draft status

mg-api-types/versions/src/dscp/bgp/config.rs
```diff
@@ -0,0 +1,446 @@
...
+        Self {
+            hold_time,
+            idle_hold_time,
+            delay_open,
+            connect_retry,
+            keepalive,
+            resolution,
+            passive,
+            remote_asn,
+            min_ttl: min_ttl.and_then(NonZeroU8::new),
```
ahl, line 165: what would happen previously with a value of zero for `min_ttl`? It looks like 0 is now converted to 255 effectively.

taspelund: Yes, 0 is deliberately converted to 255 now.

The old behavior was to pass the min_ttl to setsockopt directly (0-255). I believe the intent of the original implementation was to have 0 act like a "reset" sentinel value, but there doesn't seem to be a portable way to do that via setsockopt.

So the new behavior is to say 255 is the default (enforced via setsockopt), and unconfiguring min_ttl resets the setting back to the default.

PR #770, opened 2026-06-01
consolidate ddm protocol types into a dedicated crate

ahl:
The frozen DDMv2 protocol types (UpdateV2, PullResponseV2, UnderlayUpdateV2, TunnelUpdateV2, PathVectorV2, TunnelOriginV2, IpPrefix, Ipv4Prefix, Ipv6Prefix) were scattered across ddm-api-types, mg-common, and ddm/src/exchange/mod.rs. Collect them all into a single dedicated module with a module-level comment making the freeze guarantee impossible to miss.

Also collect v3 protocol into a single location.

Add an expectorate snapshot test over the JSON schemas for the two top-level message types (Update, PullResponse) so that any accidental change to the frozen wire format fails loudly.

Also remove the DDMv1 conversion helpers (UpdateV1 and its From impls) that are no longer needed.

Note that some of the doc tests suddenly started to try to run and flagged some errors; I don't know why they hadn't been running previously. I verbatim-ified some ASCII art and corrected a path in another test.


taspelund (approved):
Looks like a nice clean change. I especially like the expectorate tests, that's not a tool I realized was in the tool belt. LGTM

ahl:
`expectorate` was basically the first thing I built at Oxide--it's been much more useful than anticipated!

rcgoodfellow (approved):
LGTM. Vastly better organization. Thanks!

Issue #767, opened 2026-06-01
Update to latest `oxnet` and use new `SocketAddrJson` type(s) in API


jgallagher:
(This fell out of proptesting the new MGD BGP scrimlet reconciler.)

omicron currently has a bug if attempting to use an IPv6 numbered BGP peer; it formats the `BgpPeerConfig::host` field via [`format!("{ip}:179"}`](https://github.com/oxidecomputer/omicron/blob/ff546dc31ec35a9d336d66f00cb60707bfa4cd6d/nexus/src/app/background/tasks/sync_switch_configuration.rs#L820), which is incorrect for IPv6 addresses since that omits the necessary `[]`s around the IP. 

We can fix this in omicron by changing that format string to `SocketAddr::new(ip, 179).to_string()`, but even better would be if the client saw the field as a `SocketAddr` instead of a `String`. maghemite's API is already doing the right thing here, as best it can, and defines this field as a `SocketAddr`. However, as of https://github.com/oxidecomputer/oxnet/pull/82 (published today in oxnet 0.1.6), if we change this field type to `SocketAddrJson` and add this to the `mg-admin-client` progenitor spec:

```rust
    crates = {
        std = "1.0.0",
    },
```

then the client will know the field needs to be a `SocketAddr`.

taspelund:
@ahl did #771 address this? If so, are we ready to resolve the issue?

ahl:
yep

PR #771, opened 2026-06-02
Migrate Prefix{,4,6} types to oxnet types

ahl:
Migrate the Prefix types to oxnet... although we still need them for backward compatibility with old versions of the API. This required a new version of the mg-admin API so I rolled in a fix for #767 while I was here.


bgp/src/messages.rs
```diff
@@ -110,10 +109,10 @@ impl BgpWireFormat<Prefix4> for Prefix4 {
         let len = input[0];
 
         // Validate length bound for IPv4 (structural validation)
-        if len > Prefix4::HOST_MASK {
+        if len > 32u8 {
```
taspelund: Does oxnet have a constant we could use here? I'd prefer to avoid the magic number even if it's just for copy/paste safety

bgp/src/messages.rs
```diff
@@ -174,10 +174,10 @@ impl BgpWireFormat<Prefix6> for Prefix6 {
         let len = input[0];
 
         // Validate length bound for IPv6 (structural validation)
-        if len > Prefix6::HOST_MASK {
+        if len > 128u8 {
```
taspelund: Same q here

bgp/src/messages.rs
ahl: oops; it does. will fix.

falcon-lab/Cargo.toml
```diff
@@ -19,3 +19,4 @@ oxide-tokio-rt.workspace = true
 bgp.workspace = true
 colored.workspace = true
 mg-api-types.workspace = true
+mg-api-types-versions.workspace = true
```
taspelund: Why is this needed?

mg-admin-client/src/lib.rs
```diff
@@ -16,11 +16,11 @@ progenitor::generate_api!(
         slog::trace!(log, "client response"; "result" => ?result);
     }),
     derives = [schemars::JsonSchema],
+    crates = {
+        "oxnet" = "0.1.5",
```
taspelund: This is slick. How do we keep this version in sync with the version that the crate/workspace depends on?

ahl: It's not a big deal if it falls behind. All this version says is: don't use types that predate this version. So far, all the oxnet types are 0.1.0. If we start using a new type in oxnet we'll bump the version of the crate and need to bump this as well.

falcon-lab/Cargo.toml
ahl: it wasn't!

ahl (commented):
self-review; changes incoming

bgp/src/policy.rs
```diff
@@ -239,11 +236,11 @@ pub fn new_rhai_engine() -> Engine {
...
 
     engine
-        .register_type_with_name::<Prefix>("Prefix")
+        .register_type_with_name::<IpNet>("Prefix")
         .register_fn("within", prefix_within_rhai)
-        .register_type_with_name::<Prefix4>("Prefix4")
+        .register_type_with_name::<Ipv4Net>("Prefix4")
         .register_fn("within", prefix4_within_rhai)
-        .register_type_with_name::<Prefix6>("Prefix6")
+        .register_type_with_name::<oxnet::Ipv6Net>("Prefix6")
```
jgallagher, line 243: Is it fine that the Rust type name doesn't match the Rhai type name anymore? (I have basically 0 context for the Rhai bits.)

mg-api-types/versions/src/impls/rdb/prefix.rs
```diff
@@ -263,6 +249,27 @@ impl std::fmt::Display for Prefix {
     }
 }
 
+impl From<oxnet::IpNet> for Prefix {
```
jgallagher: Are these conversions purely for backwards compatibility with the last API version that included `Prefix`? I _think_ these should move to the `v1` module (i.e., the latest version where `Prefix` was defined); I generally expect the `impls` module to only import things from `latest`, and i don't want anyone to see these conversions and think `Prefix` is still a type that's available for use.

mg-api-types/versions/src/initial/bgp/messages.rs
```diff
@@ -1024,9 +1024,9 @@ pub struct RouteRefreshMessage {
 )]
 #[schemars(rename = "UpdateMessage")]
 pub struct UpdateMessage {
```
jgallagher: Should this be on the struct too and not just the fields?

```suggestion
pub(crate) struct UpdateMessage {
```

mg-lower/src/dendrite.rs
```diff
@@ -56,17 +55,12 @@ impl RouteHash {
...
         sw: &impl SwitchZone,
-        prefix: Prefix,
+        prefix: IpNet,
         path: Path,
     ) -> Result<RouteHash, Error> {
         let (port_id, link_id) = get_port_and_link(sw, &path)?;
 
-        let cidr = match prefix {
-            Prefix::V4(p) => Ipv4Net::new(p.value, p.length)?.into(),
-            Prefix::V6(p) => Ipv6Net::new(p.value, p.length)?.into(),
```
jgallagher, line 66: Since we dropped these conversions, do we have any now-unused `Error` variants that they used?

mgd/src/validation.rs
```diff
@@ -325,24 +342,16 @@ mod tests {
     use proptest::prelude::*;
 
     // Strategy for generating any IPv4 prefix (normalized or unnormalized)
-    fn any_ipv4_prefix_strategy() -> impl Strategy<Value = Prefix4> {
+    fn any_ipv4_prefix_strategy() -> impl Strategy<Value = Ipv4Net> {
         (any::<u32>(), 0u8..=32u8).prop_map(|(addr_bits, length)| {
-            // Don't use new() - we want to test both normalized and unnormalized
```
jgallagher, line 330: I'd keep this comment (and the matching ipv6 one on line 341). It's a little repetitious with the comment on the function, but this is weird enough it seems worth it

mg-api/src/lib.rs
```diff
@@ -176,13 +177,85 @@ pub trait MgAdminApi {
...
+        method = DELETE,
+        path = "/bgp/config/neighbor/{asn}/{peer}",
+        operation_id = "delete_neighbor",
+        versions = VERSION_BGP_SRC_ADDR..VERSION_PREFIX_TO_OXNET,
+    }]
+    async fn delete_neighbor_v8(
+        rqctx: RequestContext<Self::Context>,
+        path: Path<latest::bgp::config::NeighborSelector>,
+    ) -> Result<HttpResponseDeleted, HttpError> {
+        Self::delete_neighbor(rqctx, path).await
```
jgallagher: This doesn't convert any types - did we need a new version of it?

Same question for most (all?) of the rest of the `delete_*` operations bumped below this

mg-api-types/versions/src/prefix_to_oxnet/bgp/history.rs
```diff
@@ -0,0 +1,56 @@
...
+
+impl From<v2::bgp::history::Origin6> for Origin6 {
+    fn from(old: v2::bgp::history::Origin6) -> Self {
+        // v2 is frozen; compile barrier protects against upstream field additions.
+        let v2::bgp::history::Origin6 { asn, prefixes } = old;
+        Self {
+            asn,
+            prefixes: prefixes
+                .into_iter()
+                .map(|p| Ipv6Net::new_unchecked(p.value, p.length))
```
jgallagher: Is new_unchecked safe here? I don't think the old Prefix* types validated their length

Edit: Same question in a bunch of other `From<OldType> for NewType` conversions

jgallagher (commented):
LGTM, mostly just nits. One nontrivial concern is about using `new_unchecked()` in converting from old API request versions, although switching all of those to `TryFrom` may be pretty annoying :-/

bgp/src/policy.rs
ahl: I assumed it was better to be backward compatible with (what I assume is) our vast library of rhai scripts. @rcgoodfellow may have thoughts.

mg-api-types/versions/src/impls/rdb/prefix.rs
ahl: Thanks for this. Moved.

mg-api-types/versions/src/initial/bgp/messages.rs
ahl: Yes; and I followed the chain back as far as I could with `pub(crate)`in things

mg-lower/src/dendrite.rs
ahl: great catch `OxnetIpnetPrefix`

mgd/src/validation.rs
ahl: The old comment had implicit in there that new would truncate/normalize the Prefix. IpNet does not do that so I thought the comment wasn't applicable.

mg-api/src/lib.rs
ahl: nope; deleted most of those. Needed v1. Nice find.

mg-api-types/versions/src/prefix_to_oxnet/bgp/history.rs
ahl: I believe so. It's an invariant of the Prefix* types that the length is valid; do you think it would be better to new().unwrap()?

bgp/src/policy.rs
rcgoodfellow: The Rhai stuff has not made it into production interfaces yet. I don't think we need to hold this up for Rhai considerations.

mg-api-types/versions/src/prefix_to_oxnet/bgp/history.rs
jgallagher: > It's an invariant of the Prefix* types that the length is valid

Is it? It doesn't look like they enforce _any_ invariants; the fields are `pub` and they derive `Deserialize`, so I think we could get any `u8` value here, right?

mg-api-types/versions/src/initial/rdb/prefix.rs
```diff
@@ -39,3 +44,333 @@ pub enum Prefix {
     V4(Prefix4),
     V6(Prefix6),
 }
+
+impl PartialOrd for Prefix4 {
```
rcgoodfellow, line 48: Looks like this is a direct lift from what was previously in mg-api-types/versions/src/impls/rdb/prefix.rs?

rcgoodfellow (commented):
Broadly looks good to me. Since this has a bunch of API surface change associated with it let's get an associated omicron PR up to see how that goes.

mg-api-types/versions/src/prefix_to_oxnet/bgp/history.rs
ahl: Yeah. I should have stated this differently: this use of new_unchecked preserves existing behavior. I could unwrap() or we could migrate to TryFrom. I'm not sure what would be best.

mg-api-types/versions/src/initial/rdb/prefix.rs
ahl: Exactly. As @jgallagher noted, the impls should pertain only to latest.

https://github.com/oxidecomputer/maghemite/blob/951f759b1d8e61cf65c2a823e61933ff3ea0bdf1/mg-api-types/versions/src/impls/rdb/prefix.rs#L16-L20

mg-api-types/versions/src/prefix_to_oxnet/bgp/history.rs
jgallagher: Ahh, gotcha. In principle I vote for `TryFrom`: we're getting outside data that may not be valid, and allowing construction of an `IpNet` (which itself _does_ enforce invariants, as long as nobody bypasses them with `new_unchecked()`) that could have an invalid length seems pretty bad?

If the `TryFrom`s are atrocious in practice, I could see an argument for `new().unwrap()` under the claim that mgd only gets requests from other omicron services, which should be using progenitor clients, which means they should only be sending valid `IpNet`s.

ahl:
@rcgoodfellow seems to work: https://github.com/oxidecomputer/omicron/pull/10558

PR #773, opened 2026-06-02
bgp_apply fixes

### LINKED ISSUE #772: Changing ASN for an unnumbered peer?
(This fell out of writing proptests for the scrimlet-reconcilers. I'm using the Linux build of `mgd` on commit 7696ee4, because that matches the commit omicron is on where I'm doing dev. From a little snooping around it doesn't seem like the behavior I'm describing here would be different between Linux and illumos, but for full disclosure I haven't tested this on illumos. I'm describing the behavior below as "Nexus's behavior", because it is, but the scrimlet reconciler work is moving this from Nexus to sled-agent. I don't think that changes anything other than the noun of the client involved.)

The current Nexus `sync_switch_configuration` background task has a couple of properties related to BGP configuration that are relevant to this issue:

1. It only supports a single `BgpConfig` (and therefore a single ASN) for a given switch, and therefore a given mgd instance (https://github.com/oxidecomputer/omicron/blob/4c4f4e1da0fb71781f7763a5003c7b6b186dc991/nexus/src/app/background/tasks/sync_switch_configuration.rs#L539-L540)
2. The only mgd APIs it calls to sync BGP configuration are `bgp_apply_v2()` + `update_bestpath_fanout()`, and it assumes those will ensure that mgd's config is updated to match the desired config (https://github.com/oxidecomputer/omicron/blob/4c4f4e1da0fb71781f7763a5003c7b6b186dc991/nexus/src/app/background/tasks/sync_switch_configuration.rs#L987-L1008). This issue is only concerned with `bgp_apply_v2()`.

I think either 2 is incorrect or mgd's handling of `bgp_apply_v2` with unnumbered peers is behaving incorrectly or both.

To reproduce what I'm seeing, we start from an mgd configuration where there is a single configured unnumbered peer on qsfp0:

```
% cat req123.json
{
  "asn": 123,
  "originate": [],
  "peers": {},
  "unnumbered_peers": {
    "qsfp0": [
      {
        "communities": [],
        "connect_retry": 3,
        "connect_retry_jitter": {
          "max": 1.0,
          "min": 0.75
        },
        "delay_open": 0,
        "deterministic_collision_resolution": false,
        "enforce_first_as": false,
        "hold_time": 6,
        "idle_hold_time": 3,
        "interface": "tfportqsfp0_0",
        "ipv4_unicast": {
          "export_policy": "NoFiltering",
          "import_policy": "NoFiltering"
        },
        "ipv6_unicast": {
          "export_policy": "NoFiltering",
          "import_policy": "NoFiltering"
        },
        "keepalive": 2,
        "name": "unnumbered-qsfp0",
        "passive": false,
        "resolution": 100,
        "router_lifetime": 1
      }
    ]
  }
}
```

We can apply this config via `curl` then confirm that we see a router for ASN 123 and the unnumbered neighbor we expect under ASN 123:

```
% curl  -X POST -H 'content-type: application/json' -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/omicron/apply' --data @req123.json
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/routers'
[{"asn":123,"id":123,"listen":"[::]:179","graceful_shutdown":false}]
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/unnumbered-neighbors?asn=123'
[{"asn":123,"name":"unnumbered-qsfp0","group":"qsfp0","interface":"tfportqsfp0_0","act_as_a_default_ipv6_router":1,"hold_time":6,"idle_hold_time":3,"delay_open":0,"connect_retry":3,"keepalive":2,"resolution":100,"passive":false,"remote_asn":null,"min_ttl":null,"md5_auth_key":null,"multi_exit_discriminator":null,"communities":[],"local_pref":null,"enforce_first_as":false,"vlan_id":null,"ipv4_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"ipv6_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"deterministic_collision_resolution":false,"idle_hold_jitter":null,"connect_retry_jitter":{"min":0.75,"max":1.0},"src_addr":null,"src_port":null}]
```

Now suppose the desired config changes in two ways: we want to use ASN 456 and we want to bump `hold_time` from 6 to 10. From point 2 above, Nexus will attempt to configure this by only sending a new apply request containing the desired config:

```
% cat req456.json
{
  "asn": 456,
  "originate": [],
  "peers": {},
  "unnumbered_peers": {
    "qsfp0": [
      {
        "communities": [],
        "connect_retry": 3,
        "connect_retry_jitter": {
          "max": 1.0,
          "min": 0.75
        },
        "delay_open": 0,
        "deterministic_collision_resolution": false,
        "enforce_first_as": false,
        "hold_time": 10,
        "idle_hold_time": 3,
        "interface": "tfportqsfp0_0",
        "ipv4_unicast": {
          "export_policy": "NoFiltering",
          "import_policy": "NoFiltering"
        },
        "ipv6_unicast": {
          "export_policy": "NoFiltering",
          "import_policy": "NoFiltering"
        },
        "keepalive": 2,
        "name": "unnumbered-qsfp0",
        "passive": false,
        "resolution": 100,
        "router_lifetime": 1
      }
    ]
  }
}
% diff req123.json req456.json
2c2
<   "asn": 123,
---
>   "asn": 456,
17c17
<         "hold_time": 6,
---
>         "hold_time": 10,
```

If we replicate Nexus's behavior with `curl`, `mgd` gives us back a 204 success:

```
% curl -v -X POST -H 'content-type: application/json' -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/omicron/apply' --data @req456.json
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying [::1]:8889...
* Connected to ::1 (::1) port 8889
> POST /bgp/omicron/apply HTTP/1.1
> Host: [::1]:8889
> User-Agent: curl/8.5.0
> Accept: */*
> content-type: application/json
> api-version: 8.0.0
> Content-Length: 812
>
< HTTP/1.1 204 No Content
< x-request-id: 36e5b55b-c205-4b70-9487-9478654b1afe
< date: Tue, 02 Jun 2026 20:03:12 GMT
```

We do see the new ASN 456 in the list of routers:

```
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/routers'
[{"asn":123,"id":123,"listen":"[::]:179","graceful_shutdown":false},{"asn":456,"id":456,"listen":"[::]:179","graceful_shutdown":false}]
```

But we _don't_ see any neighbors in ASN 456:

```
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/unnumbered-neighbors?asn=456'
[]
```

And more surprisingly, the neighbor in ASN 123 now has the new hold time of 10:

```
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/unnumbered-neighbors?asn=123'
[{"asn":123,"name":"unnumbered-qsfp0","group":"qsfp0","interface":"tfportqsfp0_0","act_as_a_default_ipv6_router":1,"hold_time":10,"idle_hold_time":3,"delay_open":0,"connect_retry":3,"keepalive":2,"resolution":100,"passive":false,"remote_asn":null,"min_ttl":null,"md5_auth_key":null,"multi_exit_discriminator":null,"communities":[],"local_pref":null,"enforce_first_as":false,"vlan_id":null,"ipv4_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"ipv6_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"deterministic_collision_resolution":false,"idle_hold_jitter":null,"connect_retry_jitter":{"min":0.75,"max":1.0},"src_addr":null,"src_port":null}]
```

I'm not sure what the behavior of this kind of apply ought to be, if we're changing the configuration of an unnumbered peer, but I think either of these would be less surprising:

* Return an error and require the client to remove the old config first (this would be pretty annoying for Nexus, I suspect)
* Return success and ensure that the configuration of this port / peer is now attached to the correct ASN

I'm also not sure what Nexus is expected to do to apply this kind of change today, or what it may need to do if the change is more complicated (especially if we relax the "only support at most one ASN per mgd" item). @taspelund suggested deleting the ASN=123 router first, but that leaves the neighbors behind:

```
# starting back from the initial config of the peer only in ASN=123
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/routers'
[{"asn":123,"id":123,"listen":"[::]:179","graceful_shutdown":false}]
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/unnumbered-neighbors?asn=123'
[{"asn":123,"name":"unnumbered-qsfp0","group":"qsfp0","interface":"tfportqsfp0_0","act_as_a_default_ipv6_router":1,"hold_time":6,"idle_hold_time":3,"delay_open":0,"connect_retry":3,"keepalive":2,"resolution":100,"passive":false,"remote_asn":null,"min_ttl":null,"md5_auth_key":null,"multi_exit_discriminator":null,"communities":[],"local_pref":null,"enforce_first_as":false,"vlan_id":null,"ipv4_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"ipv6_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"deterministic_collision_resolution":false,"idle_hold_jitter":null,"connect_retry_jitter":{"min":0.75,"max":1.0},"src_addr":null,"src_port":null}]
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/unnumbered-neighbors?asn=456'
[]

# this delete returns a 204 and we indeed see the router gone...
% curl -X DELETE -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/router?asn=123'
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/routers'
[]

# ...but the unnumbered peer neighbor is still there (?)
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/unnumbered-neighbors?asn=123'
[{"asn":123,"name":"unnumbered-qsfp0","group":"qsfp0","interface":"tfportqsfp0_0","act_as_a_default_ipv6_router":1,"hold_time":6,"idle_hold_time":3,"delay_open":0,"connect_retry":3,"keepalive":2,"resolution":100,"passive":false,"remote_asn":null,"min_ttl":null,"md5_auth_key":null,"multi_exit_discriminator":null,"communities":[],"local_pref":null,"enforce_first_as":false,"vlan_id":null,"ipv4_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"ipv6_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"deterministic_collision_resolution":false,"idle_hold_jitter":null,"connect_retry_jitter":{"min":0.75,"max":1.0},"src_addr":null,"src_port":null}]
```

I also tried posting an apply request for ASN 123 with an empty set of peers+unnumbered peers, but that doesn't seem like it does anything at all despite getting back a 204 success:

```
# starting back from the initial config of the peer only in ASN=123
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/routers'
[{"asn":123,"id":123,"listen":"[::]:179","graceful_shutdown":false}]
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/unnumbered-neighbors?asn=123'
[{"asn":123,"name":"unnumbered-qsfp0","group":"qsfp0","interface":"tfportqsfp0_0","act_as_a_default_ipv6_router":1,"hold_time":6,"idle_hold_time":3,"delay_open":0,"connect_retry":3,"keepalive":2,"resolution":100,"passive":false,"remote_asn":null,"min_ttl":null,"md5_auth_key":null,"multi_exit_discriminator":null,"communities":[],"local_pref":null,"enforce_first_as":false,"vlan_id":null,"ipv4_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"ipv6_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"deterministic_collision_resolution":false,"idle_hold_jitter":null,"connect_retry_jitter":{"min":0.75,"max":1.0},"src_addr":null,"src_port":null}]

# send a request for ASN 123 with no peers or unnumbered peers
% cat req123-empty.json
{
  "asn": 123,
  "originate": [],
  "peers": {},
  "unnumbered_peers": {}
}

# we get back a 204...
% curl -v -X POST -H 'content-type: application/json' -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/omicron/apply' --data @req123-empty.json
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying [::1]:8889...
* Connected to ::1 (::1) port 8889
> POST /bgp/omicron/apply HTTP/1.1
> Host: [::1]:8889
> User-Agent: curl/8.5.0
> Accept: */*
> content-type: application/json
> api-version: 8.0.0
> Content-Length: 71
>
< HTTP/1.1 204 No Content
< x-request-id: 7e9784fc-8ebf-43dd-a17c-4453c14ab9c8
< date: Tue, 02 Jun 2026 20:17:08 GMT

# but the router and neighbor are still there
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/routers'
[{"asn":123,"id":123,"listen":"[::]:179","graceful_shutdown":false}]
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/unnumbered-neighbors?asn=123'
[{"asn":123,"name":"unnumbered-qsfp0","group":"qsfp0","interface":"tfportqsfp0_0","act_as_a_default_ipv6_router":1,"hold_time":6,"idle_hold_time":3,"delay_open":0,"connect_retry":3,"keepalive":2,"resolution":100,"passive":false,"remote_asn":null,"min_ttl":null,"md5_auth_key":null,"multi_exit_discriminator":null,"communities":[],"local_pref":null,"enforce_first_as":false,"vlan_id":null,"ipv4_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"ipv6_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"deterministic_collision_resolution":false,"idle_hold_jitter":null,"connect_retry_jitter":{"min":0.75,"max":1.0},"src_addr":null,"src_port":null}]
```

### LINKED ISSUE #783: mgd: 500 trying to create origin prefixes for new router after deleting an old router
Starting from a fresh `mgd` instance, I can create a router with ASN 1 and give it a prefix to originate:

```
% cat make-router-asn1.json
{
    "asn": 1,
    "id": 1,
    "listen": "[::1]:179",
    "graceful_shutdown": false
}
% curl -v -X PUT -H 'content-type: application/json' -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/router' --data @make-router-asn1.json
% cat make-origin4-asn1.json
{
    "asn": 1,
    "prefixes": [
        {"value": "1.2.3.4", "length": 31}
    ]
}
% curl -v -X PUT -H 'content-type: application/json' -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/origin4' --data @make-origin4-asn1.json
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/origin4?asn=1'
{"asn":1,"prefixes":[{"value":"1.2.3.4","length":31}]}
```

If I then delete the router, I get a 404 if asking for its origin prefixes, as expected:

```
% curl -X DELETE -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/router?asn=1'
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/origin4?asn=1'
{
  "request_id": "ed9628ff-d145-431e-991a-98017ac97cce",
  "message": "Not Found"
}
```

However, now if I create a router with ASN 2 and try to give it an origin4 prefix, I get back a 500:

```
% cat make-router-asn2.json
{
    "asn": 2,
    "id": 2,
    "listen": "[::1]:179",
    "graceful_shutdown": false
}
% curl -v -X PUT -H 'content-type: application/json' -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/router' --data @make-router-asn2.json
% cat make-origin4-asn2.json
{
    "asn": 2,
    "prefixes": [
        {"value": "5.6.7.8", "length": 31}
    ]
}
% curl -v -X PUT -H 'content-type: application/json' -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/origin4' --data @make-origin4-asn2.json
{
  "request_id": "94fc30a4-b35e-45d9-bb0a-7c05d7450f90",
  "error_code": "Internal",
  "message": "Internal Server Error"
}
```

The mgd logs claim there's a conflict:

```
19:05:22.729Z INFO slog-rs (mgd): request completed
    error_message_external = Internal Server Error
    error_message_internal = Datastore error: Conflict origin already exists
    latency_us = 464
    local_addr = [::1]:8889
    method = PUT
    module = admin
    remote_addr = [::1]:34436
    req_id = 94fc30a4-b35e-45d9-bb0a-7c05d7450f90
    response_code = 500
    unit = api_server
    uri = /bgp/config/origin4
```

taspelund:
Fixes a few issues in the bgp_apply API handler.

1) makes single ASN handling idempotent (a request with a new/different ASN results in all routers with a differing ASN being cleaned up)
2) cleans up neighbors associated with router being removed
3) moves neighbor deletion ahead of neighbor add, allowing a "move" from one ASN to another

Centralizes router deletion logic into do_delete_router() method.
Adds unit tests for do_delete_router() and for do_bgp_apply().

Fixes: #772
Fixes: #783


mgd/src/bgp_admin.rs
```diff
@@ -1235,56 +1266,36 @@ async fn do_bgp_apply(
 ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
     let log = ctx.log.clone();
 
-    // Validate originate prefixes before processing
-    validate_prefixes(&rq.originate)?;
```
jgallagher, line 1253: Why did this move to the end?

A general/vague concern with this function is: what happens if we bail out from a `?` partway through? It looks like that will have applied whatever things happened before - will that leave us in a state where a future apply can still work, or could we be in some weird / broken intermediate state?

mgd/src/bgp_admin.rs
```diff
@@ -2660,92 +2671,280 @@ mod tests {
...
+        do_bgp_apply(&ctx, req.clone())
             .await
             .expect("bgp apply request");
+        assert_eq!(
+            ctx.db.get_bgp_neighbors().expect("get bgp neighbors").len(),
+            1,
+        );
+    }
+
+    /// Regression test for https://github.com/oxidecomputer/maghemite/issues/772
```
jgallagher, line 2897: Thanks for these tests 👍

jgallagher (commented):
The changes LGTM, but I'd generally prefer if someone more familiar with mgd gave a final approval.

I did try all the reproduction steps from #772 and they all look good, with one possible caveat I wanted to ask about: If I have an unnumbered peer in ASN 123 like this:

```
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/routers'
[{"asn":123,"id":123,"listen":"[::]:179","graceful_shutdown":false}]
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/unnumbered-neighbors?asn=123'
[{"asn":123,"name":"unnumbered-qsfp0","group":"qsfp0","interface":"tfportqsfp0_0","act_as_a_default_ipv6_router":1,"hold_time":6,"idle_hold_time":3,"delay_open":0,"connect_retry":3,"keepalive":2,"resolution":100,"passive":false,"remote_asn":null,"min_ttl":null,"md5_auth_key":null,"multi_exit_discriminator":null,"communities":[],"local_pref":null,"enforce_first_as":false,"vlan_id":null,"ipv4_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"ipv6_unicast":{"nexthop":null,"import_policy":"NoFiltering","export_policy":"NoFiltering"},"deterministic_collision_resolution":false,"idle_hold_jitter":null,"connect_retry_jitter":{"min":0.75,"max":1.0},"src_addr":null,"src_port":null}]
```

and I apply a config that has ASN 123 with no peers:

```
% cat req123-empty.json
{
  "asn": 123,
  "originate": [],
  "peers": {},
  "unnumbered_peers": {}
}
% curl -v -X POST -H 'content-type: application/json' -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/omicron/apply' --data @req123-empty.json
```

then it does remove the neighbors as expected, but I'm still left with a router entry for 123:

```
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/unnumbered-neighbors?asn=123'
[]
% curl -H 'api-version: 8.0.0' 'http://[::1]:8889/bgp/config/routers'
[{"asn":123,"id":123,"listen":"[::]:179","graceful_shutdown":false}]
```

I _think_ that makes sense - I asked for a config to applied that has ASN 123 but has no peers. But is this a reasonable config at all?

The context for this is mostly: what should Nexus/sled-agent do if the desired config is "no BGP"? I think based on the above, instead of trying to apply a config (which requires an ASN), it should:

* Get the current routers (we expect to get back 0 or 1, given Nexus's current "at most one ASN per switch" constraint)
* Delete them all

Does that sound right?

mgd/src/bgp_admin.rs
taspelund: I moved it to the end because it seemed relevant to group it with the application of the originated routes.
That said, I do think we have a lot of early returns without cleanup... I'll see about doing request validation up front rather than along the way.

Question for you: are you okay with changes to the shape of this API? e.g. if I wanted to push some of the validations into the ApplyRequest type rather than putting logic into the handler, would that be ok for nexus?

taspelund:
> I think that makes sense - I asked for a config to applied that has ASN 123 but has no peers. But is this a reasonable config at all?
> 
> The context for this is mostly: what should Nexus/sled-agent do if the desired config is "no BGP"? I think based on the above, instead of trying to apply a config (which requires an ASN), it should:
> 
>     Get the current routers (we expect to get back 0 or 1, given Nexus's current "at most one ASN per switch" constraint)
>     Delete them all
> 
> Does that sound right?

I don't think the shape of the API today really provides a way to remove a BGP instance.  ApplyRequest has a mandatory ASN and no parameters for requesting deletion... plus there's no DELETE verb for it either.

So does it sound "right"? It's working as expected, but I think it would be better to change the shape to allow for a clean deletion instead of leaving inert config behind.  Thoughts?

rcgoodfellow:
> So does it sound "right"? It's working as expected, but I think it would be better to change the shape to allow for a clean deletion instead of leaving inert config behind. Thoughts?

`mgd` now has a complete REST API for all of the core router objects. The apply API was to enable compound operations with atomic semantics in the pre-RPW era of nexus. Since this is all driven from a nexus RPW now, that is no longer necessary. So I think rather than spending too much more energy on the apply API, we might consider moving nexus onto the REST API and sunsetting the apply API.

rcgoodfellow (approved):

jgallagher:
> The apply API was to enable compound operations with atomic semantics in the pre-RPW era of nexus. Since this is all driven from a nexus RPW now, that is no longer necessary. So I think rather than spending too much more energy on the apply API, we might consider moving nexus onto the REST API and sunsetting the apply API.

I've been working on switching the new sled-agent scrimlet reconcilers to use the REST API instead of the apply API. It's doable, but I don't think the reasons here are right, and I think in general we probably do want to prefer compound operations with atomic semantics over independent APIs. (We've had to do nontrivial work in Reconfigurator to change from individual imperative operations to larger, compound idempotent operations for correctness, similar to the REST vs apply question here.)

A couple of the problems we've had with individual APIs that could also apply here are:

1. Different Nexuses can duel, especially around the time settings change and different Nexus instances could be operating on different data
2. If there are dependencies between the different kinds of operations, the client has to understand them and order operations correctly. (In really bad cases there may be no way to order individual operations - we ran into this with some interactions between zones and datasets in sled-agent. I don't think that applies here, but it's some scar tissue.)

1 is my biggest concern with networking APIs generally (both in omicron and here). We need a way for a Nexus that's potentially operating on stale data to be unable to make changes that overwrite changes pushed by a Nexus operating on newer data. The most common pattern we use for that elsewhere is for Nexus to have a compound object with a generation number, and it sends that object to the server, which is responsible for (a) rejecting requests whose generation number is stale and (b) ensuring the settings described by the object are applied in an idempotent way.

In this specific case, we can sidestep this problem entirely, because with the work to move reconciliation from Nexus to the scrimlet sled-agent, there will only be a single mgd client, so none of this applies. But I wanted to jot these notes down while the thoughts were fresh.

taspelund:
I figure it is worth it to get the bgp_apply changes put in place even if we decide to go another route on the API endpoints for nexus.  The code is already written so there's no additional cost there, and we can at least plug the gaps in the interim before larger changes come in.  So I've pushed a few more commits that put the prefix validation back to the front of do_bgp_apply() and fix another bug related to old state not getting cleaned up upon BGP Router deletion (#783)

rcgoodfellow (commented):
For the ASN namspacing, while it's consistent with the current multi-instance router design that's keyed on ASN, I'm not sure if it's the direction we'll ultimately wind up heading in. From RFD 662 and in particular [this section](https://rfd.shared.oxide.computer/rfd/0662#_maghemite_bgp_instances), I think the direction we are going in is toward router instances that are tied to user defined configurations (likely keyed by a UUID that is managed by Omicron) where the same ASN may be used across configurations.

I'm not opposed to these ASN-based namespacing changes landing, as they're consistent with the current design. Just want to be clear that the direction is likely to change in the relatively near future.

taspelund:
Working through some more fun with a4x2 currently. All the zones deploy, nexus gets an external ip, nat entries are setup, but running snoop within the nexus zone doesn't show any packets arriving. BGP state looks good on both rack switches, cr1, cr2, and ce... so it's not clear yet why ingress traffic doesn't get to nexus. Possibly a firewall thing.

taspelund:
After changing basically nothing, a4x2 has started working with this maghemite hash... The baseline service thing seems like it may have been the culprit (or possibly some combination of changes where I missed an update, rebuild or something).

@rcgoodfellow this is ready for final review

rcgoodfellow (approved):

PR #775, opened 2026-06-05
unbreak omicron ci

rcgoodfellow:
Changing the linux build environment torpedoes omicron CI with

```
mgd: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by mgd)
```

### DIFF
--- a/.github/buildomat/jobs/linux.sh
+++ b/.github/buildomat/jobs/linux.sh
@@ -2,7 +2,7 @@
 #:
 #: name = "linux"
 #: variety = "basic"
-#: target = "ubuntu-24.04"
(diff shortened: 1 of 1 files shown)


taspelund (approved):

ahl:
Thanks for fixing this (Introduced by #746)

There used to be automation that would pull in every maghemite update to a PR in omicron. Do we know what happened to it? It looks like it stopped a year ago:

https://github.com/oxidecomputer/omicron/pull/8251
https://github.com/oxidecomputer/omicron/actions/runs/26994486380/job/79661309078

(will try to fix)

PR #785, opened 2026-06-11
clean up accumulated detritus

ahl:
This fixes two misconceptions I stumbled across in #771:
- there are a bunch of no-op renames with `schemars`
- there are bunch of interstitial versioned endpoints that didn't need to be created


taspelund (approved):
LGTM, thank you for the cleanup!

I also asked claude to look it over to see if there were any more instances of these anti-patterns that could be cleaned up and it seems to think you've exhaustively found all instances of them, so kudos ;-)

Ship it!

PR #792, opened 2026-06-16
Enforce MD5 auth string invariants via new type

### LINKED ISSUE #765: `mgd` panics on overlong MD5 auth key
(This fell out of proptesting the new MGD BGP scrimlet reconciler.)

The MGD API defines the MD5 auth key field as an `Option<String>`: https://github.com/oxidecomputer/maghemite/blob/7696ee48d5ee29a917dea459e281fe2e8ff20513/bgp/src/params.rs#L933

However, internally it assumes that key is at most `MAX_MD5SIG_KEYLEN` (80) bytes, and if given a key longer than that, we panic on this `copy_from_slice` on Linux: https://github.com/oxidecomputer/maghemite/blob/7696ee48d5ee29a917dea459e281fe2e8ff20513/bgp/src/connection_tcp.rs#L274-L276

There's no immediately-obvious equivalent on illumos, but it looks like mgd eventually passes the key down to libnet, which makes the same assumption: https://github.com/oxidecomputer/netadm-sys/blob/201992c327b69211c29f097cc767dc2801315fa3/libnet/src/pf_key.rs#L367-L377

I'm not sure what the relationship between mgd and libnet is - maybe libnet needs its own issue and should do validation? But mgd should reject these requests. I don't have a strong feeling whether the API type should remain `Option<String>` and mgd does validation on that string, or if we should have a stronger type here (`Option<Md5AuthKey>` or something like that) that does its own validation. The latter seems like a good idea on the surface, at least.

taspelund:
Intoduces Md5AuthString type to wrap a String behind a fallible
constructor and immutable getters. This ensures that an API client is
unable to configure an MD5 auth key that does not comply with the
inputs recommended by RFC 2385 (TCP MD5 option for BGP):
```
4.5 Key configuration

   It should be noted that the key configuration mechanism of routers
   may restrict the possible keys that may be used between peers.  It is
   strongly recommended that an implementation be able to support at
   minimum a key composed of a string of printable ASCII of 80 bytes or
   less, as this is current practice.
```

The Illumos and Linux kernel implementations both accept 80-byte keys
with arbitrary contents, so the printable ASCII constraint is purely
coming from RFC 2385 and the need for keys to round-trip through json
via OpenAPI / Dropshot.

Fixes: #765


taspelund:
The libnet dependency is also bumped in this ticket in order to pull in https://github.com/oxidecomputer/netadm-sys/pull/210. This gives us an additional layer of checking around the MD5 auth keys we submit to the kernel.

mg-api/src/lib.rs
```diff
@@ -515,12 +677,12 @@ pub trait MgAdminApi {
     }]
     async fn read_unnumbered_neighbors_v8(
         rqctx: RequestContext<Self::Context>,
-        request: Query<latest::bgp::config::AsnSelector>,
```
rcgoodfellow, line 562: Why this change? AFAICT `latest::bgp::config::AsnSelector` is an alias for `v1::bgp::config::AsnSelector`.

mg-api/src/lib.rs
```diff
@@ -546,10 +709,10 @@ pub trait MgAdminApi {
     }]
     async fn read_unnumbered_neighbor_v8(
         rqctx: RequestContext<Self::Context>,
-        request: Query<latest::bgp::config::UnnumberedNeighborSelector>,
+        request: Query<v5::bgp::config::UnnumberedNeighborSelector>,
```
rcgoodfellow, line 759: Same question here, why change from the latest alias to the explicitly numbered version?

mgd/src/bgp_admin.rs
```diff
@@ -208,10 +209,9 @@ pub async fn read_neighbors_v1(
         .into_iter()
         .filter(|x| x.asn == rq.asn)
         .map(|x| {
-            v8::bgp::config::Neighbor::from(Neighbor::from_rdb_neighbor_info(
-                rq.asn, &x,
-            ))
-            .into()
+            let latest = Neighbor::from_rdb_neighbor_info(rq.asn, &x);
+            let v11: v11::bgp::config::Neighbor = latest.into();
+            v8::bgp::config::Neighbor::from(v11).into()
```
rcgoodfellow: This is rather confusing code to read. Constructing a v11 Neighbor to turn into a v8 neighbor to then turn into a v1 neighbor for the return object. It seems like there should be `From`/`Into` impls that take care of conversion chains so conversions don't clutter up handler code.

mgd/src/bgp_admin.rs
```diff
@@ -244,10 +244,10 @@ pub async fn read_neighbor_v1(
             format!("neighbor {} not found in db", rq.addr),
         ))?;
 
-    let result: v1::bgp::config::Neighbor = v8::bgp::config::Neighbor::from(
-        Neighbor::from_rdb_neighbor_info(rq.asn, neighbor_info),
-    )
-    .into();
+    let latest = Neighbor::from_rdb_neighbor_info(rq.asn, neighbor_info);
```
rcgoodfellow: Same comment as above on conversion chains.

mgd/src/bgp_admin.rs
```diff
@@ -1954,7 +1966,38 @@ pub(crate) mod helpers {
             rq.parameters.allow_export.clone(),
         );
 
-        let info = SessionInfo::from(&rq.parameters);
+        let latest_parameters = BgpPeerParameters {
```
rcgoodfellow: This should be handled in a `From` impl for the `v1` to `latest` version of `BgpParameters`.

mg-api/src/lib.rs
jgallagher: I believe this is consistent with the API migration guide, specifically https://github.com/oxidecomputer/dropshot-api-manager/blob/main/guides/new-version.md#for-changed-and-removed-endpoints. Since this endpoint has been removed, its input/output types should never change; they're effectively pinned to whatever `latest` was at the point at which they were removed.

If `latest` is still an alias to `v1` in this case either will compile and pass the drift checks, but leaving `latest` in a removed endpoint is a bit of a landmine for future maintainers: if someone changes `AsnSelector` in the future, they now have to go back to all the API endpoints currently using `latest` and pin them to the previous-latest version by `vN`.

taspelund: The _vX suffixed methods exist to handle requests targeting old versions of an API endpoint. Latest is a live target, which means the underlying type could change in a way that affects the behavior of the API handler without causing a compilation failure. I moved these handlers for old versions over to using a pinned version of the types to avoid future issues if this were to change. I believe this should have been done during the RFD 619 migration but was missed due to the size of that change. I didn't bother breaking it out into its own PR because it didn't seem like a large amount of clutter, but I'm happy to do so if you'd prefer that.

mgd/src/bgp_admin.rs
jgallagher: The _general_ guidance from https://github.com/oxidecomputer/dropshot-api-manager/blob/main/guides/new-version.md#add-conversions-to-or-from-the-immediately-prior-version is to only define a conversion from the most recent version being replaced, but I think based on the end of that section it assumes we're usually only converting 1 step at a time in the API trait:

> In general, don't add From impls from other prior versions. (So, if a type changed from v1 to v4 to v9, avoid implementing conversions from v9 to v1 or vice versa.) Instead, hop through intermediate versions in the API trait. In some cases it may be more efficient to have direct conversions to prior versions; use appropriate judgment.

I assume we have this `read_neighbors_v1` because converting in the API trait isn't tenable for some reason? So maybe this is a case where appropriate judgment is to define conversions for a longer chain.

taspelund: I'm pretty sure RFD 619 says to only impl conversions between the newly introduced type version and the previous version type. The handlers for the legacy API versions would then carry the burden for walking the conversion chain since Rust can't just call .into() and have it implicitly use multiple From/Into conversions in one shot.

Edit: I think John and I were typing at the same time again

mg-api/src/lib.rs
taspelund: I think John and I were typing at the same time. I would say that where we want to start exercising judgement here is when to retire old API versions.

Afaik, the only consumers of mg-admin-client are nexus, mgadm, falcon-lab and possibly a couple spots in the testbed repo (a4x2 or interop). There's really no reason we need to maintain 10+ API versions when all the clients are under our purview

mgd/src/bgp_admin.rs
rcgoodfellow: Why don't we just implement this handler as something like:

```rust
    #[endpoint {
        method = GET,
        path = "/bgp/config/neighbors",
        operation_id = "read_neighbors",
        versions = ..VERSION_MP_BGP,
    }]
    async fn read_neighbors_v1(
        rqctx: RequestContext<Self::Context>,
        request: Query<v1::bgp::config::AsnSelector>,
    ) -> Result<HttpResponseOk<Vec<v1::bgp::config::Neighbor>>, HttpError> {
        Self::read_neighbors_v4(rqctx, request)
            .await
            .map(|r| r.map(|v| v.into_iter().map(Into::into).collect()))
    }
```

in `mg-api/src/lib.rs` like the other older versions of this function? Similar for the other `_v1` impls.

taspelund: I find this is harder to parse than the current code, but I don't have strong opinions about the style here. Happy to go this way if it's something you want changed though

rcgoodfellow: It's about more than style, it limits blast radius of endpoint interoperability complexity to just the API trait and `From`/`Into` impls that only have to operate across a single version bump.

taspelund: I think I see what you mean now. I'll push another commit to tidy this up

taspelund:
@rcgoodfellow @jgallagher I've rebased this on main and applied some fixes found during an AI code review.
If you guys have some time to look this over again, it would be much appreciated

mgd/src/admin.rs
```diff
@@ -221,37 +221,8 @@ impl MgAdminApi for MgAdminApiImpl {
...
-    async fn update_neighbor_v1(
-        ctx: RequestContext<Self::Context>,
-        request: TypedBody<v1::bgp::config::Neighbor>,
-    ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
-        bgp_admin::update_neighbor_v1(ctx, request).await
-    }
-
-    // delete_neighbor_v1 and clear_neighbor_v1 are provided methods.
+    // V1 (..VERSION_MP_BGP): create/read/update/delete and clear are all
+    // provided methods that hop to the v4 handlers.
```
jgallagher, line 225: I'm not sure this comment is useful / it seems like it'll be out of date next time there's a bump. I think it's standard for the API trait to forward to newer versions, so probably don't need to note it here.

mg-api/src/lib.rs
```diff
@@ -211,13 +212,92 @@ pub trait MgAdminApi {
...
+    async fn create_neighbor_v11(
+        rqctx: RequestContext<Self::Context>,
+        request: TypedBody<v11::bgp::config::Neighbor>,
+    ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
+        Self::create_neighbor(
+            rqctx,
+            request.try_map(TryInto::try_into).map_err(|e| {
+                HttpError::for_bad_request(
+                    None,
+                    format!("invalid MD5 authentication key: {e}"),
```
jgallagher, line 241: This isn't necessarily wrong, but having the "invalid MD5 auth key" string here seems like it's tying details that aren't relevant / visible in a place they shouldn't be. Could this either be

```suggestion
                    InlineErrorChain::new(&e).to_string(),
```

and we make sure the `Display` impl for the error type includes all necessary context, or implement `From<TheConversionError> for HttpError` and put this conversion closer to where the error is defined so it's more obviously correct? Either of these would also let us avoid duplicating this string in all the other endpoints that do this same conversion.

mg-api-types/versions/src/mp_bgp/bgp/config.rs
```diff
@@ -166,6 +166,207 @@ pub struct Neighbor {
...
+            remote_asn,
+            min_ttl,
+            md5_auth_key,
+            multi_exit_discriminator,
+            communities,
+            local_pref,
+            enforce_first_as,
+            vlan_id,
+            ipv4_unicast,
+            // dropped: no v1 representation.
```
jgallagher, line 223: Is it okay to unconditionally drop these fields, or should this be a `try_from` that fails if these are populated in a way that can't be represented in v1? (Although if we're on v12 it seems pretty unlikely anything with v1 will come up anyway, so maybe this doesn't matter.)

jgallagher (commented):
Changes LGTM, just a couple small nits.

mg-api-types/versions/src/mp_bgp/bgp/config.rs
taspelund: At this point the v1 API is getting pretty darn long in the tooth... It doesn't support IPv6 or BGP unnumbered, and it didn't seem like a good idea to completely filter all neighbors from a v1 GET just because a newer API client configured the neighbor with IPv6.

As a whole though, I think this is definitely one of those situations where we (maghemite in particular) have a ton of API versions without consumers that just need to be retired.

I believe the only consumers of mg-api are mgadm, falcon-lab and omicron... And omicron is the only one that doesn't live in this repo. I'd have to check with Justin to see if any of the rkdeploy and similar tools make use of mg-api, but it should be pretty easy to bump their API rev the same way we do with omicron.

mgd/src/admin.rs
taspelund: Sure, I can strip out the comment

nicolaskagami (commented):
Always happy to see us enforcing invariants through types :)

Changes look good and straightforward, though the amount of boilerplate we need to add is a bit unfortunate.

I'd like to see @jgallagher's points addressed before merging, namely the v1 concern and the `HttpError` `From` impl.

Besides that, I noticed that we're changing the types from underneath the Db so we'll lose persisted neighbors that don't have a valid `Md5AuthString` in `get_bgp_neighbors`. That's probably fine, but I thought I'd point it out.

mg-api-types/versions/src/mp_bgp/bgp/config.rs
taspelund: As discussed in matrix, I'll plan to convert these to a fallible conversion rather than stripping out the ipv6 info, then submit a follow-up PR that retires old API versions.

taspelund:
> I'd like to see @jgallagher's points addressed before merging, namely the v1 concern and the `HttpError` `From` impl.

These have been addressed in the most recent feedback commit.

> Besides that, I noticed that we're changing the types from underneath the Db so we'll lose persisted neighbors that don't have a valid `Md5AuthString` in `get_bgp_neighbors`. That's probably fine, but I thought I'd point it out.

Yes, it definitely will. However, since the prior behavior was for maghemite to panic and we never received a bug report on this... I think it's probably safe to say no one has (successfully) deployed an invalid key 😁

taspelund:
@nicolaskagami @jgallagher I've pushed another set of commits that address the latest round of feedback (HttpError conversion, v1 API type conversion fallibility) as well as a handful of issues identified during LLM code review.

There's a fair amount more lines that have been touched as a result of fixing some of the other type conversion issues that came up during the recent changes (e.g. reworking the rdb::Prefix <--> oxnet::IpNet type conversions to be fallible instead of calling .unwrap(), fixing some BGP capability type conversions to be fallible instead of mapping enum variants in an invalid way, etc.).

If you guys feel it adds too much noise into this PR, I can split it out into a separate one. Either way, this is ready for another review. Thanks!

Issue #787, opened 2026-06-12
Deleting a BFD peer blocks waiting for sync thread to be joined


jgallagher:
I'm trying to wire up proptests to the omicron BFD config reconciler. My first few attempts at that resulted in extremely long tests (many minutes). I think I've narrowed this down to "deleting a BFD peer immediately after adding it blocks for about 1 second", which is easily reproducible:

```
% cat make-bfd-peer.json
{
    "peer": "1.2.3.4",
    "listen": "0.0.0.0",
    "required_rx": 1234,
    "detection_threshold": 10,
    "mode": "SingleHop"
}
% time curl -X PUT -H 'content-type: application/json' -H 'api-version: 8.0.0'
'http://[::1]:8889/bfd/peers' --data @make-bfd-peer.json && \
  time curl -X DELETE -H 'api-version: 8.0.0' 'http://[::1]:8889/bfd/peers/1.2.3.4'
curl -X PUT -H 'content-type: application/json' -H 'api-version: 8.0.0'     0.00s user 0.00s system 79% cpu 0.007 total
curl -X DELETE -H 'api-version: 8.0.0' 'http://[::1]:8889/bfd/peers/1.2.3.4'  0.00s user 0.00s system 0% cpu 1.002 total
```

The mgd logs confirm the handler latency for this delete was about 1 second as well:

```
14:21:57.634Z INFO slog-rs (mgd): request completed
    latency_us = 994447
    local_addr = [::1]:8889
    method = DELETE
    module = admin
    remote_addr = [::1]:36984
    req_id = ace339cc-ef54-413e-9481-cb6085d0f466
    response_code = 204
    unit = api_server
    uri = /bfd/peers/1.2.3.4
```

The [`remove_bfd_peer()` function](https://github.com/oxidecomputer/maghemite/blob/6fd0da15ebfadcaac525deb1246c597a375f9903/mgd/src/bfd_admin.rs#L174-L203) looks quite innocuous: it takes a (sync) lock to remove an entry from a `HashMap` (`bfd.daemon.remove_peer()`), then takes a different (sync) lock to remove an entry from a `HashSet`, set an atomic boolean, and remove some entries from other `HashMap`s (`bfd.dispatcher.remove()`). This seems like it should be fine: all these operations should be fast and there appear to be no blocking calls made.

[...]

PR #795, opened 2026-06-17
Intoduce tokio/async-based BFD implementation

jgallagher:
This PR adds a tokio/async-based implementation of BFD. It's heavily based on the existing implementation both in API shape and names, but has some intentional architectural differences (primarily to support pretty extensive unit testing). The non-test code is handwritten, but I leaned pretty heavily on Claude to help with writing tests (although I went over all of them by hand - it has a tendency to write flaky-in-a-not-obvious-way tests when sockets are involved, so hopefully I caught all of those) and used it to compare this implementation against the existing one. The primary motivator for moving to async is fixing #787, which it does: running the reproducer against this implementation results in the `DELETE`s returning more or less immediately (under 1ms).

The biggest architectural changes are:

1. The state machine is written in a pure "sans I/O" style, and must be driven by an external source that provides both input/output and a clock source. This allows straightforward unit testing without having to deal with timeouts, threads, or sockets.
2. Each BFD session starts 3 tokio tasks (one to drive the state machine, one to send outgoing packets, and one to sync with the RIB), plus a tokio task for the local listening address that's shared with all BFD sessions using that same listening address. This is pretty close to swapping out threads for tasks; the sync implementation started 3 OS threads per session (two to drive the state machine, one to send outgoing packets) plus a local listening thread shared between sessions. It didn't need a separate thread for RIB syncing, but we do that here to bridge between the sync and async worlds.
3. Communication between the UDP send/recv tasks and the state machine driver use bounded channels and drop packets if the channels are full. The old implementation used unbounded channels. I don't expect this to matter in practice, but it's a spot where something going off the rails (e.g., a `send_to()` getting stuck or something) used to mean we'd have unbounded memory growth now means we have a lot of dropped packets instead.

Some less-architectural-but-still-intentional changes:

1. I made the detection threshold a `NonZeroU8`, both because I was concerned about multiplying by 0 and because the RFC says such packets `MUST` be rejected. We now do so, and also return a 400 if we're asked to configure a session with a threshold of 0. This might be the wrong call and I'm happy to discuss - there was a "panic on startup if we can't re-add all peers" path, so I made that path bump any persisted 0s to 1s, but maybe I should do the same in the API handler instead of returning a 400, in case there are Nexus configs persisted that have a threshold of 0?
2. Some more careful math in the event of unreasonable large numbers (e.g., replacing an `as u32` that wrapped around on large values with `.saturating_*()` methods).
3. We now increment `SessionCounters::control_packets_received` on incoming packets - I think this was just an oversight before?
4. If we fail to bind an egress socket, the old code would retry every 5 seconds; this code retries every time it has a packet it wants to send. I'm not sure if this is better or worse but it was simpler to implement - happy to discuss.

Some things that I left unchanged but I'm not sure are correct:

1. Calculations of the "next send" and "recv deadline" timers in the state machine - I left `TODO-correctness` comments on these with my notes but didn't want to change that behavior as part of a big rework.
2. Outgoing packets don't populate the local `detection_multiplier` value (also tagged with a `TODO-correctness` in the state machine).
3. We close the egress socket and rebind on _any_ error from `UdpSocket::send_to()`. Could that leave us in a constant loop of closing/opening sockets if sends are failing for some reason not related to the local socket?
4. When removing the peer, there's no final sync with the RIB (nor is there in the sync implementation) - I think this means if we remove a peer in the `Up` state, we'll leave behind a `nexthop_shudown` of false for that address? I don't know whether this is okay or gets cleaned up by something else.
5. Should `required_min_rx` also be a `NonZero` type? If that gets set to 0 we'll also have a "multiply by 0" in the calculation of the recv deadline.

This PR is only the async implementation with no modifications to the main `mgd` binary except the change of detection threshold to `NonZeroU8` mentioned above. I'll open a second PR that swaps the implementations over.


mgd/src/main.rs
```diff
@@ -471,7 +471,15 @@ fn start_bfd_sessions(
...
-    for config in configs {
+    for mut config in configs {
+        // Backwards-compatibility protection: We now reject a detection
+        // threshold of 0 (as required by RFC 5880), but it's possible we
+        // previously persisted a config with such a threshold before adding the
+        // guards that reject them. Bump it up to 1 to ensure we don't panic in
+        // the `unwrap_or_else()` below.
+        if config.detection_threshold == 0 {
+            config.detection_threshold = 1;
+        }
```
nicolaskagami: Ideally we would enforce the `NonZero` `detection_threshold` in this input type itself so it's unrepresentable (as per "parse, don't validate").

bfd-async/src/sm.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    // TODO-correctness What should we do on "overflows an Instant"? That should
+    // be _very_ impossible given any reasonable values for `last_recv` (which
+    // are always from `Instant::now()` and `recv_timeout`, but someone passing
+    // a truly absurdly large `required_min_rx` could maybe cause problems? For
+    // now, we'll fall back to a hardcoded, large recv deadline if this addition
+    // overflows, but the actual value we pick is a total WAG.
+    last_recv
+        .checked_add(recv_timeout)
+        .unwrap_or_else(|| last_recv + Duration::from_secs(60))
```
nicolaskagami, line 266: I'm for defaulting to a max duration for `recv_timeout`. I'd also like to note that this could still panic for unreasonable values of `last_recv` (near the representational limit).

bfd-async/src/sm.rs
```diff
@@ -0,0 +1,316 @@
...
+            }
+            (State::Up, BfdPeerState::Init | BfdPeerState::Up) => None,
+        };
+        if let Some((our_next_state, result)) = transition {
+            self.state = our_next_state;
+            result
+        } else {
+            PacketReceivedResult::StateUnchanged
+        }
+    }
```
nicolaskagami: I would be in favor of simplifying away `PacketReceiveResult`, as I think it's a bit semantically overloaded and not load-bearing. Currently it does a few things:
- Indicate that a remote state is unknown (`UnknownPeerState`, determined in `packet_received`).
- Indicate that the local state has not changed (`StateUnchanged`), in which case we don't explicit our current state.
- Indicate that a local state transition took place, wherein it has no type enforcement of its relationship with the previous and next states.

The latter two could be solved using a couple of `BfdPeerState` either in a tuple or a struct for the named fields. This helps a lot with maintainability, even though we don't expect to add a new state to the BFD protocol. But I think we can simplify it more than that.

More importantly, I think `fn packet_received` should not have the same return type as the internal `fn update_remote_peer_state` so that we don't conflate their different purposes. This should help with testing and maintainability.

Speaking of tests, I see that the only place the "state transition" role of `PacketReceiveResult` is being used is in a test (`transition_table_is_exhaustive_and_correct`) that is effectively a copy of the `update_remote_peer_state` internal mapping. This works but I think it'd be more readable/maintainable to decompose it into multiple tests, each with the RFC's specific expectations for the specific transition. That way it's much clearer what we're testing. 

Also, since we're testing internals, `update_remote_peer_state` doesn't even need to return anything, since we can check the local state before and after. It could become something like this:
```rust
  fn update_peer_state(&mut self, remote_peer_state: BfdPeerState) {
      let next_state = match (self.state, remote_peer_state) {
          (State::Down, BfdPeerState::Down) => State::Init,
          (State::Down, BfdPeerState::Init) => State::Up,
          (State::Down, BfdPeerState::AdminDown | BfdPeerState::Up) => State::Down,
          
          (State::Init, BfdPeerState::AdminDown) => State::Down,
          (State::Init, BfdPeerState::Down) => State::Init,
          (State::Init, BfdPeerState::Init | BfdPeerState::Up) => State::Up,

          (State::Up, BfdPeerState::AdminDown | BfdPeerState::Down) => State::Down,
          (State::Up, BfdPeerState::Init | BfdPeerState::Up) => State::Up,
      };
      self.state = next_state;
  }
```

I know this sounds like a bit of work but I'm happy to take it on.

bfd-async/src/sm.rs
```diff
@@ -0,0 +1,316 @@
...
+            discriminator: packet.my_discriminator,
+            demand_mode: packet.demand(),
+            detection_multiplier: packet.detect_mult,
+        };
+    }
+
+    fn update_remote_peer_state(
+        &mut self,
+        peer_state: BfdPeerState,
+    ) -> PacketReceivedResult {
```
nicolaskagami: If I understood correctly, we're trying to update our local state depending on the (remote) state we observed from the incoming packet.
```suggestion
    fn update_peer_state(
        &mut self,
        remote_peer_state: BfdPeerState,
    ) -> PacketReceivedResult {
```

bfd-async/src/sm.rs
```diff
@@ -0,0 +1,316 @@
...
+            self.poll_responses.push_back(pkt);
+        }
+
+        match packet.state() {
+            packet::State::Peer(peer_state) => {
+                self.update_remote_peer_state(peer_state)
+            }
+            packet::State::Unknown(_) => PacketReceivedResult::UnknownPeerState,
+        }
+    }
```
nicolaskagami: I noticed that we don't persist the remote state (`bfd.RemoteSessionState`). 

I understand that on normal circumstances it may not matter, but there are a few RFC statements that explicitly check for it ([6.8.6](https://datatracker.ietf.org/doc/html/rfc5880#section-6.8.6), [6.8.7](https://datatracker.ietf.org/doc/html/rfc5880#section-6.8.7)).

bfd-async/src/sm.rs
```diff
@@ -0,0 +1,316 @@
...
+    // * in async mode, calculation should be max(local min rx, remote min tx) *
+    //   remote detection multiplier
+    // * in demand mode calculation should be max(local min tx, remote min rx) *
+    //   local detection multiplier
+    //
+    // I don't think we support demand mode? But that means this calculation is
+    // wrong for async (wrong multiplier, isn't considering remote min tx)?
+    let recv_timeout = local
+        .required_min_rx
+        .saturating_mul(u32::from(local.detection_multiplier.get()));
```
nicolaskagami, line 256: Agreed. According to the RFC, we should use the remote detection mult for async mode, though I admit it feels weird to use the remote one.
```
   In Asynchronous mode, the Detection Time calculated in the local
   system is equal to the value of Detect Mult received from the remote
   system, multiplied by the agreed transmit interval of the remote
   system (the greater of bfd.RequiredMinRxInterval and the last
   received Desired Min TX Interval)
```

bfd-async/src/sm.rs
```diff
@@ -0,0 +1,316 @@
...
+            result
+        } else {
+            PacketReceivedResult::StateUnchanged
+        }
+    }
+
+    fn make_packet_to_send(&self) -> packet::Control {
+        // TODO-correctness Should this set detect_mult to
+        // `self.local.detection_multiplier`? The old state machine didn't, but
+        // this means we always send the default value instead of our config.
```
nicolaskagami: Yes. We should send our detection multiplier.

bfd-async/src/sm.rs
nicolaskagami: An argument could be made here to ignore packets signaling an unknown state. Ideally, that would happen along with a more strongly typed packet input. Your addition of a `NonZeroU8` field is a great step in this direction (as per "parse, don't validate") but I don't think we should tackle it in this PR.

nicolaskagami (commented):
(doing a first pass for now focusing on the state machine)

Overall this looks great! This is exactly the place where the sans-io approach with proptesting is really strong. I see you made many of the same decisions I made when working on a similar project.

I agreed with all your `TODO-correctness` comments. It seems there are a few features missing from our current BFD implementation, so I wonder what our overall plan is for it.

I added a few suggestions which might be a little tiresome but I'd be more than happy to tackle them.

bfd-async/src/sm.rs
jgallagher: > I'd also like to note that this could still panic for unreasonable values of last_recv (near the representational limit).

This is true, but in practice `last_recv` is fully under our control, and is always the result of calling `Instant::now()` (unlike any of the timeout parameters, which ultimately come from outside, either directly from the BFD peer or indirectly by way of Nexus -> mgd in terms of parameters). I think it's okay to leave the `+` in here because of that?

I could also drop this in a comment if you think that'd be an improvement. I'm on the fence.

bfd-async/src/sm.rs
jgallagher: Yeahhh this is good feedback. I did the state machine first and was expecting this enum to be useful in the driver, but the driver ended up being cleaner doing exactly what you noted we could do instead (check the state before and after). I'll rework this.

mgd/src/main.rs
jgallagher: 100% agreed, but this bubbles out to the persisted state, which I have no context for so didn't want to change as part of this PR. I'll file an issue that we should push `NonZeroU8` out to both rdb and the API?

jgallagher: Oof, ok, so I looked very briefly into this and I think we may have a slightly bigger landmine here than I thought. `BfdPeerConfig` is part of the versioned HTTP API, so it's defined in the `mg-api-types-versions`, and we'll follow the standard RFD 619 process for revving that type. If we wanted to change `detection_threshold` to a `NonZeroU8`, we'd:

* add a new API version
* add a new `BfdPeerConfig` with the changed field type
* add either `From` or `TryFrom` impls to convert if we get requests from an older client (depending on whether we want to reject requests that have a multiplier of 0, or silently convert them to 1)

That would all work fine at the API level. However, we also persist JSON-ified `BfdPeerConfig`s on disk, but _with no versioning information_ AFAICT. When we then tried to load them here: https://github.com/oxidecomputer/maghemite/blob/dc5317131cb3cd1f37e24a245c2e389f86052f6f/rdb/src/db.rs#L471-L484

we'd be trying to deserialize a JSON blob from whatever version was originally persisted as the latest `BfdPeerConfig` version. Any of these are possible:

* it just works (e.g., if the change is or happens to be wire-compatible; in this case going from `u8` to `NonZeroU8` is wire compatible as long as there are no `0` values)
* it fails (e.g., if we rename fields or add new required fields) - we'd log an error but then skip that config
* it appears to work but silently does the wrong thing (e.g., if we change the semantics of some field in a way that's handled by the `From` or `TryFrom` conversions at the API level but which don't get a chance to run with this "deserialize old blob as new type" path)

We've been burned by this kind of on-disk lack of versioning thing in omicron a few places and have had to go back and add extra infrastructure around on-disk persisted values. Maybe we need to do that here before revving any API types that are also persisted on disk?

bfd-async/src/sm.rs
jgallagher: Done in b4e7150

mgd/src/main.rs
jgallagher: Oh actually maybe all of this is irrelevant - we do persist on disk, but the switch zone has ephemeral storage, so there's not actually a way for us to see an old version disk. I'm not sure why we persist to disk at all, actually - I guess it's for crash recovery (but only if `mgd` itself crashes - if the sled reboots we don't have any persistence so have to wait for nexus / sled-agent to tell us what to do).

taspelund: The on-disk db (sled::Db) is only used for daemon restart/crash recovery, like you said. mgd effectively trusts that there is no need to handle backwards compatibility issues with sled::Db, since all upgrades happen via a wholesale zone replacement that mops up the old filesystem anyway.

jgallagher: Yeah, makes sense. I think that means I can take out this backcompat shim entirely, right? (We should still push `NonZeroU8` out into `BfdPeerConfig`, but that can happen as a followup.)

taspelund: Yeah, I think so.  And if we're going to start enforcing the value be non zero (which I think is a good thing!) then we may consider just bumping the API rev as part of this.  I'm happy to take that on as a follow-up so you don't have to deal with a bunch of boilerplate and crate org stuff.

bfd/src/packet.rs
```diff
@@ -161,7 +164,7 @@ impl Default for Control {
             // default state machine state is down
             flags: BfdPeerState::Down.wire_format(),
             // default to detection threshold multipler of 3
-            detect_mult: 3,
+            detect_mult: NonZeroU8::new(3).expect("3 is not 0"),
```
taspelund: Seems like this could be a const rather than an inline expect?

taspelund: Or possibly a thin wrapper type with a default inner value of 3 -- I see the magic number reused in all kinds of Default impls in lib.rs

mgd/src/main.rs
jgallagher: I don't mind bumping the API version; I've done that a bunch in omicron. But this is only one bit of the BFD peer config that's a little suspect - I filed https://github.com/oxidecomputer/omicron/issues/10657 on the omicron side, but it's almost a 1-to-1 port of mgd's type. Should we also consider:

1. Should `required_rx` either be a `u32` (to match BFD control packets) or a `Duration` (since that's what it actually is, then we put the onus of converting to microseconds on mgd internally)?
2. Should `required_rx` enforce any upper or lower bounds (via a newtype or runtime checks)?
3. Is allowing the caller the ability to specify any local listening `IpAddr` right?
4. Should `get_bfd_peers()` return an `IdOrdMap<BfdPeerConfig>` (keyed by remote address) instead of a `Vec<BfdPeerConfig>`?

taspelund: These are all great questions... which I don't have the answer to!  I haven't had to dig into the BFD protocol or implementation before, so I think I'll have more opinions as I get further along the review

bfd-async/src/daemon.rs
```diff
@@ -0,0 +1,108 @@
...
+            hash_map::Entry::Occupied(_) => {
+                warn!(
+                    self.log, "attempt to add peer that already exists";
+                    "component" => bfd::COMPONENT_BFD,
+                    "module" => bfd::MOD_DAEMON,
+                    "unit" => bfd::UNIT_PEER,
+                    "peer" => %peer,
+                );
+                Err(AddPeerError::PeerExists(peer))
+            }
```
taspelund, line 73: Is this correct? I get that semantically trying to "add" a peer that's already there would hit a conflict, but what about updating parameters on an existing peer? I'm not sure if that is something we'd rather have exposed via an "update" method or if "add" should be idempotent

jgallagher: I'm not sure about correct in principle, but it does match the current behavior on main: https://github.com/oxidecomputer/maghemite/blob/3b64226bcaf0825a653bfbee983e5ab3b1422d20/mgd/src/bfd_admin.rs#L144-L149

I think this was an intentional choice for urgency, based on https://github.com/oxidecomputer/omicron/issues/4921.

taspelund: Yep, sure looks like it. I can look at a fix for this in a follow-up. Would you mind adding in a TODO/XXX comment in the meantime?

bfd-async/src/egress_src_port_iter.rs
```diff
@@ -0,0 +1,96 @@
...
+use std::sync::atomic::AtomicU16;
+use std::sync::atomic::Ordering;
+
+const OFFSET_RANGE: u16 = (u16::MAX - EgressSrcPortIter::SOURCE_PORT_BEGIN) + 1;
+
+/// Helper for choosing a source port for egress sockets in BFD.
+///
+/// Per RFC 5881 §4, BFD control packets MUST have a source port in the range
+/// `[49152, 65535]`. We assume we'll never have more than 16384 active
+/// sessions, so do not support the RFD's note that source ports "MAY be reused
```
hawkw:
```suggestion
/// sessions, so do not support the RFC's note that source ports "MAY be reused
```

bfd-async/src/rib.rs
```diff
@@ -0,0 +1,202 @@
...
+            // Only actually sync if we've had a change. (E.g., if the state
+            // went from "down" to "init", we don't need to do anything, because
+            // either way it's still shut down.)
+            if Some(shutdown) != prev_syncd_shutdown {
+                let sink = sink.clone();
+                let nexthop = self.nexthop;
+                let result = tokio::task::spawn_blocking(move || {
+                    sink.set_nexthop_shutdown(nexthop, shutdown);
+                })
+                .await;
```
hawkw, line 79: it's a bit of a bummer that we spawn a new blocking task every time we want to write something, rather than makingthe worker that does the writing be a dedicated `std::thread::spawn`ed thread that just does the blocking write itself, and then blocks until it has more work to do. but i see that we're using `watch::Receiver::changed()` to wait for a change, and, of course, there is no blocking version of that, so meh whatever.

bfd-async/src/session.rs
```diff
@@ -0,0 +1,329 @@
...
+
+            maybe_pkt = self.listener_rx.recv() => match maybe_pkt {
+                Some(pkt) => Event::PacketReceived(pkt),
+                None => Event::ShuttingDown,
+            },
+
+            () = tokio::time::sleep_until(deadline.into()) => {
+                Event::DeadlineExpired
+            }
+        }
```
hawkw: take it or leave it: could this instead be implemented as
```suggestion
        // Cancel-safety: both `recv()` and `sleep_until()` are cancel safe
        // (`recv()` per its docs; `sleep_until()` because we want to discard it
        // if we receive a packet). Neither arm contains an `.await`, avoiding
        // any opportunity for futurelock.
        match tokio::time::timeout_at(deadline.into(), self.listener_rx.recv()) {
            Ok(Some(pkt)) => Event::PacketReceived(pkt),
            Ok(None) => Event::ShuttingDown,
            Err(_) => Event::DeadlineExpired,
        }
```

which IMO is a bit better than reaching for the more expressive --- and therefore more potentially fraught/more easily futurelock-able --- `select!`?

mgd/src/main.rs
jgallagher: de5ab22 bumps the API version to make this param a `NonZeroU8`. Happy to add more changes pending answers to the other questions, or that can be deferred for later.

bfd/src/packet.rs
jgallagher: Added a `DEFAULT_DETECT_MULTIPLIER` const in 4924ec6

bfd-async/src/daemon.rs
jgallagher: Added in 8b3ff9a

bfd-async/src/session.rs
jgallagher: Thanks, I agree that's better. It also gets rid of the only (non-test) `select!`, which feels good. Pulled in as efbbe69

bfd-async/src/egress_src_port_iter.rs
```diff
@@ -0,0 +1,96 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::sync::atomic::AtomicU16;
+use std::sync::atomic::Ordering;
+
+const OFFSET_RANGE: u16 = (u16::MAX - EgressSrcPortIter::SOURCE_PORT_BEGIN) + 1;
```
nicolaskagami: microNIT: I'd keep this and `SOURCE_PORT_BEGIN` beside each other, wherever that is.

bfd-async/src/egress.rs
```diff
@@ -0,0 +1,246 @@
...
+        }
+    }
+
+    Err(BindEgressSocketError::SrcPortsInUse {
+        // We know these must be `Some(_)` becaues `max_ports_to_try` is a
+        // `NonZeroUsize`, so we must have iterated at least once in the for
+        // loop above, and the first iteration sets both these values.
+        first_port_tried: first_port_tried.unwrap(),
+        last_port_tried: last_port_tried.unwrap(),
+    })
```
nicolaskagami:
```suggestion
    let first_port_tried = src_port_iter.next();
    let mut last_port_tried = first_port_tried;

    for _ in 0..max_ports_to_try.get() {
        match bind_egress_socket(SocketAddr::new(local_ip, last_port_tried))
            .await
        {
            Ok(socket) => return Ok(socket),
            Err(BindEgressSocketError::Bind(err))
                if err.kind() == io::ErrorKind::AddrInUse => {}
            Err(err) => return Err(err),
        }
        last_port_tried = src_port_iter.next();
    }

    Err(BindEgressSocketError::SrcPortsInUse {
        first_port_tried,
        last_port_tried,
    })
```
We can remove the `unwrap`s.

bfd-async/src/dispatcher.rs
```diff
@@ -0,0 +1,426 @@
...
+        // This is pretty spicy: we're using std's `UdpSocket` so we can
+        // _synchronously_ bind, even though this function is ultimately called
+        // from an async context. The arguments for this are:
+        //
+        // 1. Binding a UDP socket doesn't involve any network traffic, so
+        //    shouldn't block for long
+        // 2. Reworking this to allow async binding is pretty involved; we'd
+        //    need to either use an async Mutex (lots of footguns there) or move
+        //    to an actor task pattern to manage the listeners.
+        let socket = std::net::UdpSocket::bind(listen_addr).map_err(|err| {
```
nicolaskagami, line 299: This is ok. The only reason it could block for long is if you passed it a `listen_addr` that required name resolution, which the method accepts (`ToSocketAddrs`).

bfd-async/src/dispatcher.rs
```diff
@@ -0,0 +1,426 @@
...
+    /// Returns the spawned task's handle, or `None` for test fakes that don't
+    /// bind a real socket. A `Listener` holding `None` is shut down/dropped as
+    /// a no-op.
+    fn spawn(
+        &self,
+        listen_addr: SocketAddr,
+        sessions: SharedSessions,
+        log: Logger,
+    ) -> Result<Option<JoinHandle<()>>, AddPeerError>;
+}
```
nicolaskagami, line 277: I'd rather not change the signature just for tests. You could remove the `Option` from here and from `Listener` and just return `Ok(tokio::spawn(async {}))` for the `FakeBackend`.

(tip: for `Listener::shutdown` you can `let _: Result<_, _> = (&mut self.listen_task).await;`)

bfd-async/src/sm.rs
nicolaskagami: I think it'd be slightly better to deal with it separately: 
- Make sure `recv_timeout` is sensible when it is set ("by construction").
- If the checked add fails you return the maximum representable or something like that. In that case it should only fail if the value was tempered with or we're at the heat-death of the universe so we can return whatever. Not doing this last unchecked add at least doesn't trick us into checking (with our brains) every time we look at it.

bfd-async/src/egress.rs
```diff
@@ -0,0 +1,246 @@
...
+                    self.log, "udp egress send_to failed; will rebind";
+                    "remote" => %self.remote_addr,
+                    InlineErrorChain::new(&err),
+                );
+                // Drop the socket so the next packet triggers a fresh bind.
+                //
+                // TODO-correctness Should we _always_ drop and rebind on a send
+                // failure, or should we be matching on particular kinds of
+                // errors? Always rebinding is consistent with the prior (sync)
+                // implementation.
```
nicolaskagami, line 162: Looking at how `send_to` can fail it seems it's mostly when trying to convert the target address and very little to do with our socket. So maybe you don't even need to drop it, but I don't think it matters too much.

nicolaskagami (approved):
It's looking pretty good. 

I'd love other people's opinion on the semantic concerns, particularly on the correctness of the detection multiplier.
Additionally:
- Not syncing the state when removing a peer sounds concerning to me, but I guess it depends on how we're using it.
- Are we implementing all features we want from BFD?

I also have a few observations below that are entirely syntactical.

bfd-async/src/sm.rs
jgallagher: Hmm, I'm not sure how to do _either_ of these with 100% confidence:

1. `recv_timeout` is at least partly under the control of the remote end, at least if we fix the calculation to match what the RFD says it should be
2. `Instant` doesn't provide `saturating_add()` or `Instant::MAX`, so there's no way (AFAIK) to construct the maximum representable value.

nicolaskagami: 1. We could just define a newtype. IMO the ideal newtype would represent a sensible configuration (which includes this timeout) so that we have a clear place to deal with these misconfiguration errors (where we convert the input type into this newtype). In this case, for example, there's a local strategy for dealing with an unreasonable value but it may still behave erratically elsewhere.
2. If an `Instant` + "something reasonable" (as per 1.) overflows then return that `Instant`.

bfd-async/src/egress_src_port_iter.rs
jgallagher: fixed in 70d092ed

bfd-async/src/egress.rs
jgallagher: I think as-written that has a couple related off-by-one issues: we'll run `last_port_tried = src_port_iter.next()` as the last statement in the final loop iteration, which means

* the error we return has one higher `last_port_tried` than we actually tried
* we'll cause a port to be skipped in the shared `src_port_iter` iterator

I reworked it some in 16f0bf32 to address this while still removing the unwrap - look okay?

bfd-async/src/dispatcher.rs
jgallagher: works for me! done in 0598e2f3

bfd-async/src/egress.rs
nicolaskagami: Yep.

jgallagher:
I think I've addressed most of the directly-addressable comments; the remaining ones I'm not 100% sure how to proceed:

1. How much validation to push into types
    a. The mgd equivalents of [omicron#10657](https://github.com/oxidecomputer/omicron/issues/10657) for the API (https://github.com/oxidecomputer/maghemite/pull/795#discussion_r3469345586)
    b. How much validation to put on the incoming packets from the remote peer; e.g., sane timeouts (https://github.com/oxidecomputer/maghemite/pull/795#discussion_r3573215084)
2. Adherence to the BFD RFC
    a. Remembering / acting on the remote peer state (https://github.com/oxidecomputer/maghemite/pull/795#discussion_r3455031839)
    b. Recv timeout calculation (https://github.com/oxidecomputer/maghemite/pull/795#discussion_r3455237097)
    c. What to do with unknown state packets (https://github.com/oxidecomputer/maghemite/pull/795#discussion_r3455314886)
    d. Not syncing any state with the RIB when removing a peer (https://github.com/oxidecomputer/maghemite/pull/795#pullrequestreview-4626222029)

I don't think this PR is making any of this worse, so I'm somewhat tempted to propose merging this as-is and opening issues for the above, to be tackled in the future (time TBD, person TBD). But if folks feel strongly that we should address some of these as a part of this PR, that's doable.

bfd-async/src/dispatcher/end_to_end.rs
```diff
@@ -0,0 +1,167 @@
...
+        only_state(&daemon_a) == BfdPeerState::Up
+            && only_state(&daemon_b) == BfdPeerState::Up
+    })
+    .await
+    .expect("both sessions reach Up");
+
+    // Drop B's peer: this aborts B's session (so it stops sending) and shuts
+    // down B's listener socket.
+    let handle = daemon_b
+        .remove_peer(LOCALHOST)
```
taspelund: Should this be `addr_a.ip()` instead of `LOCALHOST`?

bfd-async/src/dispatcher/proptests.rs
```diff
@@ -0,0 +1,231 @@
...
+        // A peer may only be registered once, regardless of address. This is
+        // checked before any bind is attempted.
+        if self.peer_to_addr.contains_key(&peer) {
+            return EnsureOutcome::PeerExists;
+        }
+        // A bind only happens when this address has no listener yet.
+        let listener_exists = self.peer_to_addr.values().any(|a| *a == addr);
+        if !listener_exists && fail_addrs.contains(&addr) {
+            return EnsureOutcome::BindFailed;
+        }
```
taspelund: What is the intent behind `fail_addrs` if it's only checked when a listener doesn't already exist? Why combine the two conditions into one check?

bfd-async/src/dispatcher/tests.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    // The channel buffers at most its capacity (8); the rest is dropped. (UDP
+    // itself may also drop, so we only assert the upper bound and liveness.)
+    let mut count = 0;
+    while rx.try_recv().is_ok() {
+        count += 1;
+    }
+    assert!(count >= 8, "expected a full channel worth of packets");
+
+    // The task is still alive: with room in the channel, a fresh packet flows.
```
taspelund, line 245: The ~spice~ packets must flow

bfd-async/src/dispatcher/tests.rs
```diff
@@ -0,0 +1,289 @@
...
+        .expect("removing last peer yields a handle");
+
+    // Awaiting shutdown guarantees the task was aborted and dropped.
+    handle.shutdown().await;
+
+    assert!(backend.was_listener_task_dropped());
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn rx_closed_after_last_peer_removed() {
```
taspelund, line 271: Is this worth splitting into its own test? It seems like 80% of this test is boilerplate and the single assert could be consolidated into `shutdown_waits_for_task_to_be_dropped`

mg-api-types/versions/src/bfd_nonzero_detect_mult/mod.rs
```diff
@@ -0,0 +1,10 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Version `BFD_NONZER_DETECT_MULT` of the Maghemite Admin API.
```
taspelund: typo: missing "O" at the end of 'NONZERO`

bfd-async/src/dispatcher.rs
```diff
@@ -0,0 +1,420 @@
...
+            return None;
+        };
+
+        match listener.remove_peer(peer) {
+            ListenerRemovePeerResult::RemovedNowEmpty => {
+                // we can safely `unwrap()` this: we _have_ a reference to the
+                // `listener` now, which we got from the map by this same key.
+                // we know it's still there.
+                let listener = self.listeners.remove(&local_addr).unwrap();
+                Some(ListenerShutdownHandle(listener))
```
taspelund, line 168: Could we cleanly avoid the unwrap by just replacing this with `self.listeners.remove(&local_addr).map(|listener| ListenerShutdownHandle(listener))`?

bfd-async/src/dispatcher.rs
```diff
@@ -0,0 +1,420 @@
...
+                        "local" => ?self.socket.local_addr(),
+                        InlineErrorChain::new(&err),
+                    );
+                    continue;
+                }
+            };
+
+            // Do we expect traffic from this peer address?
+            let Some(session) =
+                self.sessions.read().unwrap().get(&peer.ip()).cloned()
```
taspelund: Can we use `mg_common::read_lock` instead of `.read().unwrap()`?

bfd-async/src/dispatcher.rs
```diff
@@ -0,0 +1,420 @@
...
+            ListenerRemovePeerResult::NotRemoved
+        }
+    }
+
+    fn insert_peer(
+        &self,
+        peer: IpAddr,
+        session: SessionHandle,
+    ) -> Result<(), AddPeerError> {
+        let mut sessions = self.sessions.write().unwrap();
```
taspelund: Can we use `mg_common::write_lock` instead of `.write().unwrap()`?

bfd-async/src/dispatcher.rs
```diff
@@ -0,0 +1,420 @@
...
+
+    async fn shutdown(mut self) {
+        self.listen_task.abort();
+        // Discard the result; this can only fail if the listen task
+        // panicked, but we're shutting it down anyway.
+        let _: Result<_, _> = (&mut self.listen_task).await;
+    }
+
+    fn remove_peer(&self, peer: IpAddr) -> ListenerRemovePeerResult {
+        let mut sessions = self.sessions.write().unwrap();
```
taspelund: `write_lock!()` here as well

bfd-async/src/egress/tests.rs
```diff
@@ -0,0 +1,154 @@
...
+    assert!(
+        srcport2 >= EgressSrcPortIter::SOURCE_PORT_BEGIN,
+        "unexpected src port {srcport2}"
+    );
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn exits_when_channel_closed() {
+    let (tx, _counters, handle) =
+        spawn_egress("127.0.0.1:1".parse().unwrap(), Arc::default());
```
taspelund: We have `mg_common::ip` that does this parse/unwrap in a short macro. Can we use that?

taspelund: Sorry, I meant `sockaddr` since this has a port

bfd-async/src/session/tests.rs
```diff
@@ -0,0 +1,276 @@
...
+        Instant::now(),
+    );
+    let (state_tx, state_rx) = watch::channel(sm.state());
+    let counters = Arc::new(SessionCounters::default());
+
+    let driver = tokio::spawn(
+        DriverTask {
+            sm,
+            counters: Arc::clone(&counters),
+            remote_addr: "127.0.0.1:3784".parse().unwrap(),
```
taspelund: Can we use `sockaddr!()` here too?

bfd-async/src/rib.rs
```diff
@@ -0,0 +1,202 @@
...
+    const NEXTHOP_IP: IpAddr = IpAddr::V4(Ipv4Addr::new(10, 0, 0, 1));
+
+    /// A `NexthopSink` that records every write, so tests can assert on the
+    /// exact sequence of `(nexthop, shutdown)` updates the task performs.
+    #[derive(Clone, Default)]
+    struct RecordingSink(Arc<Mutex<Vec<(IpAddr, bool)>>>);
+
+    impl NexthopSink for RecordingSink {
+        fn set_nexthop_shutdown(&self, nexthop: IpAddr, shutdown: bool) {
+            self.0.lock().unwrap().push((nexthop, shutdown));
```
taspelund: We have `mg_common::lock` that cleans up `.lock().unwrap()` -- can we use that?

bfd-async/src/rib.rs
```diff
@@ -0,0 +1,202 @@
...
+        }
+    }
+
+    impl RecordingSink {
+        fn writes(&self) -> Vec<(IpAddr, bool)> {
+            self.0.lock().unwrap().clone()
```
taspelund: `mg_common::lock!()` here too please

bfd-async/src/dispatcher/end_to_end.rs
```diff
@@ -0,0 +1,167 @@
...
+    fn spawn(
+        &self,
+        _listen_addr: SocketAddr,
+        sessions: SharedSessions,
+        log: Logger,
+    ) -> Result<JoinHandle<()>, AddPeerError> {
+        let socket = self
+            .socket
+            .lock()
+            .unwrap()
```
taspelund: `mg_common::lock!()` please

bfd-async/src/dispatcher/proptests.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// Small fixed domains so collisions (shared listeners) and reuse actually occur
+// within a short op sequence. The addresses are never bound, so the specific
+// IPs/ports are arbitrary.
+fn arb_listen_addr() -> impl Strategy<Value = SocketAddr> {
+    let addrs = vec![
+        "127.0.0.1:3784".parse().unwrap(),
+        "127.0.0.1:4784".parse().unwrap(),
+        "192.0.2.1:3784".parse().unwrap(),
```
taspelund: `mg_common::sockaddr!()`

bfd-async/src/dispatcher/proptests.rs
```diff
@@ -0,0 +1,231 @@
...
+    // There is exactly one listener per distinct in-use listen address.
+    let actual_listeners: BTreeSet<SocketAddr> =
+        dispatcher.listeners.keys().copied().collect();
+    assert_eq!(actual_listeners, model.listen_addrs());
+
+    // Each listener's session map holds exactly the peers the model assigns to
+    // that address, and no listener is empty.
+    for (addr, listener) in &dispatcher.listeners {
+        let actual: BTreeSet<IpAddr> =
+            listener.sessions.read().unwrap().keys().copied().collect();
```
taspelund: `mg_common::read!()`

bfd-async/src/dispatcher/tests.rs
```diff
@@ -0,0 +1,289 @@
...
+
+impl TestBackend {
+    fn as_backend(self: &Arc<Self>) -> Arc<dyn ListenerBackend> {
+        Arc::clone(self) as Arc<dyn ListenerBackend>
+    }
+
+    fn expect_bound_address(&self) -> SocketAddr {
+        self.bound_address
+            .lock()
+            .unwrap()
```
taspelund: `mg_common::lock!()`

bfd-async/src/dispatcher/tests.rs
```diff
@@ -0,0 +1,289 @@
...
+impl ListenerBackend for TestBackend {
+    fn spawn(
+        &self,
+        listen_addr: SocketAddr,
+        sessions: super::SharedSessions,
+        log: Logger,
+    ) -> Result<tokio::task::JoinHandle<()>, crate::AddPeerError> {
+        assert_eq!(listen_addr, LISTEN_ADDR);
+
+        let mut bound_address = self.bound_address.lock().unwrap();
```
taspelund: `mg_common::lock!()`

bfd-async/src/dispatcher/tests.rs
```diff
@@ -0,0 +1,289 @@
...
+
+#[tokio::test(flavor = "multi_thread")]
+async fn drops_packet_from_unregistered_peer() {
+    let log = test_logger();
+    let backend = Arc::new(TestBackend::default());
+    let mut dispatcher = Dispatcher::with_backend(backend.as_backend());
+
+    // Register a peer that is *not* the loopback source address our client
+    // sends from, so the listener sees a mismatched source IP.
+    let peer: IpAddr = "192.0.2.1".parse().unwrap();
```
taspelund: `mg_common::ip!()`

bfd-async/src/egress/tests.rs
```diff
@@ -0,0 +1,154 @@
...
+use slog::Discard;
+use slog::o;
+use std::net::Ipv4Addr;
+use std::time::Duration;
+use tokio::net::UdpSocket;
+use tokio::time::timeout;
+
+#[tokio::test]
+async fn egress_socket_sets_ttl_v4() {
+    let sk = bind_egress_socket("127.0.0.1:0".parse().unwrap())
```
taspelund: `mg_common::sockaddr!()`

bfd-async/src/egress/tests.rs
```diff
@@ -0,0 +1,154 @@
...
+        .await
+        .expect("created v4 egress socket");
+    let sock = Socket::from(sk.into_std().expect("converted socket"));
+    assert_eq!(sock.ttl_v4().unwrap(), DEFAULT_BFD_TTL);
+}
+
+#[tokio::test]
+async fn egress_socket_sets_hop_limit_v6() {
+    let sk = bind_egress_socket("[::1]:0".parse().unwrap())
```
taspelund: `mg_common::sockaddr!()`

bfd-async/src/dispatcher.rs
```diff
@@ -0,0 +1,420 @@
...
+                }
+            };
+
+            // Do we expect traffic from this peer address?
+            let Some(session) =
+                self.sessions.read().unwrap().get(&peer.ip()).cloned()
+            else {
+                warn!(
+                    self.log, "unknown peer; dropping";
+                    "local" => ?self.socket.local_addr(),
```
taspelund, line 355: What does a prefixed `?` do?

bfd-async/src/dispatcher.rs
```diff
@@ -0,0 +1,420 @@
...
+                    "peer" => %peer,
```
taspelund, line 356: or a prefixed `%`?  I assume these are slog specific?

bfd-async/src/egress/tests.rs
```diff
@@ -0,0 +1,154 @@
...
+
+use super::*;
+use slog::Discard;
+use slog::o;
+use std::net::Ipv4Addr;
+use std::time::Duration;
+use tokio::net::UdpSocket;
+use tokio::time::timeout;
+
+#[tokio::test]
```
taspelund: small style nit: any reason we have tests sitting above test helpers in the file? could we move these below the helpers?

bfd-async/src/egress/tests.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn binds_ports_in_expected_range() {
+    // Bind two peers, then spawn two egress tasks. Send a message to each so
+    // our peer can detect the port on which the egress tasks were bound; both
+    // should be in the range [49152, ..]. We generally expect to get exactly
+    // 49152 and 49153, but we can't assert that: if one or both of those ports
+    // are already in use on whatever system this test is running, we'll search
+    // beyond that.
```
taspelund, line 68: This is the second time I've seen mention of binding a socket and attempting to record the ephemeral ports.  Would this be a good use case for rain's [port_file](https://crates.io/crates/port-file) crate?

bfd-async/src/dispatcher/end_to_end.rs
jgallagher: Yeah. In practice `addr_a.ip()` _is_ `LOCALHOST`, but that's not obvious here.

bfd-async/src/dispatcher/proptests.rs
jgallagher: Good question; I don't remember! If `fail_addrs` stays the same for an entire test execution (which I think it does), there's no reason to also check for an existing listener. I'll take that bit out.

bfd-async/src/dispatcher/tests.rs
jgallagher: Ehh, "80% of the test is boilerplate" makes it sound like this test is a bunch of boilerplate, but it's like 8 lines out of 10 :sweat_smile:. My personal preference is "more smaller tests" rather than stacking assertions up, as long as the "more smaller" don't involve a bunch of test time overhead. But I don't feel strongly; I can combine if you want me to.

bfd-async/src/dispatcher.rs
jgallagher: We could avoid the unwrap like that, but I think it muddies the meaning some; `self.listeners.remove(&local_addr).map(|listener| ListenerShutdownHandle(listener))` looks like "`local_addr` might not be in `self.listeners`, and that's fine - we'll return `None` if it's not there", which is not the situation we're in here.

bfd-async/src/dispatcher.rs
jgallagher: Sure. I'm not a huge fan of macros like that (or the `parse().unwrap()` ones), but if that's the style of the repo I'll switch these over.

bfd-async/src/dispatcher.rs
jgallagher: Yeah, the `?` prefix is to use `Debug` formatting, and a `%` prefix is to use `Display` formatting. In this case `self.socket.local_addr()` is a `std::io::Result<SocketAddr>`, so doesn't impl `Display`, hence the difference.

https://docs.rs/slog/latest/slog/macro.log.html#fmtdisplay-and-fmtdebug-values

bfd-async/src/egress/tests.rs
jgallagher: Reason was these tests don't use the helpers, so it was keeping the helpers closer to the tests that use them. But I don't feel strongly about this; happy to move them.

bfd-async/src/egress/tests.rs
jgallagher: Hmm yeah, still not a fan of this kind of macro, and I think `sockaddr` / `ip` aren't quite correct hygiene-wise; if I import them, I get this error:

```
error: cannot find macro `parse` in this scope
  --> bfd-async/src/egress/tests.rs:25:33
   |
25 |     let sk = bind_egress_socket(sockaddr!("[::1]:0"))
   |                                 ^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `sockaddr` (in Nightly builds, run with -Z macro-backtrace for more info)
```

and have to also import `mg_common::parse` for them to work. I think they need to use `$crate::` when calling another macro? https://doc.rust-lang.org/reference/macros-by-example.html#r-macro.decl.hygiene.crate

bfd-async/src/dispatcher/tests.rs
taspelund: haha true enough. I don't feel super strongly about it one way or the other, so I'm okay if you'd rather leave it as is :)

bfd-async/src/egress/tests.rs
taspelund: I had no idea that was even a thing... sounds like I've got a little cleanup to do for the macros so we can stop importing `mg_common::parse`

bfd-async/src/egress/tests.rs
taspelund: I thought clippy has barked at me for doing this before... but maybe I'm confusing that with having non-test code below tests? idk, either way it's a nice to have. thank you for making the change!

bfd-async/src/dispatcher/end_to_end.rs
jgallagher: fixed in 8b282cff

bfd-async/src/dispatcher/proptests.rs
jgallagher: fixed in 8b282cff

mg-api-types/versions/src/bfd_nonzero_detect_mult/mod.rs
jgallagher: fixed in 8b282cff

bfd-async/src/dispatcher.rs
jgallagher: grep and I believe we replaced all the locks and `parse().unwrap()`s in 52ee4546

bfd-async/src/egress/tests.rs
jgallagher: Moved in 8b8b611f

bfd-async/src/rib.rs
```diff
@@ -0,0 +1,203 @@
...
+    async fn run_impl<S: NexthopSink>(mut self, sink: S) {
+        let mut prev_syncd_shutdown = None;
+
+        loop {
+            // Is the nexthop shut down? We only set this to false when the
+            // session is `Up`. Once a session becomes `Up`, the only valid
+            // transition is to `Down`; we can freely move between `Down`,
+            // `AdminDown`, and `Init` without modifying the RIB, and we only go
+            // back to "not shut down" if we transition back to `Up`.
+            let shutdown = match *self.state_rx.borrow_and_update() {
```
taspelund, line 62: So the way this watch channel works is conceptually similar to a 1-element FIFO where the sender overwrites the element each time they try to write? The reader gets notified when there's been a change to the element, so they check the contents and do some work?

bfd-async/src/session/tests.rs
```diff
@@ -0,0 +1,278 @@
...
+    state_rx: watch::Receiver<BfdPeerState>,
+    counters: Arc<SessionCounters>,
+}
+
+// Spawn a driver task and wrap the extra bits we need in a test harness.
+//
+// Most tests pass a very large `required_rx` to avoid any flakiness introduced
+// by spurious recv timeouts. Only tests exercising recv timeouts specifically
+// pass a "reasoanble" value here.
+fn spawn_driver(required_rx: Duration) -> Harness {
```
taspelund: nit: why not just make this a constructor for `Harness`?

mg-api-types/versions/src/bfd_nonzero_detect_mult/bfd.rs
```diff
@@ -0,0 +1,68 @@
...
+use std::num::NonZeroU8;
+
+#[derive(Debug, Copy, Clone, Deserialize, Serialize, JsonSchema)]
+pub struct BfdPeerConfig {
+    /// Address of the peer to add.
+    pub peer: IpAddr,
+    /// Address to listen on for control messages from the peer.
+    pub listen: IpAddr,
+    /// Acceptable time between control messages in microseconds.
+    pub required_rx: u64,
```
taspelund, line 19: If we're already bumping the API version, any reason we shouldn't convert this to `NonZeroU32`?  Seems like the wire representation for control packets encodes the microseconds as a u32, so I don't see why we should allow users to configure anything greater than `u32::MAX`.

bfd-async/src/sm/tests.rs
```diff
@@ -0,0 +1,533 @@
...
+
+    let mut pkt = packet_with(BfdPeerState::Down);
+    pkt.set_poll();
+    sm.packet_received(pkt, now);
+
+    // A final-flagged response must be queued and available immediately.
+    let out = sm.packet_to_send(now).expect("a response is queued");
+    assert!(out.r#final(), "poll response must set the final flag");
+    assert!(!out.poll(), "poll response must not itself set poll");
+}
```
taspelund, line 85: Should this test pass an `Instant` to `packet_to_send()` that's >= `next_send_at`? That way we're testing that a poll response is returned even when the tx deadline has passed

taspelund (commented):
Somehow only some of my comments ended up attached to a review, but overall I'd say I'm happy with the changes.

Just going through the codebase, I see a handful of issues that concern me with the implementation, but I don't suspect they were introduced by this PR.

I do plan to run the PR through some adversarial reviews from claude just to see what else I may have overlooked.

bfd-async/src/egress/tests.rs
jgallagher: I don't think so? IIUC `port_file` is for when we're spawning a child process that is going to bind to port 0, and we need a wait for it to communicate back to us what port it actually bound. In this case we're binding the sockets in-process, so no need for an out-of-bound communication method.

bfd-async/src/rib.rs
jgallagher: I think that's right, although I don't really think of it that way. (In particular, there's no "out" portion of "FIFO" - if you call `borrow_and_update()` twice with no changes, you'll see the same value both times.) It's (literally) a `RwLock` with some special sauce akin to an async condvar for notifying readers, so I try to think of it like that:

* all the `send_*` methods acquire the write lock then notify all readers that a change has occurred (except for `send_if_modified(f)`, which _conditionally_ notifies the readers depending on what `f` returns)
* `reader.borrow_and_update()` acquires the read lock and marks the current value as seen
* `reader.borrow()` acquires the read lock and does NOT mark the current value as seen (this is rarely what you want)
* `reader.changed()` returns a future that becomes ready whenever the value has changed since the last time _this reader_ called `borrow_and_update()` (which means it's ready immediately if the value has already changed or has never been read)

mg-api-types/versions/src/bfd_nonzero_detect_mult/bfd.rs
jgallagher: I think I'd put this in my "how many API changes do we want to make as part of _this PR_" question in https://github.com/oxidecomputer/maghemite/pull/795#issuecomment-4962102442. E.g., should we make this a `u32`, or should we go all the way to making it a `Duration`? What about the other things described in https://github.com/oxidecomputer/omicron/issues/10657, most of which also apply to the MGD types?

My preference is to not try to fix all the API issues in the same PR as introducing this different impl. But I blurred that line by already doing the `NonZeroU8` one :(

bfd-async/src/sm/tests.rs
jgallagher: Hmm, I don't think I follow. This test is confirming that if we receive a packet with poll set, we immediately have a response to send. It sounds like you're suggesting a different test - if we receive a poll packet but `packet_to_send` isn't called for a long time, we still have a packet to send?

bfd-async/src/session/tests.rs
jgallagher: sure: 59cf3a97

bfd-async/src/sm/tests.rs
taspelund: I saw in another test that you have a `let later = now + ...`. I was thinking maybe we pass `later` to `packet_to_send` so we know the poll response is returned even if we are past the deadline to send a regular (non-final) packet

mg-api-types/versions/src/bfd_nonzero_detect_mult/bfd.rs
taspelund: That's fair. This was just one that I thought was small in scope and could be included in the API bump we're already doing, I don't have any other API shape changes that I think should include in this PR.

One thing to note: moving to NonZero would supposedly address an issue reported by AI as part of its adversarial review. It claims that a user configuring required_rx as 0 will result in a busy loop... Which I know also wasn't introduced by this PR, but I opportunistically hoped to reduce adding more API revs

bfd-async/src/sm/tests.rs
jgallagher: Ah got it. I think that's a separate test - does [d3c8f26](https://github.com/oxidecomputer/maghemite/pull/795/commits/d3c8f26d9bf94e252f4f2ff212222be337b14c24) capture what you're describing?

mg-api-types/versions/src/bfd_nonzero_detect_mult/bfd.rs
jgallagher: Hmm I'm not sure making the required rx a NonZero type is valid. From RFC 5880:

> Required Min RX Interval
>
> This is the minimum interval, in microseconds, between received
> BFD Control packets that this system is capable of supporting,
> less any jitter applied by the sender (see [section 6.8.2](https://datatracker.ietf.org/doc/html/rfc5880#section-6.8.2)).  If
> this value is zero, the transmitting system does not want the
> remote system to send any periodic BFD Control packets.

Us busy looping on zero is certainly wrong, but it sounds like the fix for that is more complex, and not to forbid zero entirely.

bfd-async/src/sm/tests.rs
taspelund: Yep, that looks great. Thanks!

mg-api-types/versions/src/bfd_nonzero_detect_mult/bfd.rs
taspelund: Fair enough. I obviously overlooked that when I was going over RFC 5880, and it clearly says zero is valid. I think it's fine to leave as-is then, and we can follow-up with a fix later.

taspelund (approved):
Sorry, forgot to flip this to approve. The remaining issues I've seen in this code were not introduced by this rewrite

PR #797, opened 2026-06-17
Fix router solicitation parsing issue

rcgoodfellow:
The following code attempts to parse an ICMPv6 message first as an ICMPv6 router advertisement and then as an ICMPv6 router solicitation.

https://github.com/oxidecomputer/maghemite/blob/2f90f709dd9916ad145c919871a9efc1c6e17887/ndp/src/manager.rs#L292-L307

Router solicitations can be 8 bytes when they don't have a link-layer address in the payload. This causes a buffer overrun in ISPF versions prior to oxidecomputer/ispf#4.

This PR

1. Adds test to reproduce this issue. See run failure [here](https://github.com/oxidecomputer/maghemite/pull/797/checks?check_run_id=82015467364)
2. Squashes the bug by updating ISPF.

### DIFF
--- a/ndp/src/manager.rs
+++ b/ndp/src/manager.rs
@@ -437,7 +437,7 @@ mod test {
 
     #[test]
     fn router_solicitation_with_link_layer_addr() {
-        // Attempt to parse a anICMPv6 router solicitation with a link-layer
+        // Attempt to parse an ICMPv6 router solicitation with a link-layer
         // address as a router advertisement. This should produce an error. It
         // should not successfully parse, and it should not panic.
         //
--- a/ndp/src/manager.rs
+++ b/ndp/src/manager.rs
@@ -430,3 +430,60 @@ pub struct InterfaceAdvertisementInfo {
     /// Thread state for rx/tx loops (None if interface not active)
     pub thread_state: Option<InterfaceThreadState>,
 }
+
+#[cfg(test)]
+mod test {
+    use crate::packet::{Icmp6RouterAdvertisement, Icmp6RouterSolicitation};
+
+    #[test]
+    fn router_solicitation_with_link_layer_addr() {
+        // Attempt to parse a anICMPv6 router solicitation with a link-layer
+        // address as a router advertisement. This should produce an error. It
+        // should not successfully parse, and it should not panic.
+        //
+        //     0                   1                   2                   3
+        //  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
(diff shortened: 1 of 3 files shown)


ndp/src/manager.rs
```diff
@@ -430,3 +430,60 @@ pub struct InterfaceAdvertisementInfo {
...
+        let data: [u8; 16] = [
+            133, 0, 11, 22, // type, code checksum
+            0, 0, 0, 0, // reserved
+            1, 1, 0xab, 0xbb, // Link layer addr
+            0xcc, 0xdd, 0xee, 0xff,
+        ];
+
+        if Icmp6RouterAdvertisement::from_wire(&data).is_ok() {
+            panic!("expected error")
+        }
```
taspelund: I would tighten this up to `Icmp6RouterAdvertisement::from_wire(&data).expect_err("RA parsing should fail for RS");`

ndp/src/manager.rs
```diff
@@ -430,3 +430,60 @@ pub struct InterfaceAdvertisementInfo {
...
+        // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+        // |     Type      |     Code      |          Checksum             |
+        // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+        // |                            Reserved                           |
+        // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+
+        let data: [u8; 8] = [133, 0, 11, 22, 0, 0, 0, 0];
+        if Icmp6RouterAdvertisement::from_wire(&data).is_ok() {
+            panic!("expected error")
+        }
```
taspelund: Similar feedback here. I'd probably tighten this to something like `Icmp6RouterAdvertisement::from_wire(&data).expect_err("RA parsing should fail for RS");`

taspelund (commented):
Small nit on the test assertion. Otherwise LGTM.
Should we have test cases for the reverse? i.e. making sure an RA can't be parsed as an RS

ndp/src/manager.rs
rcgoodfellow: done

ndp/src/manager.rs
rcgoodfellow: done

rcgoodfellow:
> Should we have test cases for the reverse? i.e. making sure an RA can't be parsed as an RS

done

ndp/src/manager.rs
taspelund: I meant using `.expect_err()` to consolidate that into a single line mirroring the `.expect()` call for the success case

taspelund (approved):

PR #800, opened 2026-06-24
bgp unnumbered: consolidate NDP-backed unnumbered state

### LINKED ISSUE #801: BGP unnumbered doesn't handle active interfaces state changes
The BGP unnumbered codebase has some interface lifecycle handling, which was introduced primarily to handle the scenario where nexus sends mgd config for an unnumbered interface which hasn't yet been created in the underlying OS.

However, this implementation has gaps:
1) If the OS interface were to be deleted, the active interface doesn't go back into the pending state
2) If the OS interface were to have its ifindex / scope_id change (e.g. upon OS interface deletion/re-creation), this is not detected and the cache is not updated (meaning BGP may keep trying to bind its TCP sockets to an old/wrong/invalid/missing ifindex).

These issues haven't been observed/reported yet, but they are real gaps in the codebase which should be handled.

### LINKED ISSUE #802: ICMPv6 Router Solicitations are generated with invalid checksums
Bad codepath in ndp/src/util.rs:
```
// dst == None here
cksum(src, dst, ICMP6_RS_ULP_LEN, &mut out);

// but None means ff02::2 here
let dst = SocketAddrV6::new(
    dst.unwrap_or(ALL_ROUTERS_MCAST),
    0,
    0,
    ifindex,
);
```
`cksum(None)` internally uses `ALL_NODES_MCAST` (`ff02::1`), which is correct for Router Advertisements. But `send_rs(None)` sends Router Solicitations to `ALL_ROUTERS_MCAST` (`ff02::2`).

ICMPv6 checksums include the IPv6 pseudo-header destination, so multicast RS packets are checksummed for `ff02::1` but sent to `ff02::2`. Receivers should drop them as invalid

taspelund:
The previous BGP unnumbered implementation split ownership across bgp,
mgd, and ndp. Before this change, active interface state effectively
lived in at least three places:

* the manager's scope/name lookup
* ndp::NdpManager's managed interface map
* BGP's mock unnumbered backend used by tests

Each represented the same conceptual active interface from a different
angle, so changes to interface activation, removal, router-discovery
state, or tests had to preserve implicit synchronization between them.

At a high level, the old shape looked like this:
```
  mgd
  ├── pending interfaces
  ├── active interface scope/name map
  └── UnnumberedManagerNdp

  ndp
  ├── NdpManager
  └── managed interface map
      ├── per-interface tx/rx threads
      └── discovered router state

  bgp
  ├── unnumbered trait/types
  └── unnumbered mock backend
```
Restructure the implementation around a dedicated unnumbered crate with
a single active-interface registry:
```
  unnumbered
  ├── BgpUnnumbered trait
  ├── UnnumberedManager
  │   ├── pending interfaces
  │   └── InterfaceMap
  │       └── UnnumberedInterface
  │           ├── ifname
  │           ├── scope_id
  │           ├── local link-local address
  │           └── RouterDiscoveryRuntime
  │               ├── Manual cfg(test)
  │               └── Ndp
  │                   ├── RouterDiscoveryState
  │                   └── ndp::RouterDiscoveryThreads
  └── BGP-facing errors/query types

  ndp
  └── per-interface router-discovery primitives

  bgp
  └── consumes BgpUnnumbered

  mgd
  └── wires/administers UnnumberedManager
```
The resulting crate dependency graph is:
```
  mgd ───────┐
             ▼
          unnumbered ───▶ ndp
             ▲
             │
  bgp ───────┘
```
Now active unnumbered interface state has one source of truth:
InterfaceMap. Presence in that map means the interface is active,
and the stored UnnumberedInterface owns the identity, discovered OS
state, and router-discovery runtime for that interface. There is no
separate NDP manager interface collection to keep synchronized with the
BGP-facing scope/name lookup.

NDP is now compartmentalized as one router-discovery runtime backend.
UnnumberedInterface owns a RouterDiscoveryRuntime enum: production
interfaces use the Ndp backend, which owns RouterDiscoveryState and
ndp::RouterDiscoveryThreads, while tests can use the Manual backend to
manipulate discovered-router state without starting NDP tx/rx threads.
That keeps NDP packet/thread mechanics inside the ndp-backed runtime
instead of leaking a collection-level NDP manager into BGP or mgd.

This also clarifies the layer boundaries:

* BGP asks for only the information it needs:
  * get active interface by scope ID
  * get discovered NDP neighbor by interface
* unnumbered owns pending/active interface lifecycle
* each active interface owns its router-discovery runtime
* ndp owns packet handling and per-interface router-discovery threads
* mgd wires/administers the manager instead of owning its internals

Along the way:

* remove bgp::unnumbered and the heavy BGP unnumbered mock backend
* replace the test seam with small BgpUnnumbered trait stubs and manual
  router-discovery state for manager tests
* fold ScopeMap into InterfaceMap/UnnumberedInterface
* validate active interfaces at construction time:
  * non-zero scope ID
  * unicast link-local local address
* make router-discovery runtime state explicit and non-optional for
  active interfaces
* move latest admin endpoints under /bgp/unnumbered/...
* make latest unnumbered admin queries daemon-global rather than
  ASN-scoped
* rename latest API operation/type names to BGP unnumbered terminology
  while keeping legacy /ndp/... compatibility shims
* consolidate inbound link-local session-key resolution so listeners and
  the dispatcher do not independently resolve and look up the same
  session

The legacy v5 /ndp/... API shape is preserved through downgrade shims,
but the latest v12 API now exposes the unnumbered model directly.

Fixes #801 
Fixes #802


taspelund:
Took this on a trip through a4x2 and it looks good.

Manual testing of the new interface lifecycle (configured interface goes to pending when removed from the system, goes back to active after it's recreated. Output below (tfport automatically recreated by control plane):
```
root@oxz_switch:~# mgadm ndp status manager
BGP Unnumbered Manager State
============================================================

Monitor Thread: Running

Pending Interfaces (configured, waiting for system): 0
  (none)

Active Interfaces: 2
  tfportqsfp1_0
  tfportqsfp0_0
root@oxz_switch:~# ipadm
ADDROBJ           TYPE     STATE        ADDR
lo0/v4            static   ok           127.0.0.1/8
lo0/v6            static   ok           ::1/128
oxControlService0/ll addrconf ok        fe80::8:20ff:fe83:89fd%oxControlService0/10
oxControlService0/omicron6 static ok    fd00:17:1:d01::2/64
oxBootstrap0/ll   addrconf ok           fe80::8:20ff:fee0:2271%oxBootstrap0/10
oxBootstrap0/bootstrap6 static ok       fdb0:a840:2500:1::2/64
tfportrear0_0/ll  addrconf ok           fe80::aa40:25ff:fe0f:b140%tfportrear0_0/10
tfportrear1_0/ll  addrconf ok           fe80::aa40:25ff:fe0f:b141%tfportrear1_0/10
tfportrear2_0/ll  addrconf ok           fe80::aa40:25ff:fe0f:b142%tfportrear2_0/10
tfportrear3_0/ll  addrconf ok           fe80::aa40:25ff:fe0f:b143%tfportrear3_0/10
tfportqsfp0_0/ll  addrconf ok           fe80::aa40:25ff:fe0f:b160%tfportqsfp0_0/10
tfportqsfp1_0/ll  addrconf ok           fe80::aa40:25ff:fe0f:b164%tfportqsfp1_0/10
root@oxz_switch:~# ipadm delete-if tfportqsfp0_0
root@oxz_switch:~# ipadm
ADDROBJ           TYPE     STATE        ADDR
lo0/v4            static   ok           127.0.0.1/8
lo0/v6            static   ok           ::1/128
oxControlService0/ll addrconf ok        fe80::8:20ff:fe83:89fd%oxControlService0/10
oxControlService0/omicron6 static ok    fd00:17:1:d01::2/64
oxBootstrap0/ll   addrconf ok           fe80::8:20ff:fee0:2271%oxBootstrap0/10
oxBootstrap0/bootstrap6 static ok       fdb0:a840:2500:1::2/64
tfportrear0_0/ll  addrconf ok           fe80::aa40:25ff:fe0f:b140%tfportrear0_0/10
tfportrear1_0/ll  addrconf ok           fe80::aa40:25ff:fe0f:b141%tfportrear1_0/10
tfportrear2_0/ll  addrconf ok           fe80::aa40:25ff:fe0f:b142%tfportrear2_0/10
tfportrear3_0/ll  addrconf ok           fe80::aa40:25ff:fe0f:b143%tfportrear3_0/10
tfportqsfp1_0/ll  addrconf ok           fe80::aa40:25ff:fe0f:b164%tfportqsfp1_0/10
root@oxz_switch:~#
root@oxz_switch:~# mgadm ndp status manager
BGP Unnumbered Manager State
============================================================

Monitor Thread: Running

Pending Interfaces (configured, waiting for system): 1
  tfportqsfp0_0 (router_lifetime: 1800s)

Active Interfaces: 1
  tfportqsfp1_0
root@oxz_switch:~# mgadm ndp status manager
BGP Unnumbered Manager State
============================================================

Monitor Thread: Running

Pending Interfaces (configured, waiting for system): 0
  (none)

Active Interfaces: 2
  tfportqsfp1_0
  tfportqsfp0_0
root@oxz_switch:~# 
```

Similarly, the scope_id changes are reflected in the lifecycle too:
```
root@oxz_switch:~# mgadm ndp status interfaces
Interface      Local Address              Scope  Discovered Peer                        TX   RX   Reachable
tfportqsfp1_0  fe80::aa40:25ff:fe0f:b164  9      fe80::aa40:25ff:fe00:11%tfportqsfp1_0  Run  Run  Yes
tfportqsfp0_0  fe80::aa40:25ff:fe0f:b160  10     fe80::aa40:25ff:fe00:9%tfportqsfp0_0   Run  Run  Yes
root@oxz_switch:~#
root@oxz_switch:~#
root@oxz_switch:~#
root@oxz_switch:~# ipadm delete-if tfportqsfp0_0
root@oxz_switch:~# mgadm ndp status interfaces
Interface      Local Address              Scope  Discovered Peer                        TX   RX   Reachable
tfportqsfp1_0  fe80::aa40:25ff:fe0f:b164  9      fe80::aa40:25ff:fe00:11%tfportqsfp1_0  Run  Run  Yes
tfportqsfp0_0  fe80::aa40:25ff:fe0f:b160  10     fe80::aa40:25ff:fe00:9%tfportqsfp0_0   Run  Run  Yes      <<<<  still here until next monitor sweep
root@oxz_switch:~# mgadm ndp status interfaces
Interface      Local Address              Scope  Discovered Peer                        TX   RX   Reachable
tfportqsfp1_0  fe80::aa40:25ff:fe0f:b164  9      fe80::aa40:25ff:fe00:11%tfportqsfp1_0  Run  Run  Yes     <<<<  deleted from active interfaces
root@oxz_switch:~# mgadm ndp status interfaces
Interface      Local Address              Scope  Discovered Peer                        TX   RX   Reachable
tfportqsfp1_0  fe80::aa40:25ff:fe0f:b164  9      fe80::aa40:25ff:fe00:11%tfportqsfp1_0  Run  Run  Yes
tfportqsfp0_0  fe80::aa40:25ff:fe0f:b160  11     fe80::aa40:25ff:fe00:9%tfportqsfp0_0   Run  Run  Yes     <<<< re-added
```

taspelund:
I realized while testing this that I forgot to update the mgadm syntax to align with the new API filepath, so I'm going to update that now.

taspelund:
Had to pull in latest main to get the juniper falcon-lab fix for CI. Should be good to go now.

taspelund:
I've added an omicron branch which pulls in these new API endpoints, and it all passes a4x2.
omicron PR here: https://github.com/oxidecomputer/omicron/pull/10705

mg-api-types/versions/src/unnumbered_refactor/unnumbered.rs
```diff
@@ -0,0 +1,289 @@
...
+}
+
+/// Unnumbered manager state showing overall health and interface status.
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+pub struct UnnumberedManagerState {
+    /// Whether the interface monitor thread is running
+    pub monitor_running: bool,
+    /// Status of every interface configured for unnumbered operation,
+    /// keyed by interface name.
+    pub interfaces: BTreeMap<String, UnnumberedInterfaceStatus>,
```
cfzimmerman, line 38: Two questions / comments. Possibly naive as I'm not fully clear what the versioning requirements are.
 
- Why `BTreeMap` and not `HashMap`? If ordering/stability is important, I often appreciate a comment why. (Unless that's obvious to a regular contributor)
- When the contents of a string on the heap should never change after initialization, I often prefer `Box<str>` over `String` as a way to express that. Imo the purpose of a `String` type is to be mutable, and idk if that's needed here. But no need to change if that's an implementation lift.

mg-api-types/versions/src/unnumbered_refactor/unnumbered.rs
```diff
@@ -0,0 +1,289 @@
...
+    pub time_since_last_rx: Duration,
+    /// Effective reachable time governing expiry of this entry
+    pub effective_reachable_time: Duration,
+    /// Router lifetime from RA (seconds)
+    pub router_lifetime: u16,
+    /// Reachable time from RA (milliseconds)
+    pub reachable_time: u32,
+    /// Retransmit timer from RA (milliseconds)
+    pub retrans_timer: u32,
+}
```
cfzimmerman, line 100: Some of the time units in this module use `Duration`, and others use a number + comment. Can they all just be `Duration`?

cfzimmerman: I guess that forces handling edge cases if the underlying protocol wants these types. Like saturating an excessive millisecond count to `u32` or handling that as an error. Is that why you did it this way?

mg-api-types/versions/src/unnumbered_refactor/unnumbered.rs
cfzimmerman: Reading further, order stability makes sense if there are `get_*_state` HTTP endpoints. So standard practice is to choose `BTreeMap` for any serialized k:v collection?

mg-common/src/thread.rs
```diff
@@ -33,9 +33,16 @@ impl ThreadState {
         matches!(self, ThreadState::Ready)
     }
 
-    /// Check if the thread is currently running
+    /// Check if the thread is currently running.
+    ///
+    /// A thread that has been started but whose function has since returned
+    /// (or panicked) is not running: liveness comes from the JoinHandle, not
+    /// from the state transition.
```
cfzimmerman, line 40: Is the part after the colon necessary? The first clause seems clear enough to me.

mgadm/src/unnumbered.rs
```diff
@@ -0,0 +1,222 @@
...
+pub enum StatusCmd {
+    /// Show BGP unnumbered manager state
+    Manager,
+
+    /// List all active BGP unnumbered interfaces
+    Interfaces,
+
+    /// Show detailed state for a specific interface
+    Interface { interface: String },
+}
```
cfzimmerman, line 36: This seems kind of verbose. Is it possible to accomplish the goal with fewer than three new types?

mgd/src/bgp_admin.rs
```diff
@@ -529,173 +526,187 @@ pub async fn clear_unnumbered_neighbor(
...
+    get_router!(ctx, rq.asn)?
+        .get_session(rq.interface.clone())
+        .ok_or(Error::NotFound(
+            "session for unnumbered neighbor not found".into(),
+        ))?;
+
+    let ndp_detail = ctx
+        .bgp
+        .unnumbered_manager
+        .get_interface_detail(&rq.interface)
```
cfzimmerman: If the first block is a check for subsequent logic, is there any risk of a race condition here?

ndp/src/packet.rs
```diff
@@ -146,14 +179,80 @@ impl Default for Icmp6RouterSolicitation {
...
+        }
+        offset = next;
+    }
+
+    Ok(has_slla)
+}
+
 #[derive(Debug, thiserror::Error)]
-pub enum Icmp6RsFromWireError {
+pub enum Icmp6RsParseError {
```
cfzimmerman, line 218: Nice descriptive errors

ndp/src/router_discovery.rs
```diff
@@ -0,0 +1,668 @@
...
+}
+
+/// The `RouterDiscovery` worker runs router discovery for an individual interface.
+///
+/// Discovery is started on construction and stopped automatically when the
+/// worker is dropped via the `ManagedThread` Drop implementation.
+#[derive(Debug)]
+pub struct RouterDiscoveryThreads {
+    /// Handle to the transmit loop thread
+    _tx_thread: Arc<ManagedThread>,
```
cfzimmerman: Why `_` prefix?

mg-api-types/versions/src/unnumbered_refactor/unnumbered.rs
cfzimmerman: Reading even further, I see that the interface type is pretty locked in and not worth messing with rn.

Long run, however, I'm intrigued by the prospect of using a stack allocated string instead. We could then have a  copyable `Interface` newtype instead of using `String`. Looks like illumos bounds these at 32 bytes anyway:  https://www.smartos.org/man/4p/if#:~:text=%23define%20LIFNAMSIZ%20%20%2032

unnumbered/src/error.rs
```diff
@@ -0,0 +1,36 @@
...
+/// Error type for UnnumberedManager operations.
+#[derive(Debug, Clone)]
+pub enum UnnumberedError {
+    /// Interface not found on the system
+    InterfaceNotFound(String),
+    /// Interface exists but is not IPv6
+    NotIpv6(String),
+    /// Other interface resolution error
+    ResolutionFailed { interface: String, reason: String },
+}
```
cfzimmerman, line 14: I'm curious when you decide to use `thiserror` or not. I see it above in `mgd` and below in `interface.rs` but not here or in this PR: https://github.com/oxidecomputer/oxnet/pull/84

I default to always using `thiserror`, so I'm intrigued when you choose otherwise.

unnumbered/src/manager.rs
```diff
@@ -0,0 +1,1121 @@
...
+        activate: F,
+    ) where
+        F: Fn(&Ipv6NetworkInterface, u16) -> Result<UnnumberedInterface, E>,
+        E: std::fmt::Display,
+    {
+        let mut to_deactivate = Vec::new();
+
+        {
+            let configured = lock!(configured_interfaces);
+            let active = lock!(active_interfaces);
```
cfzimmerman, line 267: If we're locking these two simultaneously, perhaps we want an "Invariant" doc comment somewhere that the lock order must always be `(configured) -> (active)`.

unnumbered/src/manager.rs
```diff
@@ -0,0 +1,1121 @@
...
+        let ndp::RouterAdvertisementInfo {
+            address,
+            first_seen,
+            last_seen,
+            router_lifetime,
+            reachable_time,
+            effective_reachable_time,
+            retrans_timer,
+            expired,
+        } = info;
```
cfzimmerman: If so inclined, you can also unpack directly in the function signature

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7b34716a9c1348c31ba9df5891e8df01

cfzimmerman (commented):
Big project and an interesting feature. Most of these are just questions, some of which I answered myself :)
Looks good!

mg-api-types/versions/src/unnumbered_refactor/unnumbered.rs
taspelund: I don't have strong reasoning for picking BTreeSet over HashSet. The biggest motivation for me picking it is because it seems to be the common choice for set collections elsewhere in the codebase, so I followed that without spending a whole lot of time on the collection itself. I chose a set because I wanted the automatic de-duplication we get over something like a Vec.

I do like the idea of a Box<str> to make it immutable (hopefully without the hassle of lifetime annotations everywhere?).

I know one thing we're still figuring out is how to make a better type wrapper between omicron and maghemite around illumos interface names (well, tfport names in particular), but it hasn't really been solidified yet... so that's why we're using String here currently.

mg-api-types/versions/src/unnumbered_refactor/unnumbered.rs
taspelund: To be honest, I was mainly focused on trying to standardize on `Duration` instead of stringly-typed timestamps and didn't really think about converting these uints into Durations. That seems like a reasonable thing to do, so I'll convert the rest for uniformity.

mg-common/src/thread.rs
taspelund: I'll clean it up, thanks!

unnumbered/src/manager.rs
taspelund: Oh that's slick. I didn't even know this was an option. Thanks!

unnumbered/src/manager.rs
taspelund: good call. will add a clarification comment

unnumbered/src/error.rs
taspelund: I haven't personally spent a lot of time with `thiserror`, so I tend to follow whatever patterns are preexisting in the codebase. For the oxnet PR, I didn't want to add a new repo dependency when the tradeoff was 10 or so extra lines of code. For this one, it seemed like the number of variants was relatively small and straightforward, so it just didn't occur to me to reach for `thiserror`

taspelund: Do you feel like this is a good candidate for `thiserror`? If so, I'd be interested to hear more!

ndp/src/router_discovery.rs
taspelund: I think at the time I added this, they were unused... but it's clearly not the case anymore so I'll fix it. Thanks!

mgd/src/bgp_admin.rs
taspelund: Yes, there is. However, this is a read-only endpoint for a legacy API version. And omicron was never plumbed to use or expose this endpoint. So there aren't really any potential callers of this aside from old builds of mgadm.
Looking back over this did make me realize that I don't like the current shape of the handlers, so I'm going to clean that up though :D

unnumbered/src/error.rs
cfzimmerman: `thiserror` is nice for standardization and minimizing LOC, but deferring to precedent makes sense. Was just curious if any past experience with limits or footguns in `thiserror` motivated a pattern.

PR #803, opened 2026-06-25
Yank out type duplication for BGP peers

### LINKED ISSUE #564: Type consolidation: remove/consolidate redundant types similar to bgp::session::SessionInfo
There look to be at least 5 unique types that all have similar contents/uses, which is to represent a BGP neighbor and/or its configuration:
- `rdb::types::BgpNeighborInfo`
- `bgp::session::SessionInfo`
- `bgp::params::BgpPeerConfig`
- `bgp::config::PeerConfig`
- `bgp::params::Neighbor`

These should be cleaned up to the extent that is reasonable.

taspelund:
This PR removes the duplicated numbered/unnumbered BGP peer configuration paths introduced as BGP unnumbered support grew. The latest implementation goes from 11 overlapping API, storage, and session shapes to four core types:

1. `NeighborConfig`
2. `Neighbor`
3. `SessionInfo`
4. `NeighborInfo`

Older shapes remain only in versioned compatibility APIs. Overall, this removes more than 1,000 lines while making numbered and unnumbered peers follow the same configuration, persistence, and session-management paths.

## API and type model

The new `UNIFY_BGP_NEIGHBORS` API is version 13 and is exposed through `latest`.

- `PeerId` remains solely the peer identity: `Ip` for numbered peers or `Interface` for unnumbered peers.
- `NeighborConfig` contains that identity, `group`, the optional TCP `port`, and all session parameters. The former `BgpPeerParameters` layer is inlined without `serde(flatten)`.
- `Neighbor` is the read/stored composition `{ asn, config: NeighborConfig }`.
- Single-neighbor create and update operations continue to use `/bgp/config/neighbor`, with ASN carried by `Neighbor`.
- Numbered and unnumbered clear operations use one `NeighborResetRequest` and one handler.

`ApplyRequest` is now an explicitly tagged operation:

```json
{"action":"apply","asn":47,"originate":[],"peers":[]}
{"action":"delete","asn":47}
```

`Apply` represents complete desired state for a router. An empty peer list retains the router, while `Delete` explicitly removes the router and its state.

Apply peers are represented internally as `IdOrdMap<NeighborConfig>`, keyed only by `PeerId`, so one peer cannot appear in multiple groups. Serde and OpenAPI retain a normal JSON array representation, and Progenitor generates `Vec<NeighborConfig>` for clients. Duplicate peer identities are rejected when the request is deserialized.

## Internal simplification

- Removed `bgp::config::PeerConfig` and its conversion layer.
- Removed the latest-path `BgpPeerConfig`, `UnnumberedBgpPeerConfig`, `BgpPeerParameters`, `BgpNeighborInfo`, `BgpUnnumberedNeighborInfo`, and `BgpNeighborParameters` duplication.
- The RDB now persists `Neighbor` directly in one tree keyed by `(ASN, PeerId)` instead of maintaining separate numbered and unnumbered storage paths.
- Router/session creation, ensure, update, reset, and teardown now dispatch through `PeerId` rather than parallel numbered/unnumbered implementations.
- The mgd API handlers are collapsed onto unified create/read/update/delete/clear operations. Older split endpoints remain as version-bounded adapters.
- mgadm and falcon-lab construct the same nested `Neighbor` shape used by the API and RDB.
- `deterministic_collision_resolution` and timer jitter settings now survive persistence instead of being reconstructed with defaults.

## Compatibility and review fixes

- v11 requests upgrade into the unified v13 shape; a peer appearing in multiple legacy groups is rejected with HTTP 400.
- v13 apply requests downgrade exhaustively by matching `PeerId::Ip` and `PeerId::Interface`, rather than silently dropping a peer kind. Explicit router deletion cannot be represented by v11 and returns a conversion error.
- Compatibility read endpoints return 404 when a unified lookup finds the wrong peer kind.
- Session updates validate timer changes before mutating peer metadata.
- Applying an empty peer set no longer removes the router and then fails while updating origins.
- Applying a new ASN tears down displaced router state through the normal router-deletion path.

## Persistence note

The sled database is used for recovery from daemon crashes and restarts within a deployed zone. mgd upgrades replace and rebuild the zone, so no cross-version migration of the retired numbered/unnumbered neighbor tree is required.

## Validation

- `cargo fmt --all --check`
- `cargo test -p mg-api-types-versions`
- targeted mgd BGP apply tests
- Clippy with warnings denied across affected crates
- `just openapi-check mgd`
- `cargo check --manifest-path mg-admin-client/Cargo.toml`

Fixes: #564


taspelund:
marking this ready for review now. I still need to build out the omicron PR, but I don't think that should hold up any reviews on the maghemite side

rdb/src/db.rs
```diff
@@ -425,45 +399,6 @@ impl Db {
         Ok(result)
     }
 
-    pub fn get_unnumbered_bgp_neighbors(
-        &self,
-    ) -> Result<Vec<BgpUnnumberedNeighborInfo>, Error> {
-        let tree = self.persistent.open_tree(BGP_UNNUMBERED_NEIGHBOR)?;
```
nicolaskagami, line 431: Are we ok with ignoring this tree on the transition to this version? There's no attempt to rescue it.

mgd/src/bgp_admin.rs
```diff
@@ -1515,14 +1214,27 @@ async fn do_bgp_apply(
...
+
+            let mut routers = lock!(ctx.bgp.router);
+            let mut remove = false;
+            if let Some(r) = routers.get(&nbr.asn) {
+                remove = lock!(r.sessions).is_empty();
+            }
+            if remove && let Some(r) = routers.remove(&nbr.asn) {
+                r.shutdown()
+            };
+        }
```
nicolaskagami: We're deleting the empty router here, but we're also trying to get a router immediately after with `get_router!` so I imagine we'll error out with a 404 `NotFound` from it on a legal "drain all" operation.

bgp/src/session.rs
```diff
@@ -8807,11 +8770,18 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
     ) -> Result<(), Error> {
-        let mut reset_needed = self.update_session_config(cfg)?;
-        reset_needed |= self.update_session_info(info)?;
+        // Identity is enforced by the lookup mechanism (we only update a
+        // session found via its PeerId); name and group are mutable metadata.
+        // All other parameters live in SessionInfo.
+        let name = lock!(neighbor.name).clone();
+        let peer_group = lock!(neighbor.peer_group).clone();
+        *lock!(self.neighbor.name) = name;
+        *lock!(self.neighbor.peer_group) = peer_group;
```
nicolaskagami, line 8790: Do we want to write here before validating `update_session_info`? I'd assume not at least in the long term.

nicolaskagami (commented):
Nice! It's always great to see code being simplified away :)

I found a couple of things that are worth addressing down below and:
- The PR description mentions a `NeighborGroupSelector` which is nowhere to be found in this repo.
- I like `PeerId` but I think we could do more of it, such as with `act_as_a_default_ipv6_router`. I'm thinking about a better format overall.

rdb/src/db.rs
taspelund: So the sled::Db actually lives on disk solely for the purposes of recovery upon daemon crash/restart. All upgrades of mgd are deployed via a full teardown and rebuild of an Illumos zone. So there's no backwards compat we need to worry about with the on-disk db

mgd/src/bgp_admin.rs
taspelund: Good call. I don't recall why I added that, but I've pushed new commits that address this and make the `ApplyRequest` (the type exposed to omicron via the API endpoint) include an enum that allows for full deletion of BGP routers... so this is no longer ambiguous.

bgp/src/session.rs
taspelund: Good call. Fixed in latest commits.

taspelund:
> Nice! It's always great to see code being simplified away :) 

Thanks! I am much happier with this shape than I was with the parallel un/numbered paths.

>     * The PR description mentions a `NeighborGroupSelector` which is nowhere to be found in this repo.

I fixed up the PR description (rather, I had claude do it because there are a lot of changes in this PR) with a summary of what's been done.

>     * I like `PeerId` but I think we could do more of it, such as with `act_as_a_default_ipv6_router`. I'm thinking about a better format overall.

I understand the urge to move things like `act_as_a_default_ipv6_router` into `PeerId::Interface`. It's the only part of any BGP peer type that is truly unique to unnumbered peers, and structurally it would make sense for that to be where unnumbered attributes live. That said, I am sort of partial to the idea of having this be a small and simple enum that only does what it says on the tin: it identifies a peer.

Maybe there's another shape we could pursue where `PeerId` is only ever derived, rather than it also functioning as a config shape? That seems like it would give the best of both worlds -- `PeerId` stays relatively pure, un/numbered-specific config structurally prevented from being unrepresentable on the types where it doesn't apply.

nicolaskagami:
> I understand the urge to move things like act_as_a_default_ipv6_router into PeerId::Interface. It's the only part of any BGP peer type that is truly unique to unnumbered peers, and structurally it would make sense for that to be where unnumbered attributes live. That said, I am sort of partial to the idea of having this be a small and simple enum that only does what it says on the tin: it identifies a peer.

That's not exactly what I was suggesting, exactly for the reason you give.

> Maybe there's another shape we could pursue where PeerId is only ever derived, rather than it also functioning as a config shape? That seems like it would give the best of both worlds -- PeerId stays relatively pure, un/numbered-specific config structurally prevented from being unrepresentable on the types where it doesn't apply.

Yep. We could have `PeerId` come from a method instead of being a field and have an enum for exclusive fields.

bgp/src/connection_channel.rs
rcgoodfellow, line 1: Given all the testing infrastructure we've built up with falcon lab and the amount of additional complexity abstracting connections over multiple underlying substrates costs, i'm not sure if its worth keeping the channel model and the indirection layer around.

I think there has also been some work recently on running tests based on multiple `mgd` daemons on the same host peering with each other over localhost addresses. That seems like it be more worthwhile than the channel-based tests.

bgp/src/connection_tcp.rs
```diff
@@ -207,6 +207,7 @@ impl BgpListener<BgpConnectionTcp> for BgpListenerTcp {
         conn: &BgpConnectionTcp,
         min_ttl: Option<u8>,
         md5_key: Option<String>,
+        _listen_port: std::num::NonZeroU16,
```
rcgoodfellow, line 210: remove leading `_`. You can use `let _ = listen_port;` in the linux stub.

bgp/src/session.rs
```diff
@@ -799,6 +788,102 @@ impl UnusedEvent {
     }
 }
 
+/// Address and port configuration for a session.
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, PartialEq, Eq)]
+pub struct SessionAddrInfo {
+    /// TCP port configured on the remote peer. `BGP_PORT` is used when the peer
```
rcgoodfellow, line 794: What is meant by the second sentence in this comment? `remote_port` is not optional here. If this is about how a user of this structure fills this field in, I think that commentary should go in the using code and not here as this can very easily get out of sync.

bgp/src/session.rs
```diff
@@ -8807,71 +8771,62 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
                 .map_err(|e| Error::EventSend(e.to_string()))?;
         }
 
+        // Identity is enforced by the lookup mechanism (we only update a
+        // session found via its PeerId); name and group are mutable metadata.
+        // All other parameters live in SessionInfo.
+        let name = lock!(neighbor.name).clone();
+        let peer_group = lock!(neighbor.peer_group).clone();
+        *lock!(self.neighbor.name) = name;
+        *lock!(self.neighbor.peer_group) = peer_group;
```
rcgoodfellow, line 8790: This can probably simplify to

```rust
        *lock!(self.neighbor.name) = lock!(neighbor.name).clone();
        *lock!(self.neighbor.peer_group) = lock!(neighbor.peer_group).clone();
```

To reduce the chance of `name` and `peer_group` accidentally being held to long if code gets added after this later.

mg-api-types/versions/src/impls/bgp/config.rs
```diff
@@ -2,273 +2,87 @@
...
-// Each conversion below uses full-struct destructuring at the source
-// boundary as a compile barrier: adding a field to a published or
-// runtime type fails to bind here, forcing a deliberate decision about
-// how (or whether) the new field threads through to the target type.
-// Bindings prefixed with `_:` are intentionally dropped — that decision
-// is documented at the destructure site.
+impl Default for JitterRange {
+    fn default() -> Self {
+        Self {
+            min: 0.75,
```
rcgoodfellow, line 14: These constants deserve some comments.

mg-api-types/versions/src/mp_bgp/bgp/config.rs
```diff
@@ -188,6 +167,236 @@ pub struct ApplyRequest {
...
+            ipv4_unicast: Some(Ipv4UnicastConfig {
+                nexthop: None,
+                import_policy: ImportExportPolicy4::from(allow_import),
+                export_policy: ImportExportPolicy4::from(allow_export),
+            }),
+            ipv6_unicast: None,
+            deterministic_collision_resolution: false,
+            idle_hold_jitter: None,
+            connect_retry_jitter: Some(JitterRange {
+                min: 0.75,
```
rcgoodfellow, line 218: second time these magic constants are coming up. should probably be a named constant with comments explaining why these numbers.

mg-api-types/versions/src/unify_bgp_neighbors/bgp/config.rs
```diff
@@ -0,0 +1,632 @@
...
+use crate::v4::bgp::config::JitterRange;
+use crate::v11::bgp::config::{Ipv4UnicastConfig, Ipv6UnicastConfig};
+
+/// Default TCP port used for BGP connections.
+const BGP_PORT: u16 = 179;
+
+/// BGP neighbor configuration: everything about a peer except the `asn` router
+/// it belongs to.
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, PartialEq)]
+pub struct NeighborConfig {
```
rcgoodfellow, line 40: Should probably have doc comments on all members for API structs? Same goes for the structs that follow below.

mg-api-types/versions/src/unify_bgp_neighbors/bgp/config.rs
```diff
@@ -0,0 +1,632 @@
...
+        /// `peer` must be unique; duplicate peer identities are rejected.
+        peers: IdOrdMap<NeighborConfig>,
+    },
+    /// Delete the router and all state owned by it.
+    Delete { asn: u32 },
+}
+
+// ===== upgrade conversions (v11 split -> v13 unified), total =====
+
+impl From<crate::v11::bgp::config::Neighbor> for Neighbor {
```
rcgoodfellow, line 122: There is a LOT of version conversion going on here. It seems like we should account for this in our e2e tests like falcon lab by creating variants of our tests that drive things from clients that are one or even a few versions behind.

mg-api-types/versions/src/unify_bgp_neighbors/rdb/mod.rs
rcgoodfellow, line 1: I think this file can be removed?

mg-api-types/versions/src/unify_bgp_neighbors/rdb/neighbor.rs
rcgoodfellow, line 1: I think this file can be removed?

mg-api/src/lib.rs
```diff
@@ -384,19 +456,8 @@ pub trait MgAdminApi {
         Self::update_neighbor_v8(rqctx, request.map(Into::into)).await
     }
 
-    // V1/V2 API - legacy Neighbor type with combined import/export policies.
```
rcgoodfellow, line 387: At some point we can probably start pruning when it's guaranteed that things as far back as V1/V2 will never be in use?

rdb/src/db.rs
```diff
@@ -79,6 +75,33 @@ const BESTPATH_FANOUT: &str = "bestpath_fanout";
...
+    &key[ASN_KEY_LEN..]
+}
+
+/// Remove every entry in `tree` scoped to `asn`.
+fn remove_asn_scoped(tree: &sled::Tree, asn: u32) -> Result<(), Error> {
+    for item in tree.scan_prefix(asn.to_be_bytes()) {
+        let (key, _) = item?;
+        tree.remove(key)?;
+    }
+    Ok(())
```
rcgoodfellow, line 102: I think we may need a `tree.flush()` here to ensure that all the removes took place by the time this function returns.

rcgoodfellow (commented):
Thanks for the consolidation work here @taspelund. Comments follow.

Since the API surface change is large for this PR, I'd like to see an omicron branch that's working e2e in CI and in voxel before this lands.

PR #807, opened 2026-06-26
buildomat: collect .falcon if falcon-lab fails

taspelund:


.github/buildomat/jobs/falcon-lab.sh
```diff
@@ -70,12 +70,38 @@ gw=$(bmat address ls -f extra -Ho gateway)
...
+	if [[ ! -d .falcon ]]; then
+		return 0
+	fi
+
+	# Preserve only propolis logs before cleanup removes them. Avoid archiving
+	# arbitrary workspace contents, since cargo-bay may contain secrets and
+	# future Falcon versions may add other files here.
+	rm -rf "${artifact_dir}"
+	mkdir -p "${artifact_dir}"
+	for path in .falcon/*.{out,err}; do
```
nicolaskagami, line 87: Are all `out`/`err` files of interest in this folder? This doesn't catch anything in subdirectories.

.github/buildomat/jobs/falcon-lab.sh
```diff
@@ -70,12 +70,38 @@ gw=$(bmat address ls -f extra -Ho gateway)
...
 pfexec ./dhcp-server "${first}" "${last}" "${gw}" "${server}" &> /work/dhcp-server.log &
 
+collect_falcon_artifacts() {
+	local test_name=$1
+	local artifact_dir="/work/falcon-${test_name}"
+	local path
+
+	if [[ ! -d .falcon ]]; then
+		return 0
+	fi
```
nicolaskagami, line 80: We cleanup `"${artifact_dir}"` before and after but never `.falcon`. Couldn't it persist data between runs?

nicolaskagami (approved):
LGTM, whatever that's worth as a newbie :)

Just had a couple of questions below.

.github/buildomat/jobs/falcon-lab.sh
taspelund: As of today, the only items in subdirectories are binaries. So this deliberately leaves those items behind. Good question though!

.github/buildomat/jobs/falcon-lab.sh
taspelund: I _think_ they'll be overwritten each time the topology is launched, rather than appended to. But that's a good call out, I'll double check

taspelund: It looks like they're overwritten/truncated. The binary calls `std::fs::File::create` without any open options, so they should be clean. That said, it's an easy thing to add just to have an extra level of certainty in case that changes

taspelund: Ok, it actually looks like falcon cleans up everything in .falcon/ except for .falcon/bin since that has the propolis-server binary and the OVMF firmware.

So I'll add in a removal as a defensive measure, but will leave .falcon/bin

PR #819, opened 2026-07-08
ddmd/mgd: publish bound admin port to a file

sunshowers:
This is useful for omicron tests that tell the services to bind to port 0.

Let's rely on a simple, stable format for this rather than querying the kernel (which can [cause process crashes](https://www.illumos.org/issues/18222)) or grepping through logs.


Cargo.toml
```diff
@@ -85,19 +85,21 @@ slog-term = "2.9"
...
 dropshot-api-manager-types = "0.7.2"
 expectorate = "1.2.0"
 schemars = { version = "0.8.22", features = [ "uuid1", "chrono" ] }
 tokio = { version = "1.49", features = ["full"] }
 serde_repr = "0.1"
 anyhow = "1.0.103"
+port-file = "0.1.0"
 hyper = "1.10.1"
 hyper-util = { version = "0.1", features = ["full"] }
 serde_json = "1.0.149"
```
sunshowers, line 105: Might be worth sorting these in the future.

taspelund (approved):
LGTM overall. Seems like this is supposed to more or less be the ephemeral port version of a pidfile?

One question: is there a reason we don't have ddmd/mgd to cleanup the portfile on shutdown?

taspelund:
The CI job `falcon` is failing due to known flakiness when using falcon's serial console. Re-running

sunshowers:
> LGTM overall. Seems like this is supposed to more or less be the ephemeral port version of a pidfile?

Yeah basically. (The port-file crate can also be used for pids, JSON blobs, or really anything that's serializable as text, but I wanted to market it as something for ephemeral ports specifically.)

> One question: is there a reason we don't have ddmd/mgd to cleanup the portfile on shutdown?

I don't think that's right in general, because if the port file gets written out and then gets deleted, the test harness might not realize that the port file got written out and might end up spinning until the timeout. In general we recommend that the harness/supervisor create a fresh temporary directory, and it'll be responsible for cleaning that up per standard `tempfile`/`camino-tempfile` semantics.

taspelund:
Ok, works for me!

PR #853, opened 2026-07-27
justfile: expand targets to support ddm

taspelund:


nicolaskagami (approved):
LGTM and `just` works on my linux (no kstat) though a test failed (`test_neighbor_thread_lifecycle_no_leaks`) and a few timed out

taspelund:
> LGTM and `just` works on my linux (no kstat) though a test failed (`test_neighbor_thread_lifecycle_no_leaks`) and a few timed out

I develop locally on Mac and on my Illumos workstation, so the Linux tests don't get a lot of mileage... At least from me. I do think the failures are probably worth chasing down though. Could you file an issue with the failed tests and their logs? I'll chase it down when I have time unless you wanted to give it a shot

PR #869, opened 2026-07-31
[1/n] bgp: refactor SessionRunner so unsafe impl Send/Sync become unnecessary

sunshowers:
Previously, the `SessionRunner` type carried a hand-written `unsafe impl Send/Sync`. This was only required by `event_rx`, and safety was only upheld through an unwritten module-level invariant that at most one thread is ever inside the `fsm_start` loop. This invariant was enforced by the `running` compare-exchange.

Move the receiver into a mutex from which it is grabbed before the `fsm_start` loop, and thread it down to method calls. This allows the types to automatically implement `Send` and `Sync`.

Also clean up some other `unsafe impl Send/Sync` instances that are no longer (or never were) necessary.

The natural next step is to make the `running` atomic bool go away, and the next commit in the stack does this.

### DIFF
--- a/bgp/src/session.rs
+++ b/bgp/src/session.rs
@@ -1691,7 +1691,7 @@ pub struct FsmDriver<Cnx: BgpConnection + 'static> {
 }
 
 impl<Cnx: BgpConnection + 'static> FsmDriver<Cnx> {
-    pub fn new(
+    pub(crate) fn new(
         runner: Arc<SessionRunner<Cnx>>,
         event_rx: Receiver<FsmEvent<Cnx>>,
     ) -> Self {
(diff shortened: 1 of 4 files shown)


bgp/src/session.rs
```diff
@@ -1673,8 +1681,60 @@ pub struct SessionRunner<Cnx: BgpConnection + 'static> {
...
+            "receiver available to the thread that won the running CAS",
+        );
+
+        self.runner.run_fsm(&event_rx);
+
+        // Restore the event receiver *before* on_shutdown clears `running`.
+        // This order is important for LIFO discipline. (But it also suggests
+        // that running is redundant and should go away.)
+        *lock!(self.event_rx) = Some(event_rx);
+        self.runner.on_shutdown();
```
sunshowers, line 1735: I considered using an RAII guard here but we ship with `panic = "abort"` so it is pretty moot.

taspelund (approved):

PR #870, opened 2026-07-31
[2/n] bgp: retire the SessionRunner::running atomic

sunshowers:
Followup to #869. Retire the `SessionRunner::running` atomic and make `event_rx` the source of truth.

There is a very small behavioral difference here (a session is now unstartable for all of `on_shutdown` rather than just up to the point where `running` was cleared) but that seems like an improvement.

Also add a warning about a start being rejected. This is never expected during a normal single-spawn lifecycle, and when it does fire it is either a duplicate start or a restart race.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -736,9 +736,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.6.4"
+version = "4.6.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d91e0c145792ef73a6ad36d27c75ac09f1832222a3c209689d90f534685ee5b7"
+checksum = "301b56658598e48f3648647ac6fc887be7e7108eddfa4e9b63fcf3ec58c0cadf"
 dependencies = [
  "clap_builder",
  "clap_derive",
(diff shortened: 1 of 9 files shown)


taspelund:
The falcon CI job failure looks to be another classic issue with the serial protocol -- this time with RC codes not being propagated from the guest up to falcon.

We already want to replace the serial protocol but that's a non-trivial amount of work. In the meantime, I've submitted [an improvement to softnpu](https://github.com/oxidecomputer/softnpu/pull/224) to retry its `pkg install` if it gets back `EXIT_LOCK` -- since that's retryable.

taspelund (approved):
easy win with a nice clean outcome. approved!

nicolaskagami (approved):
LGTM

I think we're safely coalescing needlessly disparate state but it might add some flakiness to some bgp tests.

Some tests (all derivatives of `basic_peering_helper`) check for the state to move to `Idle` before trying to restart the session. Since the state transitions to `Idle` before the the `event_rx` is returned the `fsm_start` could lose the race.

I see you have more planned changes on top of this, so I'll just point out that we should just care that we gate the state (and its expectation of being able to start) around the lock for the long term.

sunshowers:
> I think we're safely coalescing needlessly disparate state but it might add some flakiness to some bgp tests.

I don't quite understand -- how is this PR making things worse in that regard? This state would have been previously reachable, no?

nicolaskagami:
> how is this PR making things worse in that regard? This state would have been previously reachable, no?

Yeah, it was reachable but it just jumped a bit at me that we were adding a log call between  `Idle` being set and our gate. TBF, I hadn't noticed that we already had a log call there. I guess there's barely any difference then. I was mostly trying to call attention to the race in general.

PR #871, opened 2026-07-31
[3/n] bgp: track shutdown completion as distinct from not-requested

sunshowers:
Replace the `AtomicBool` with an atomic over a three-state enum. In upcoming commits in the stack, we're going to test for the value here.

Note that  `request_shutdown` remains an unconditional store, preserving the previous `store(true)` semantics including a latent hazard where a request landing on an already-stopped runner makes its next start stop immediately.

This commit has no externally-visible behavior changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -736,9 +736,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.6.4"
+version = "4.6.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d91e0c145792ef73a6ad36d27c75ac09f1832222a3c209689d90f534685ee5b7"
+checksum = "301b56658598e48f3648647ac6fc887be7e7108eddfa4e9b63fcf3ec58c0cadf"
 dependencies = [
  "clap_builder",
  "clap_derive",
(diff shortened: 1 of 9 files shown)


taspelund (approved):
I _love_ the idea of sticking an enum inside an Atomic int... that's genius.

The pre-stored shutdown seems less than ideal, but that's something that can be addressed in a follow-on PR.

nicolaskagami (approved):
LGTM

I had a comment on your previous PR: 
> Some tests (all derivatives of `basic_peering_helper`) check for the state to move to `Idle` before trying to restart the session. Since the state transitions to `Idle` before the the `event_rx` is returned the `fsm_start` could lose the race.
> I see you have more planned changes on top of this, so I'll just point out that we should just care that we gate the state (and its expectation of being able to start) around the lock for the long term.
> _Originally posted by @nicolaskagami in https://github.com/oxidecomputer/maghemite/pull/870#pullrequestreview-4834775239_

In the long run, should the `basic_peering_helper` tests start checking this shutdown state or should we set the `FsmState` after the lock?

sunshowers:
> In the long run, should the basic_peering_helper tests start checking this shutdown state or should we set the FsmState after the lock?

As mentioned to @taspelund yesterday, I would personally advocate for not having shared mutable state at all (which results in all these fragilities) and instead using the actor model to provide a coherent view of state at all times :)

PR #877, opened 2026-08-14
Add ddm_apply API endpoint

taspelund:
Adds a new API endpoint (/ddm/omicron/apply) which provides an
idempotent method for clients to supply a list of front-panel interfaces
on which to run DDM.

Adds GET /interfaces API endpoint to query interfaces known to DDM. This
exposes all configured interfaces, not just ones that have a known peer.

Replace ifname string match with a dpd client query for the "uplink"
link-prop. Route installation into dpd is now skipped when DDM learned
about the route from a peer on a link where "uplink" is false, rather
than when the ifname contains "tfportrear".

Refactors start_state_machines() to be a method of struct HandlerContext
in order to centralize state where it can be mutated by the API handler.

Consolidates HandlerContext.peers and HandlerContext.event_channels
into a single IdOrdMap<SmContext>. `event_channels` was a duplicate
collection holding just mpsc Senders, which were already accessible from
`peers` without even needing to go through a lock first.

Adds unit tests as well as a new falcon-lab topology (ddm-lifecycle) and
an associated test case. This is currently commented out in buildomat,
since the full FSM lifecycle has not yet been implemented (removing an
interface from DDM does not trigger shutdown of all its threads).

Sets SO_REUSEADDR on the unicast DDM socket, aligning with the multicast
socket behavior. On daemon restart it was observed that bind would fail
with EADDRINUSE if the kernel hadn't yet cleared out old socket state.


taspelund:
Note for reviewers: the remove path for FSMs/interfaces is not fully plumbed yet. That will come in a follow-up PR dependent on this one, as it will involve adjusting some of the FSM internals and I want to make sure it's not holding up the use of ddm_apply for multi-rack PoC stuff.

Also, the vast majority of this PR (in terms of LoC) is a new falcon-lab test. The actual changes in ddm/ddmd are a few hundred lines.

ddm-api-types/versions/src/ddm_apply/db.rs
```diff
@@ -0,0 +1,32 @@
...
+/// Information about a DDM interface and its FSM state.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub struct InterfaceInfo {
+    pub name: String,
+    pub addr: Ipv6Addr,
+    pub status: PeerStatus,
+    pub peer_addr: Option<Ipv6Addr>,
+    pub peer_host: Option<String>,
+    pub peer_kind: Option<RouterKind>,
+    pub solicitations_sent: u64,
```
andrewjstone, line 32: Break out stats into their own struct

ddm-api-types/versions/src/ddm_apply/db.rs
```diff
@@ -0,0 +1,32 @@
...
+use serde::{Deserialize, Serialize};
+use std::net::Ipv6Addr;
+
+/// Information about a DDM interface and its FSM state.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub struct InterfaceInfo {
+    pub name: String,
+    pub addr: Ipv6Addr,
+    pub status: PeerStatus,
+    pub peer_addr: Option<Ipv6Addr>,
```
andrewjstone: Create a `Peer` struct.

ddm/src/sm/mod.rs
```diff
@@ -246,15 +250,171 @@ pub struct SmContext {
...
+
+    // A single shared runtime lives for the duration of the test binary.
+    // SmContext holds an Arc<Handle>; without a live runtime the handle would
+    // be invalid, but our tests never drive any async work through it.
+    static RT: OnceLock<tokio::runtime::Runtime> = OnceLock::new();
+
+    fn rt_handle() -> Arc<tokio::runtime::Handle> {
+        let rt = RT.get_or_init(|| tokio::runtime::Runtime::new().unwrap());
+        Arc::new(rt.handle().clone())
+    }
```
cfzimmerman: Why use an `Arc` here? The handle implements `Clone` and already wraps an `Arc`.

https://github.com/tokio-rs/tokio/blob/625954f365727668cb02d04172b34f1149637728/tokio/src/runtime/scheduler/mod.rs#L31-L44

ddm/src/sm/mod.rs
```diff
@@ -246,15 +250,171 @@ pub struct SmContext {
...
+    pub fn interface_info(&self) -> InterfaceInfo {
+        let status = self.iface.peer_status();
+        let peer = lock!(self.iface.peer_identity).clone();
+        InterfaceInfo {
+            name: lock!(self.iface.if_name).clone(),
+            addr: *lock!(self.iface.addr),
+            status,
+            peer_addr: peer.as_ref().map(|p| p.addr),
+            peer_host: peer.as_ref().map(|p| p.hostname.clone()),
+            peer_kind: peer.as_ref().map(|p| p.kind),
```
cfzimmerman: You can skip `clone` if you change the order.

```rust
peer_addr: peer.as_ref().map(|p| p.addr),
peer_kind: peer.as_ref().map(|p| p.kind),
peer_host: peer.map(|p| p.hostname),
```

ddm/src/admin.rs
```diff
@@ -126,18 +189,51 @@ pub enum DdmAdminApiImpl {}
...
+        Ok(HttpResponseUpdatedNoContent())
+    }
+
     async fn get_peers(
         ctx: RequestContext<Self::Context>,
     ) -> Result<HttpResponseOk<HashMap<u32, PeerInfo>>, HttpError> {
         let ctx = lock!(ctx.context());
         let mut result = HashMap::new();
-        for sm in &ctx.peers {
+        read_lock!(ctx.peers).iter().for_each(|sm| {
```
cfzimmerman: Would it feel more natural to use `filter_map`?

```rust
read_lock!(...)
    .iter()
    .filter_map(|sm| {
        ...
        Some((if_index, PeerInfo { ... }))
    })
    .collect()
```

ddm/src/sys.rs
```diff
@@ -235,6 +238,17 @@ pub fn add_routes_dendrite(
...
+///
+/// Tfport interfaces are named `tfport<port>_<link>`, such as
+/// `tfportrear0_0` or `tfportqsfp10_1`.
+fn tfport_to_dpd(ifname: &str) -> Option<(PortId, types::LinkId)> {
+    let (port, link) = ifname.strip_prefix("tfport")?.rsplit_once('_')?;
+    let port_id = port.parse::<PortId>().ok()?;
+    let link_id = types::LinkId(link.parse::<u8>().ok()?);
+    Some((port_id, link_id))
+}
+
```
cfzimmerman, line 250: Is this logic unique to ddm/maghemite? If not, I wonder if it should be exported by something in dendrite. No need for changes in this PR, though.

falcon-lab/src/ddm.rs
```diff
@@ -31,6 +36,29 @@ impl DdmNode {
...
 
+    pub async fn wait_for_api(
+        &self,
+        client: &Client,
+        timeout: Duration,
+        d: &Runner,
+    ) -> Result<()> {
+        let start = Instant::now();
+        loop {
+            match client.get_peers().await {
```
cfzimmerman: Is `get_peers` bounded in duration? If not, is this at risk of waiting beyond `timeout`? Is a `tokio::time::timeout_at` warranted?

cfzimmerman (commented):
I'm mostly just reviewing these to learn the codebase. Looks good!

ddm/src/sys.rs
taspelund: No, it's not. This is another instance of where the `String` type isn't sufficient for carrying an interface name between components, like what came up in our call today.

I'd really prefer to use the upcoming "tfport type", but ultimately that'll depend on how quickly we need to get this landed to unblock multi-rack

falcon-lab/src/ddm.rs
taspelund: Good call. Will revisit the timeouts in the next commit

ddm/src/sm/mod.rs
taspelund: I left the shape of the handle as-is from before this PR. I'll take a look and see if I can yank the Arc wrapper without too big of a blast radius

ddm-api/src/lib.rs
```diff
@@ -46,6 +48,28 @@ api_versions!([
...
+    #[endpoint {
+        method = POST,
+        path = "/ddm/omicron/apply",
+        versions = VERSION_DDM_APPLY..,
+    }]
+    async fn ddm_apply(
+        ctx: RequestContext<Self::Context>,
+        request: TypedBody<v3::config::ApplyRequest>,
+    ) -> Result<HttpResponseUpdatedNoContent, HttpError>;
+
```
internet-diglett, line 77: I feel like some of the useful explanations that were put in the PR description would fit really well here as doc strings, because I had to go back and look at the PR description to know what was going on here.

internet-diglett:
> Note for reviewers: the remove path for FSMs/interfaces is not fully plumbed yet. That will come in a follow-up PR dependent on this one, as it will involve adjusting some of the FSM internals and I want to make sure it's not holding up the use of ddm_apply for multi-rack PoC stuff.

Does this work actually need to be in `main` in order for it to be used in the mutli-rack PoC? I don't want to artificially create blockers, but the lack of cleanup logic seems to create a situation where we're actually leaking threads, and it would be very unfortunate for this to make it onto a production system because we didn't manage to land the followup maghemite PR in time.

ddm-api/src/lib.rs
taspelund: Good call. Didn't think about that when I wrote it... and I forgot to circle back to it lol
I'll fix it here momentarily

taspelund:
> Does this work actually need to be in `main` in order for it to be used in the mutli-rack PoC? I don't want to artificially create blockers, but the lack of cleanup logic seems to create a situation where we're actually leaking threads, and it would be very unfortunate for this to make it onto a production system because we didn't manage to land the followup maghemite PR in time.

I talked with @andrewjstone and he said he's fine with the merge waiting for the teardown fixes. Thank you for the quick review, although it turns out it wasn't so urgent 😅

taspelund:
I've pushed new commits which implement the back half of the FSM lifecycle.

We now thread ownership of FSM threads back out through to the FSM handles, which now have a centralized owner called the `Overseer`. Existing ztest cases pass (`ddm_trio_v3` / `ddm_quartet`), as does the new falcon-lab test (`ddm-trio ddm-apply-lifecycle`).

I will be out on PTO for the next 2 weeks, but I think this is in a good enough spot to be reviewed.

ddm-api-types/versions/src/ddm_apply/config.rs
```diff
@@ -0,0 +1,17 @@
...
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
+pub struct ApplyRequest {
+    /// The complete set of dynamically managed interface names DDM should
+    /// peer over. Dynamic interfaces not in this set are torn down and their
+    /// routes withdrawn. Interfaces supplied on the ddmd command line are
+    /// static and unaffected by this request.
```
andrewjstone, line 15: What interfaces are supplied via the command line typically?

ddm-api-types/versions/src/ddm_apply/config.rs
```diff
@@ -0,0 +1,17 @@
...
+    pub ddm_interfaces: BTreeSet<String>,
```
andrewjstone, line 16: Presumably interface names are not arbitrary. It would be great if we could use a stronger type here and validate that the request parses into valid interface names before we attempt to call apply. In that case we wouldn't attempt to start bogus interfaces.

I traced this call through to `Overseer::add_state_machine` and that call does not ever return an error. It spawns a thread for the state machine that also doesn't appear to ever exit unless stopped explicitly. Does this thread just keep running indefinitely even if the interface doesn't exist? Perhaps we should exit early another way, by checking the interfaces exist when we receive the set of strings, rather than strongly typing them?

ddm-api-types/versions/src/ddm_apply/db.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// How long a DDM interface lives, and therefore who may remove it.
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, JsonSchema,
+)]
+#[serde(rename_all = "snake_case")]
+pub enum InterfaceLifetime {
+    /// Supplied on the ddmd command line. Lives for the life of the daemon;
+    /// `ddm_apply` neither adds nor removes it.
+    Static,
```
andrewjstone, line 31: I mentioned this elsewhere, but I'm unclear why this distinction exists.

ddm-api/src/lib.rs
```diff
@@ -46,6 +48,33 @@ api_versions!([
...
+    /// Declarative endpoint describing the set of interfaces DDM should run on.
+    /// DDM FSMs will be spawned for requested interfaces that are unknown to
+    /// DDM, and stopped for known interfaces that aren't in the request.
+    #[endpoint {
+        method = POST,
+        path = "/ddm/omicron/apply",
+        versions = VERSION_DDM_APPLY..,
+    }]
+    async fn ddm_apply(
```
andrewjstone, line 73: I think the function name and endpoint name are pretty generic and unclear. What if we called this something like `ddm_interfaces` instead?

ddmd/src/main.rs
```diff
@@ -6,30 +6,27 @@ use camino::Utf8PathBuf;
...
 #[derive(Debug, Parser)]
 #[command(version, about, long_about = None, styles = get_styles())]
 struct Arg {
-    /// Address objects to route over.
-    #[arg(short, long = "addr", name = "addr")]
-    addresses: Vec<String>,
+    /// Interfaces to route over. These are static: `ddm_apply` cannot remove
+    /// them.
+    #[arg(short = 'a', long = "addr", name = "addr")]
+    interfaces: Vec<String>,
```
andrewjstone, line 29: This should probably be a `HashSet` or `BTreeSet`. It would also be helpful to indicate in this comment or elsewhere, what interfaces are expected to be static and why.

ddm/src/sm/overseer.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Static machines come from the ddmd command line and live for the life of
+/// the daemon. Dynamic machines are owned by `apply`, which reconciles only
+/// that set, so a client that knows nothing about the static interfaces
+/// cannot strip them. Every other Overseer accessor sees both kinds.
+///
+/// FSM threads never take the Overseer lock. They reach the mesh through the
+/// shared `Arc` in [`super::SmContext::event_channels`], and they are free to
+/// take the mesh read lock, database locks, and the `InterfaceState` mutex
+/// while `apply` is joining them.
+pub struct Overseer {
```
andrewjstone, line 41: I really dislike this name, given the historical connotations. What about `FsmManager`?

ddm/src/sm/overseer.rs
```diff
@@ -0,0 +1,185 @@
...
+            tx: tx.clone(),
+            event_channels: self.mesh.clone(),
+            rt: self.template.rt.clone(),
+            hostname: self.template.hostname.clone(),
+            iface: Arc::new(InterfaceState::default()),
+            stats: Arc::new(SessionStats::default()),
+            log: self.template.log.clone(),
+        };
+
+        write_lock!(self.mesh).insert(ifname, tx);
```
andrewjstone, line 120: We ignore if there is already interface present here. I'm not sure of the ramifications in this case. At a minimum, we'll leak the channels. We assert below that the state machine doesn't exist, so presumably we should do the same here.

ddm/src/sm/state.rs
```diff
@@ -4,118 +4,126 @@
...
+        let thread = spawn(move || {
             let mut state: Box<dyn State> =
-                Box::new(Init::new(ctx.clone(), log.clone()));
-            loop {
-                (state, rx) = state.run(rx);
+                Box::new(Init::new(thread_ctx, log, thread_stop));
+            while let Step::Next(next, event) = state.run(rx) {
+                state = next;
+                rx = event;
             }
```
andrewjstone, line 43: We should log the fact that we stopped this state machine here.

ddm/src/exchange/runtime.rs
```diff
@@ -47,13 +49,36 @@ pub struct HandlerContext {
...
+
+impl ExchangeServer {
+    /// Gracefully close the server and wait for in-flight handlers.
+    pub(crate) fn close(
+        self,
+        rt: &tokio::runtime::Handle,
+        log: &Logger,
+        if_name: &str,
+    ) {
+        if self.exited.load(Ordering::Relaxed) {
```
andrewjstone, line 66: Worth logging this exit?

ddm/src/admin.rs
```diff
@@ -47,10 +51,9 @@ pub struct RouterStats {
 
 #[derive(Clone)]
 pub struct HandlerContext {
-    pub event_channels: Vec<Sender<Event>>,
     pub db: Db,
     pub stats: Arc<RouterStats>,
-    pub peers: Vec<SmContext>,
+    pub overseer: Arc<RwLock<Overseer>>,
```
andrewjstone, line 56: This change forces every HTTP handler task to take a lock now. The call to `apply` can take a while, and this forces every handler task to be blocked behind it. 

Found with help from Claude: 
> every other handler — even ones that never touch the
  overseer, like get_originated — piles up on the mutex, each pinning a worker.

ddm/src/admin.rs
```diff
@@ -126,30 +129,70 @@ pub enum DdmAdminApiImpl {}
...
+            // apply() joins every departing FSM thread, and FSM teardown
+            // re-enters this runtime via Handle::block_on (closing the
+            // exchange server, deleting routes through DPD). Those futures
+            // need a free worker to make progress. If the join itself sat on
+            // a worker, teardown could starve the runtime and deadlock while
+            // holding the Overseer write lock.
+            tokio::task::spawn_blocking(move || {
+                write_lock!(overseer).apply(desired)
+            })
+            .await
```
nicolaskagami, line 166: Putting it in a `spawn_blocking` is not enough to avoid worker starvation.

There are other consumers (e.g. in `admin.rs`) that would block their tokio workers waiting for this, which is itself `join()`s the departing threads, which blocks on `route_ipv6_delete`, which needs active workers to advance. This means that we can deadlock with enough concomitant of these consumers, and basically one is spawned for every API call in `admin.rs`.

These other threads also wait on `self.mesh`, which is currently only held for brief periods but I think this imposes a difficult variant to maintain, even if the starvation problem is solved.

I think both lock should have a more streamlined usage pattern.

ddm-api-types/versions/src/ddm_apply/config.rs
taspelund: All backplane links are supplied here via SMF

ddm-api-types/versions/src/ddm_apply/db.rs
taspelund: The distinction here is so that omicron cannot turn on/off interfaces that were supplied by SMF.
i.e.
SMF is static (backplane links), omicron is dynamic (QSFP ports).

The nomenclature was more of a way to reference this without directly naming something QSFP or Front Panel, in case dynamic management is later needed for interface types that are yet unknown

ddm/src/sm/overseer.rs
taspelund: I wasn't aware of the connotations there, that's my bad. I literally pulled out this name because I'd been watching Fallout

ddm-api/src/lib.rs
taspelund: I used the name `ddm_apply` because it seemed like a parallel to mgd's `bgp_apply` which is the main API used by omicron. I'm not married to the name though, so I'm open to suggestions

andrewjstone: Makes sense. I think the same of `bgp_apply` now too lol.

ddm/src/sm/overseer.rs
andrewjstone: Yeah, I didn't think it was intentional. I just noticed it right away and thought it would be good to change before merge.

ddm-api-types/versions/src/ddm_apply/config.rs
andrewjstone: Ah that makes sense.

ddm-api-types/versions/src/ddm_apply/db.rs
andrewjstone: Makes sense.

ddm/src/admin.rs
rcgoodfellow: I think a way to concisely state the pattern that must be followed today is this:

_A synchronous blocking operation that blocks other asynchronous runtime threads (e.g. tokio workers) may not depend on asynchronous await operations to become unblocked._

As you state @nicolaskagami, this is not a great place to be. It requires in-depth analysis to maintain and it's rather easy to violate.

nicolaskagami: That sums it up very well, given our current codebase patterns. 

However, it should be noted that the "synchronous blocking operation" is actually "synchronous blocking in an asynchronous context", meaning that its async worker will be blocked and therefore unable to work on other asynchronous things like the ones this task could otherwise depend on.

PR #881, opened 2026-08-17
lab: add error wrapper to fix build warnings

taspelund:
Adds a pointer-sized wrapper around libfalcon::error::Error. This avoids using large amount of stack memory for large error variants, e.g. libfalcon::error::Error::Propolis, which clippy complains about because it's 176+ bytes. Wrapping it in a `Box` keeps the stack cost to just one ptr while preserving the full error chain via #[error(transparent)].

### DIFF
--- a/lab/src/error.rs
+++ b/lab/src/error.rs
@@ -1,0 +1,15 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Copyright 2026 Oxide Computer Company
+
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
+pub struct Error(Box<libfalcon::error::Error>);
+
+impl From<libfalcon::error::Error> for Error {
+    fn from(e: libfalcon::error::Error) -> Self {
+        Error(Box::new(e))
+    }
+}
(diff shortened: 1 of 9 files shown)


lab/src/error.rs
```diff
@@ -0,0 +1,15 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Copyright 2026 Oxide Computer Company
+
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
+pub struct Error(Box<libfalcon::error::Error>);
```
internet-diglett, line 9: So I went and took a look at this Error type and yeah, we have _a lot_ going on in there. This seems to be a reasonable approach.

internet-diglett (approved):

nicolaskagami (approved):
LGTM

Two thoughts:
- Should we consider doing this in `libfalcon`? Is this the only place where this is bothersome?
- It's a bit difficult to get the underlying type now. Is there a reason against exposing it (as `pub`, implementing `Deref`, etc)?

taspelund:
>     * Should we consider doing this in `libfalcon`? Is this the only place where this is bothersome?

Yeah, I think we probably should fix it there as well

>     * It's a bit difficult to get the underlying type now. Is there a reason against exposing it (as `pub`, implementing `Deref`, etc)?

No reason. I'll push a new commit

nicolaskagami (approved):

PR #882, opened 2026-08-17
buildomat: delete interop jobs

taspelund:
These jobs were used as the basis of automated interop testing for mgd. However, since falcon-lab was introduced we've stopped using them. Let's delete these jobs in favor of just going with falcon-lab.


nicolaskagami (commented):
Seems fine to me but I do wonder:

- Are they becoming stale?
- How does falcon-lab supplant them?

taspelund:
> Seems fine to me but I do wonder:
> 
>     * Are they becoming stale?

I'm not understanding what you're asking. Can you rephrase?

>     * How does falcon-lab supplant them?

falcon-lab has topologies for BGP unnumbered and BFD w/ static routing, both of which have accompanying tests. These are run by CI as part of the `falcon` job (the actual test is in `.github/buildomat/jobs/falcon-lab.sh`).

Prior to falcon-lab, we used the interop crate in the testbed repo for these kind of interop tests, and those network topologies were copied over to falcon-lab as an in-repo replacement

nicolaskagami (commented):
> I'm not understanding what you're asking. Can you rephrase?

I was mostly trying to understand if the test had diverged from the code or its intended purpose.

taspelund:
Ah, I see. I wouldn't say it's diverged so much as it's stopped being maintained. I don't think it's kept up with changes to RSS or API methods, particularly because it's annoying to maintain it as part of a separate repo

nicolaskagami (approved):

rcgoodfellow (approved):

PR #884, opened 2026-08-20
bfd: actually insert peers into persistent db

taspelund:
BFD peers were not being inserted into the sled db, which means they
 weren't recoverable on daemon restart/crash.

 This moves all of the BFD API handlers' core logic into methods of
 BfdContext, and makes BfdContext.daemon private. With BfdContext being
 the sole owner of its inner state, all locking invariants can be upheld
 without requiring any participation from its callers.

 API operations now lock the daemon slightly earlier, ensuring updates to
 both the DB and the runtime state happen during the critical section of
 the mutex. This prevents concurrent API Add/Del requests from partially
 trampling over each other by enforcing serialization.

 Peer add requests now implement a rollback function upon runtime errors.
 e.g.
 1) Add peer to DB (success)
 2) Add peer to daemon (failure)
 3) Rollback DB entry

 Rollback failures return a 500 error.

 before:
 ```
 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 8e0f (empty) (no description)
 › ./target/debug/mgadm bfd add-peer 100.64.0.0 0.0.0.0 1000 3 single-hop

 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 8e0f (empty) (no description)
 › ./target/debug/mgadm bfd get-peers
 Peer        Listen   Required Rx  Detection Threshold  Mode       Status
 100.64.0.0  0.0.0.0  1000         3                    SingleHop  Down

 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 8e0f (empty) (no description)
 › pkill mgd
 pkill: signalling pid 33096: Operation not permitted

 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 8e0f (empty) (no description) ✕
 › sudo !!

 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 8e0f (empty) (no description) ✕
 › sudo pkill mgd

 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 8e0f (empty) (no description)
 › pkill mgd
 pkill: signalling pid 35931: Operation not permitted

 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 8e0f (empty) (no description) ✕
 › ./target/debug/mgadm bfd get-peers
 Peer  Listen  Required Rx  Detection Threshold  Mode  Status

 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 8e0f (empty) (no description)
 ›
 ```

 after:
 ```
 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 1651 bfd: actually insert peers into persistent db ±1 +15 -9
 › ./target/debug/mgadm bfd get-peers
 Peer  Listen  Required Rx  Detection Threshold  Mode  Status

 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 1651 bfd: actually insert peers into persistent db ±1 +15 -9
 › ./target/debug/mgadm bfd add-peer 100.64.0.0 0.0.0.0 1000 3 single-hop

 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 1651 bfd: actually insert peers into persistent db ±1 +15 -9
 › pgrep mgd
 42931

 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 1651 bfd: actually insert peers into persistent db ±1 +15 -9
 › sudo pkill mgd

 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 1651 bfd: actually insert peers into persistent db ±1 +15 -9
 › pgrep mgd
 43407

 treyaspelund@Tallon-IV …/oxidecomputer/maghemite/bfd-db xsml 1651 bfd: actually insert peers into persistent db ±1 +15 -9
 › ./target/debug/mgadm bfd get-peers
 Peer        Listen   Required Rx  Detection Threshold  Mode       Status
 100.64.0.0  0.0.0.0  1000         3                    SingleHop  Down
 ```

### DIFF
--- a/mgd/src/bfd_admin.rs
+++ b/mgd/src/bfd_admin.rs
@@ -6,10 +6,9 @@ use anyhow::Result;
 use bfd::Daemon;
 use bfd::Session;
 use bfd::SessionCounters;
-use dropshot::{HttpError, HttpResponseUpdatedNoContent, TypedBody};
+use dropshot::HttpError;
 use mg_api_types::bfd::BfdPeerConfig;
 use mg_api_types::bfd::BfdPeerInfo;
-use mg_api_types::bfd::DeleteBfdPeerPathParams;
 use mg_common::lock;
 use slog::Logger;
 use slog_error_chain::InlineErrorChain;
@@ -27,6 +26,13 @@ impl BfdContext {
     pub fn new(log: Logger) -> Self {
         Self {
             daemon: Arc::new(Mutex::new(Daemon::new(log.clone()))),
+        }
+    }
+
+    #[cfg(test)]
+    fn new_for_test(log: Logger) -> Self {
+        Self {
+            daemon: Arc::new(Mutex::new(Daemon::new_for_test(log))),
         }
     }
 
@@ -74,7 +80,7 @@ impl BfdContext {
     ///
     /// This reads the peer configs from the DB itself so callers cannot start
     /// an unpersisted peer through the production API.
-    pub fn restore_peers(
+    pub fn restore_db_peers(
         &self,
         db: rdb::Db,
     ) -> Result<(), crate::error::Error> {
@@ -94,10 +100,8 @@ impl BfdContext {
     pub fn add_new_peer(
         &self,
         db: rdb::Db,
-        request: TypedBody<BfdPeerConfig>,
-    ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
-        let rq = request.into_inner();
-
+        rq: BfdPeerConfig,
+    ) -> Result<(), HttpError> {
         let mut daemon = lock!(self.daemon);
 
         db.clone()
@@ -113,7 +117,7 @@ impl BfdContext {
             return Err(crate::error::Error::Bfd(e).into());
         }
 
-        Ok(HttpResponseUpdatedNoContent())
+        Ok(())
     }
 
     /// Remove a BFD peer from the persistent DB and the running daemon.
@@ -121,16 +125,16 @@ impl BfdContext {
     pub async fn remove_peer(
         &self,
         db: rdb::Db,
-        peer: DeleteBfdPeerPathParams,
-    ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
+        peer: IpAddr,
+    ) -> Result<(), HttpError> {
         let handle = {
             let mut daemon = lock!(self.daemon);
-            db.remove_bfd_neighbor(peer.addr).map_err(|e| {
+            db.remove_bfd_neighbor(peer).map_err(|e| {
                 HttpError::for_internal_error(
                     InlineErrorChain::new(&e).to_string(),
                 )
             })?;
-            daemon.remove_peer(peer.addr)
+            daemon.remove_peer(peer)
         };
 
         if let Some(handle) = handle {
@@ -147,7 +151,7 @@ impl BfdContext {
             handle.shutdown().await;
         }
 
-        Ok(HttpResponseUpdatedNoContent {})
+        Ok(())
     }
 
     pub fn session_counters(&self) -> Vec<(IpAddr, Arc<SessionCounters>)> {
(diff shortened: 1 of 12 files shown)


mgd/src/bfd_admin.rs
```diff
@@ -31,95 +28,274 @@ impl BfdContext {
...
-            ),
-            AddPeerError::Bind { .. }
-            | AddPeerError::SetSocketNonBlocking(_)
-            | AddPeerError::StdToTokio(_) => HttpError::for_internal_error(
-                InlineErrorChain::new(&err).to_string(),
-            ),
-        })
+    /// Restore persisted BFD peers to the running daemon.
+    ///
+    /// This reads the peer configs from the DB itself so callers cannot start
```
rcgoodfellow, line 81: I'm not really following this comment. I get why we want to be able to restore peers from the db. But I don't see how ensuring callers cannot start an unpersisted peer directly follows from that. I would think ensuring persistence is a property of the functions that create peers, not the db restoration function.

mgd/src/bfd_admin.rs
```diff
@@ -31,95 +28,274 @@ impl BfdContext {
...
+    /// This first tries to update the persistent DB with the BFD peer config,
+    /// starting its runtime only upon success; if the DB update fails, an error
+    /// is returned. A rollback of the DB update is attempted if the subsequent
+    /// runtime update fails, returning an internal error upon rollback failure.
+    pub fn add_new_peer(
+        &self,
+        db: rdb::Db,
+        rq: BfdPeerConfig,
+    ) -> Result<(), HttpError> {
+        let mut daemon = lock!(self.daemon);
```
rcgoodfellow, line 105: The PR description says this lock is moved up to have the db write and the act of adding the peer to the daemon be in the same critical region. Should we just move the db update into `daemon.add_peer` so that would be naturally enforced since the daemon needs to be locked to make that call? When I look at this code without the PR description context, my immediate thought is that the daemon lock is being held too long and that line 111 should be `lock!(self.daemon).add_peer(...` instead of the lock up top. It's easy to see that optimization being made later on and unwinding the intent here.

The db could call could be transformed into more of an ensure than an add within `daemon.add_peer` to make the db part idempotent.

mgd/src/bfd_admin.rs
```diff
@@ -31,95 +28,274 @@ impl BfdContext {
...
+
+    /// Remove a BFD peer from the persistent DB and the running daemon.
+    /// The associated peer session will be stopped immediately.
+    pub async fn remove_peer(
+        &self,
+        db: rdb::Db,
+        peer: IpAddr,
+    ) -> Result<(), HttpError> {
+        let handle = {
+            let mut daemon = lock!(self.daemon);
```
rcgoodfellow, line 131: Same comment as above on the eager lock here. Consider moving the db operations into `daemon.remove_peer` instead.

mgd/src/bfd_admin.rs
```diff
@@ -31,95 +28,274 @@ impl BfdContext {
...
+            db.remove_bfd_neighbor(peer).map_err(|e| {
+                HttpError::for_internal_error(
+                    InlineErrorChain::new(&e).to_string(),
+                )
+            })?;
+            daemon.remove_peer(peer)
+        };
+
+        if let Some(handle) = handle {
```
rcgoodfellow, line 140: What is the drawback of doing this while holding the lock? Potentially waiting forever on the listen task to abort?

mgd/src/error.rs
```diff
@@ -47,6 +50,20 @@ impl From<Error> for HttpError {
...
             Error::NotFound(_) => Self::for_not_found(None, value.to_string()),
+            Error::Bfd(ref err) => match err {
+                bfd::AddPeerError::PeerExists(ip_addr) => {
+                    Self::for_client_error(
+                        Some(err.to_string()),
+                        ClientErrorStatusCode::CONFLICT,
+                        format!("bfd peer {ip_addr} already exists"),
+                    )
+                }
+                bfd::AddPeerError::Bind { .. }
```
rcgoodfellow, line 61: A bind error may not always be an internal error, we could have been handed a futile bind address through the API?

rdb/src/db.rs
```diff
@@ -471,6 +471,16 @@ impl Db {
         let tree = self.persistent.open_tree(BFD_NEIGHBOR)?;
         let key = cfg.peer.to_string();
         let value = serde_json::to_string(&cfg)?;
+        match tree.contains_key(key.as_str()) {
```
rcgoodfellow, line 474: Related to the comment above about moving db operations into `Damon::add_peer`. If we do that we may want this to be more of an idempotent/ensure type of call rather than returning an error when the entry already exists.

rcgoodfellow (commented):
Thanks @taspelund. Comments follow, but first a higher level thought.

Now that BFD is being managed by a reconciler in Omicron, and early networking is also a reconciler, do we even need a persistence layer  for BFD at all? Or even for mgd as a whole at all?

taspelund:
@rcgoodfellow are we going to close this and instead go about yanking out sled in future work? If so, do we care if BFD's interactions with sled are incomplete in the meantime?

PR #898, opened 2026-09-01
falcon-lab: separate topologies from scenarios

taspelund:
Refactor falcon-lab to have separate abstractions for Topologies and Scenarios. A Topology is a set of nodes, their backend (VM image), and the links interconnecting them. A Scenario is layered atop a Topology, describing the configuration, diagnostics, and test cases being applied.

The old run targets now look like this:
- mgd-unnumbered -> (mgd-duo, bgp-unnumbered)
- quartet-unnumbered -> (interop, bgp-unnumbered)
- quartet-bfd-static-routing -> (interop, bfd-static-routing)

### DIFF
--- a/falcon-lab/src/main.rs
+++ b/falcon-lab/src/main.rs
@@ -1,12 +1,10 @@
 //! Falcon test lab
 
 use crate::dendrite::NpuvmCommits;
-use crate::test::{
-    cleanup_mgd_unnumbered_test, cleanup_quartet_bfd_static_test,
-    cleanup_quartet_unnumbered_test, run_mgd_unnumbered_test,
-    run_quartet_bfd_static_test, run_quartet_unnumbered_test,
+use crate::scenario::{
+    InteropScenario, MgdDuoScenario, Scenario, ScenarioOptions,
 };
-use clap::{Parser, Subcommand};
+use clap::{Args, Parser, Subcommand, ValueEnum};
 
 mod bgp;
 mod ddm;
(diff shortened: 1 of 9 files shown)


falcon-lab/src/main.rs
```diff
@@ -128,3 +145,66 @@ async fn run() -> anyhow::Result<()> {
...
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn run_options_follow_scenario() {
+        let cli = Cli::try_parse_from([
```
cfzimmerman, line 155: cool, TIL you can test clap like this.

falcon-lab/src/scenario.rs
```diff
@@ -42,46 +43,220 @@ use std::{
...
+        cleanup_interop_deployment(self.deployment())
+    }
+}
+
+#[cfg(test)]
+mod deployment_tests {
+    use super::Deployment;
+
+    #[test]
+    fn deployment_names_are_falcon_compatible() {
```
cfzimmerman, line 195: In general, it would be cool if we could use types to make this test unnecessary. Although fair if that seems unnecessary effort/complexity.

But if falcon imposes constraints like this, it might behoove us to put that logic in falcon, document why it's imposed, and export it for clients to use.

falcon-lab/src/scenario.rs
```diff
@@ -167,22 +353,25 @@ where
...
         if diag_on_fail {
-            collect_diagnostics(&ad, ox, cr1, cr2, cr3, &topo_name, protocols)
-                .await;
+            collect_interop_diagnostics(
+                &ad, ox, peer, routers, &topo_name, protocols,
+            )
+            .await;
             ox.collect_ndp_diagnostics(&ad, &mgd, &topo_name).await;
+            peer.collect_ndp_diagnostics(&ad, &peer_mgd, &topo_name)
+                .await;
```
cfzimmerman, line 364: If performance warrants complexity, this seems like a candidate for `tokio::join!`

https://docs.rs/tokio/latest/tokio/macro.join.html

falcon-lab/src/scenario.rs
```diff
@@ -558,61 +773,87 @@ pub async fn run_quartet_unnumbered_test(
...
+        "tfportqsfp1_0",
+        "tfportqsfp2_0",
+        "tfportqsfp3_0",
+    ] {
         let addr = format!("{link}/ll");
         ox.illumos()
             .addrconf(&ad, &addr)
             .await
             .context(format!("create {addr}"))?;
     }
```
cfzimmerman, line 806: If immediately calling `format!` with a static suffix, why not just hard code "/ll"? ex. `"tfportqsfp0_0/ll"`

falcon-lab/src/scenario.rs
cfzimmerman, line 1103: Again if these have any performance sensitivity and no ordering constraint, then `tokio::join!` might be attractive.

cfzimmerman (approved):
Looks like a nice improvement, but I don't have the context to make a full judgement on design. That said, if a CI test PR passes CI tests, that seems positive.

falcon-lab/src/scenario.rs
taspelund: I have another PR in the stack that adds a lot of parallelism using tokio join. I think I hit these spots, but I'll double check and update it if not

falcon-lab/src/scenario.rs
taspelund: Agreed. I replied to your other comment with details

falcon-lab/src/scenario.rs
taspelund: Yeah, this is resolved in #901

falcon-lab/src/scenario.rs
taspelund: Resolved in #901

falcon-lab/src/scenario.rs
taspelund: done.

falcon-lab/src/scenario.rs
taspelund: Yeah, let me double check. I thought this was a limitation stemming from falcon, but now I'm questioning my recollection :D

nicolaskagami (approved):
Very nice

✅ Improves our abstractions
✅ Clear intent
✅ Well compartmentalized changes

I think there's some opportunity for simplification in this code (e.g. `RunTopology`/`CleanupTopology` could be generics, and perhaps  we could lose boilerplate matching scenarios to topologies as well), but it's fine the way it is.

falcon-lab/src/scenario.rs
taspelund: Yeah, the limitation is in illumos, which libfalcon is sort of the shepherd of for these kinds of setups... so I think the type should be added to libfalcon and then trickled down to falcon-lab.  I'll make a note to submit a PR to libfalcon to use better types + a separate follow-on to fixup maghemite to use the better types.

taspelund:
## Revision history

| Version | Changes from previous version | Submitted commit |
| --- | --- | --- |
| 2 (current) | [`a0af47b3..1cca6304`](https://github.com/oxidecomputer/maghemite/compare/a0af47b3f66fa7a3566fa113181cd695450a0bcd..1cca63044937be8558e2ea777c5f75ebdb24e111) | [`1cca6304`](https://github.com/oxidecomputer/maghemite/commit/1cca63044937be8558e2ea777c5f75ebdb24e111) |
| 1 | Initial version | [`a0af47b3`](https://github.com/oxidecomputer/maghemite/commit/a0af47b3f66fa7a3566fa113181cd695450a0bcd) |

<sub>Generated by jj-stack from this pull request's force-push history.</sub>

falcon-lab/src/scenario.rs
taspelund: Just pushed an update via jj-stack that replaces this test with a compile-time assertion. This is meant as a stop-gap between now and when the libfalcon changes are made + incorporated into falcon-lab.

taspelund:
> Very nice
> 
> ✅ Improves our abstractions ✅ Clear intent ✅ Well compartmentalized changes
> 
> I think there's some opportunity for simplification in this code (e.g. `RunTopology`/`CleanupTopology` could be generics, and perhaps we could lose boilerplate matching scenarios to topologies as well), but it's fine the way it is.

Thanks! I took your advice and made things more generic, which even resulted in the falcon-lab `Deployment` type getting removed. So things are definitely looking nicer!

cfzimmerman (approved):

PR #911, opened 2026-09-11
dynamic external ddm peers

rcgoodfellow:
## Background

This PR adds support for dynamic external peers. Up until now ddm peers have been created for every backplane port on an Oxide rack. That meant that at startup one peer state machine is spun up for each physical port and lasts the lifetime of the `ddmd` process. Additional peer state machines are not created beyond that point, and existing state machines are not torn down.

With the introduction of regional cluster underlay networks, there is a need for DDM peerings to take place over external ports to extend the underlay network across racks. The lifetime of an external peering is a function of operator configuration and not of rack construction like the internal backplane peerings.

PR #877 was a first swing at accomplishing dynamic external peers. That PR is build around a design with a centralized overseer held behind a synchronous lock where 1) then endpoint that modifies peers takes an overseer write lock, 2) many other DDM admin API operations require a read lock on the overseer, and 3) release of the overseer lock depends on asynchronous runtime tasks making progress ([1](https://github.com/oxidecomputer/maghemite/pull/877#discussion_r3983402397), [2](https://github.com/oxidecomputer/maghemite/pull/877#discussion_r3983431241)). This presents the opportunity for ddm to become deadlocked.

This PR takes a bunch of cues from #877, but takes a more decentralized approach to external peers. The admin API handler is an orchestrator rather than a resource owner. External peer threads are launched by admin API handlers. Shutdown is done through the event bus. Peers also announce themselves to their siblings through the event bus. No new global lock is required for this approach. However it does require careful ordering of operations for things like peer expiration to close race windows with outstanding threads that are shutting down like the initial pull thread of the exchange state.

### New API Endpoints

Addition of symmetric `get_external_peers` and `set_external_peers` API endpoints. These endpoints provide an interface for the control plane to manage external peerings. The `set_external_peers` endpoint is a total specification. Any active peerings not present in a request will be removed and any peerings present in the specification that do not exist in `ddmd` will be created.

### Implementation

Adding dynamic external peers has some basic underlying machinery needs.

- In-memory bookkeeping to keep track of what external peers are currently running. This is not maintained persistently by `ddmd` as the control plane should reconciling the single persistent source of truth to the daemon.
- A mechanism to start and stop external peers sessions.
- Ensuring that dynamic peers are included in stats collection.
- Dynamic peers are specifically for transit-to-transit router sessions, so we need path vector loop breaking.
- Transit-to-transit redistribution.
- Peer state machines that can deal with sibling peer state machines disappearing.

#### Starting New Peers

The start mechanism for dynamic peers mirrors how static internal peers are started. A dynamic peer is started with default parameters on a specified illumos address object. We may want to extend the external peers API to allow for session parameterization, but that will not be included in this MVP PR.

The set of event channels for communicating with other peer sessions is cloned from the handler context and the new peers' event channel transmitter is added to the handler context event channel. When the peer reaches the established state it broadcasts a `AdminEvent::NewExternalPeer(Sender<Event>)` message to all sibling peers. This is a new admin event introduced in this PR. It's a way for a new dynamic peer to announce itself to its siblings  and provide them with an event transmitter that can be used to send it peering events.

When a resident peer receives a `NewExternalPeer(Sender<Event>)` event, they attempt to redistribute their routes to the new peer through the sender provided. If that is successful they add the sender to their list of event channels. The only way this can fail is if the new external peer has been destroyed by the time the resident peer attempts to redistribute through it (this can happen if the peer is destroyed shortly after creation which is very possible under automation). In this situation, the new peer's event sender is not added to the resident peer's sibling peer event channel set.

#### Removing Peers

Removing peers takes place through event channels. When a `set_external_peers` request comes in that removes a peer, the API handler which maintains an event sender for each peer sends a `AdminEvent::Shutdown` event to the running state machine for the peer. That causes the peer to halt the state machine and return. If the peer is in the established state, it uses the existing `expire_peer` mechanism to notify existing peers that its session is going down and that the appropriate withdraws need to be made in sibling sessions.

The API handler also removes the peer form the in-memory database and handler context.

#### Transit-transit redistribution

Until now transit-transit redistribution has been stubbed out. We were notably missing path breaking which has been added in this PR.

## Testing

A new zone-based falcon test environment has been added to `tests/src/ddm.rs` called `test_external_peer_sextet()`. This is a two sidecar/scrimlet four sled testing environment. A number of changes needed to be made to the common machinery for setting up zone tests to accommodate two sidecar/scrimlet nodes as much of the infrastructure assumed a single sidecar/scrimlet node.

The new tests themselves are run in `run_sextet_tests`.  They set up the two-switch/four-sled topology with a set of prefix announcements coming from the sled routers. The test then proceed to add and remove external peers, making sure that the observable result through the API is what we expect in terms of how many peerings each router has, and what prefixes each router has. This is done initially for a few specific cases. Then a test suite runs that iterates through 100 random peer changes and validates that the result is correct according to expectation functions computed over the set of external peers that was just given to `set_external_peers`.


.github/buildomat/jobs/test-ddm-sextet.sh
```diff
@@ -0,0 +1,18 @@
...
+#:   "/work/*.log",
+#: ]
+
+source .github/buildomat/test-ddm-common.sh
+
+#
+# trio tests
+#
+
+banner "trio"
```
nicolaskagami: wasn't it supposed to be:
```suggestion
#
# sextet tests
#

banner "sextet"
```

ddm-api-types/versions/src/external_peers/external_peers.rs
```diff
@@ -0,0 +1,12 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeSet;
+
+#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema, PartialEq, Eq)]
+pub struct ExternalPeers {
+    pub address_objects: BTreeSet<String>,
```
nicolaskagami, line 11: We might want to enforce some structure onto these `Strings` at some point.

ddm/src/sm/mod.rs
```diff
@@ -38,27 +41,50 @@ pub enum AdminEvent {
...
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum PeerEvent {
+    /// Upon reception of this event, a state machine is to push the update to
+    /// its peer.
     Push(ddm_protocol::v3::Update),
+
+    /// Upon reception of this event, a state machine is to redistribute the
+    /// update to it's sibling routers through it's event channels. The state
+    /// machine is responsible maintaining the path vector and performing loop
```
nicolaskagami:
```suggestion
    /// machine is responsible for maintaining the path vector and performing loop
```

ddm/src/sm/mod.rs
```diff
@@ -38,27 +41,50 @@ pub enum AdminEvent {
 
     /// Synchronize with active peers by pulling their prefixes.
     Sync,
+
+    /// A new external peer has been added to the router that can be reached
+    /// using the provided sender.
+    NewExternalPeer(Sender<Event>),
+
+    /// Shutdown on recipt of this event.
```
nicolaskagami:
```suggestion
    /// Shutdown on receipt of this event.
```
Still sounds a bit weird to me but I'm not a native english speaker.

tests/src/ddm.rs
```diff
@@ -802,6 +862,593 @@ async fn run_quartet_tests(
...
+}
+
+async fn run_sextet_tests(
+    _zs1: &RouterZone<'_>,
+    _zs2: &RouterZone<'_>,
+    _zs3: &RouterZone<'_>,
+    _zs4: &RouterZone<'_>,
+    _zt1: &RouterZone<'_>,
+    _zt2: &RouterZone<'_>,
+) -> Result<()> {
```
nicolaskagami, line 1062: It's the first I'm seeing of this but surely there's a cleaner way to run these tests (not concerning for this PR).

ddm/src/exchange/runtime.rs
```diff
@@ -365,9 +361,10 @@ async fn push_handler_common(
     ctx: RequestContext<Arc<Mutex<HandlerContext>>>,
     update: v3::Update,
 ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
-    let ctx = ctx.context().lock().await.clone();
+    let ctx = lock!(ctx.context()).clone();
+
```
nicolaskagami, line 364: Why are we changing from an async lock here?

ddm/src/sm/mod.rs
```diff
@@ -252,9 +286,38 @@ pub struct SmContext {
...
+pub(crate) fn send(e: Event, event_channels: &mut Vec<Sender<Event>>) {
+    // Ensure our indices are unique and ordered.
+    let mut dead_channels = BTreeSet::default();
+    for (i, c) in event_channels.iter().enumerate() {
+        if c.send(e.clone()).is_err() {
+            dead_channels.insert(i);
+        }
+    }
+    // we need to remove in descending order, so we don't remove `i` and then
+    // try to remove `i+1` later which wlll be a _diffrent_ item than we grabbed
```
nicolaskagami:
```suggestion
    // try to remove `i+1` later which wlll be a _different_ item than we grabbed
```

ddm/src/sm/state.rs
```diff
@@ -335,6 +393,24 @@ impl Exchange {
...
+            let imported = self.ctx.db.imported();
+            dbg!(self.log, self.ctx.config.if_name, "imported: {imported:#?}");
+            dbg!(
+                self.log,
+                self.ctx.config.if_name,
+                "to_remove: {to_remove:#?}"
+            );
+            to_remove.retain(|x| {
+                !imported.iter().any(|y| y.destination == x.destination)
+            });
```
nicolaskagami, line 409: Since this is path vector, shouldn't we match on the path as well?
```suggestion
            to_remove.retain(|x| {
                !imported
                    .iter()
                    .any(|y| y.destination == x.destination && y.path == x.path)
            });
```

ddm/src/admin.rs
```diff
@@ -429,6 +476,129 @@ impl DdmAdminApi for DdmAdminApiImpl {
 
         Ok(HttpResponseUpdatedNoContent())
     }
+
+    async fn set_external_peers(
+        ctx: RequestContext<Self::Context>,
+        request: TypedBody<ExternalPeers>,
+    ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
```
nicolaskagami, line 487: Since `set_external_peers` doesn't wait for all the SMs to expire their peers (which was one of the links in the deadlock on #877), what is preventing a race condition for a quick succession of `set_external_peers` where the first one deletes and the second recreates?

I'm not sure exactly what we're achieving with our current context lock. I take it we want to guarantee that each API call is serialized but in this case that does not extend to their ramifications (e.g. on the state machines). I think that, for it to be consistent/useful, we'd have to actually wait for the API intent to be fully applied, which has its own difficulties as we've seen on #877.

I think the long term fix would have to be a refactor where we'd move from high-level C-style mutexes to the idiomatic Rust style of guarding mutability wherever it's neeeded. Changes should be much easier to reason about after the initial time investment.

A more immediate fix (for this race) would be to either lock SMs by their address object or associate the `Routes` with the owning SM so they don't clash.

It may also be that the race is exceedingly rare and we may not care about it at this moment. We could probably adapt the sextet test to exercise this.

nicolaskagami (commented):
Doing a first pass

Shouldn't there be a `tests/conf/dpd-ports-sextet.toml`?

ddm/src/admin.rs
rcgoodfellow: > what is preventing a race condition for a quick succession of set_external_peers where the first one deletes and the second recreates?

The second will be stuck in discovery until it can bind to the unicast listening address. The only way it can bind is if the first lets go of its unicast socket and the operating system's `TIME_WAIT` expires.

ddm-api/src/lib.rs
```diff
@@ -170,4 +171,23 @@ pub trait DdmAdminApi {
     async fn disable_stats(
         ctx: RequestContext<Self::Context>,
     ) -> Result<HttpResponseUpdatedNoContent, HttpError>;
+
+    #[endpoint {
+        method = POST,
```
andrewjstone: Should this be a `PUT` since it's idempotent?

rcgoodfellow:
> Shouldn't there be a tests/conf/dpd-ports-sextet.toml?

No. That's an artifact of how dpd/sofnpu used to be set up. I can get rid of that across the board for all test topologies across the board I believe.

ddm-api-types/versions/src/external_peers/external_peers.rs
rcgoodfellow: Yeah. Going to defer that to a follow on.

tests/src/ddm.rs
rcgoodfellow: Yeah, these are here because they do cleanup on drop. There is certainly a better way to organize these tests.

ddm/src/sm/state.rs
rcgoodfellow: Yeah, i think that's right. The sextet test case does not cover this since the pair of transit peer links are between the same two routers. But in a case where we have distinct transit paths, the path should be taken into account, and this will work with this code on the receiving end of the withdraw here:

https://github.com/oxidecomputer/maghemite/blob/373e5275c7656734318c6b34215eee637b48b695/ddm/src/exchange/runtime.rs#L681-L700

Added the path check in the latest commit.

ddm/src/sm/mod.rs
rcgoodfellow: Fixed

ddm/src/exchange/runtime.rs
rcgoodfellow: Good catch. This is a leftover from an earlier version when i was modifying the event channel fanout from the exchange runtime handler. But I moved that state mutation into the state machine through the `PeerEvent::Redistribute` event, so this is no longer needed.

I've reverted this back to the async lock.

ddm/src/sm/mod.rs
rcgoodfellow: Fixed

ddm/src/sm/mod.rs
rcgoodfellow: Fixed

.github/buildomat/jobs/test-ddm-sextet.sh
rcgoodfellow: Fixed

ddm-api-types/versions/src/external_peers/external_peers.rs
rcgoodfellow: - https://github.com/oxidecomputer/maghemite/issues/914

ddm-api/src/lib.rs
rcgoodfellow: yep, changed to `PUT`

ddm/src/admin.rs
```diff
@@ -47,14 +57,53 @@ pub struct RouterStats {
...
+    pub tunables: Tunables,
+    pub router_kind: RouterKind,
+    pub rack_id: Option<Uuid>,
+    pub sled_id: Option<Uuid>,
+    pub router_id: String,
     pub log: Logger,
 }
 
+impl HandlerContext {
+    pub fn event_channels(&self) -> impl Iterator<Item = &Sender<Event>> {
```
andrewjstone, line 73: This is a great change, along with removing the redundant vec.

ddm/src/admin.rs
```diff
@@ -429,6 +480,128 @@ impl DdmAdminApi for DdmAdminApiImpl {
...
+                            ctx.log,
+                            "removing external peeer for address object {aobj}"
+                        );
+                    } else {
+                        debug!(ctx.log, "{aobj} != {}", p.config.aobj_name);
+                    }
+                }
+            }
+        }
+        // remove peers back to front so we don't shift the order our from under
```
andrewjstone, line 580:
```suggestion
        // remove peers back to front so we don't shift the order out from under
```

I'm not really sure why this is needed. What am I missing here?

ddm/src/sm/state.rs
```diff
@@ -335,6 +393,26 @@ impl Exchange {
                 self.ctx.event_channels.len()
             );
 
+            // Only send withdraws for expirations that result in a total loss
```
andrewjstone, line 396: This would be a nice invariant to check with a property based test some day.

tests/src/ddm.rs
```diff
@@ -802,6 +862,593 @@ async fn run_quartet_tests(
...
+        }
+    }
+    impl DerefMut for DropDump {
+        fn deref_mut(&mut self) -> &mut Self::Target {
+            &mut self.c
+        }
+    }
+    impl Drop for DropDump {
+        fn drop(&mut self) {
+            // Async just loves to make things difficult, it's taken over all
```
andrewjstone, line 1084: lol 🙉

andrewjstone (approved):
Looks great. Does the right thing via https://github.com/oxidecomputer/omicron/pull/11143.

PR #919, opened 2026-09-23
BGP OPEN fixes

### LINKED ISSUE #852: BGP: Connection Collision Resolution should use ASN as a tie-breaker for eBGP sessions when both peers have the same BGP-ID
[RFC 6286 Section 2.3](https://datatracker.ietf.org/doc/html/rfc6286#section-2.3) covers the use of the ASN as part of collision resolution when both BGP peers have the same BGP-ID:
```
[2.3](https://datatracker.ietf.org/doc/html/rfc6286#section-2.3).  Connection Collision Resolution

   For a BGP speaker that supports the AS-wide Unique BGP Identifier,
   the procedures for connection collision resolution are extended as
   follows to deal with the case in which the two BGP speakers share the
   same BGP Identifier (thus, it is only applicable to an external
   peer):

      If the BGP Identifiers of the peers involved in the connection
      collision are identical, then the connection initiated by the BGP
      speaker with the larger AS number is preserved.

   This extension covers cases in which the 4-octet AS numbers are
   involved [[RFC4893](https://datatracker.ietf.org/doc/html/rfc4893)].
```

We should implement this to ensure correctness in this scenario.

taspelund:
Fix a few bugs related to OPEN handling in BGP:
1) Fix a deadlock lurking in error path of handle_open()
2) Fix an arg ordering bug in collision handling call within Established state
3) Add a missing handle_open() call to collision handling within Established state

Also finish fleshing out the AS-Wide BGP-ID support (RFC 6286):
1) Disallow matching BGP-IDs between iBGP peers
2) Fallback to ASN comparison during collision resolution if BGP-IDs match

Add relevant unit/regression tests.

Fixes: #852


bgp/src/session.rs
```diff
@@ -209,18 +209,59 @@ pub fn collision_resolution(
     exist_direction: ConnectionDirection,
     local_bgp_id: u32,
     remote_bgp_id: u32,
+    local_asn: u32,
+    remote_asn: u32,
 ) -> CollisionResolution {
+    let mut criteria = CollisionResolutionCriteria::HighestBgpId;
```
zeeshanlakhani: Since these are u32's, if you import `cmp::Ordering`, you can do something like this (getting rid of the guard + inner match style):

```rust
let (criteria, ord) = match local_bgp_id.cmp(&remote_bgp_id) {
    Equal => (
        CollisionResolutionCriteria::HighestAsn,
        local_asn.cmp(&remote_asn),
    ),
    ord => (CollisionResolutionCriteria::HighestBgpId, ord),
};

match (ord, exist_direction) {
    (Less, ConnectionDirection::Inbound)
    | (Greater | Equal, ConnectionDirection::Outbound) => {
        CollisionResolution::ExistWins(criteria)
    }
    (Less, ConnectionDirection::Outbound)
    | (Greater | Equal, ConnectionDirection::Inbound) => {
        CollisionResolution::NewWins(criteria)
}
```

bgp/src/session.rs
```diff
@@ -6874,85 +6925,96 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
+                                        om.id,
+                                        self.asn.as_u32(),
+                                        om.asn(),
                                     );
 
                                     session_log!(self,
                                         info,
                                         pc.conn,
-                                        "collision detected in established state (conn_id: {}), collision_detect_established_state enabled",
+                                        "collision detected in established state (conn_id: {}), collision_detect_established_state enabled: {resolution}",
```
zeeshanlakhani: This is being logged twice, right? Once here and then as part of the key listings.

bgp/src/session.rs
```diff
@@ -6874,85 +6925,96 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
...
-                                // the new/colliding connection got an Open.
+                                if let Err(e) =
+                                    self.handle_open(&incoming_conn, om)
+                                {
+                                    session_log!(self, warn, incoming_conn,
+                                        "colliding connection failed to handle open message ({e}), retaining established connection";
+                                        "error" => e.to_string()
+                                    );
+                                    self.bump_msg_counter(msg_kind, true);
+                                    // handle_open already unregistered the
```
zeeshanlakhani: Do we need to increment `self.counters.open_handle_failures` here?

bgp/src/session.rs
```diff
@@ -1809,13 +1850,44 @@ impl<Cnx: BgpConnection + 'static> FsmDriver<Cnx> {
     }
 }
 
+/// Criteria used to determine winner of collision resolution.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum CollisionResolutionCriteria {
```
zeeshanlakhani: I'm iffy if we need this enum vs locally scoping this (in logs for example). Nothing really matches on it, as it's only really being used as a Display string for logging and test assertions?

bgp/src/session.rs
```diff
@@ -9418,58 +9520,290 @@ mod tests {
...
+                            OpenErrorSubcode::UnacceptableHoldTime,
+                        ),
+                    );
+                    tx.send(()).unwrap();
+                });
+
+                // Rejection unregisters the connection and locks its timers.
+                // Bound the wait so recursive locking fails instead of hanging
+                // the test suite. Only join after the worker has completed.
+                rx.recv_timeout(Duration::from_secs(5)).unwrap_or_else(|e| {
```
zeeshanlakhani: Would it be worth handling the the disconnected case and joining the working on that vs panic'ing on timeout. As is, a worker panic shows up as a channel being empty and disconnected, instead of the actual assert.

```rust
if let Err(RecvTimeoutError::Timeout) = 
    rx.recv_timeout(Duration::from_secs(60)) 
{
    panic!("OPEN rejection did not complete within 60s; hold_time={hold_time},
deterministic_collision_resolution={deterministic_collision_resolution}"
    );
}

if let Err(payload) = worker.join() {
    std::panic::resume_unwind(payload);
}
```

I also bumped the timeout from 5s to 60s, as I'm not sure it should be that tight?

bgp/src/session.rs
```diff
@@ -9418,58 +9520,290 @@ mod tests {
...
+                    )
+                });
+                worker.join().unwrap();
+            }
+        }
+    }
+
+    #[test]
+    fn test_resolve_collision_matching_bgp_ids() {
+        // Cover two-octet ASNs, the two/four-octet boundary, and four-octet
```
zeeshanlakhani: minor: inline these to the tuple listings?

bgp/src/session.rs
```diff
@@ -9381,6 +9475,14 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::config::RouterConfig;
+    use crate::connection_channel::{BgpConnectionChannel, channel};
+    use crate::test::{
+        RouteExchange, create_test_session, create_test_session_info,
+    };
+    use CollisionResolution::{ExistWins, NewWins};
+    use CollisionResolutionCriteria::{HighestAsn, HighestBgpId};
```
zeeshanlakhani: nit: maybe worth spelling out the enum / variants in the test below (it's kinda mixed already).

bgp/src/session.rs
```diff
@@ -5643,17 +5712,19 @@ impl<Cnx: BgpConnection + 'static> SessionRunner<Cnx> {
                                             conn_id.short();
                                         );
 
-                                        match collision_resolution(
+                                        let resolution = collision_resolution(
```
zeeshanlakhani: Both the `OpenConfirm` collision path and the new `Established` path just return after `handle_open` fails, right? `OpenSent` does this too for anything that already sent a notification? 

Should this one (and the same below) swap the `stop()` for a `session_timer!(self, connect_retry).stop()` call only? That keeps the existing timer behavior and removes the second notification and the bump. 

I guess we'd miss `PolicyCheckFailed`, where `handle_open` doesn't send anything, and the FSM-error is currently the only notification. Thoughts?

bgp/src/session.rs
```diff
@@ -9418,58 +9520,290 @@ mod tests {
...
+                ErrorSubcode::Cease(
+                    CeaseErrorSubcode::ConnectionCollisionResolution,
+                ),
+            );
+        }
+    }
+
+    #[test]
+    fn test_established_collision_rejects_matching_ibgp_id() {
+        for deterministic_collision_resolution in [false, true] {
```
zeeshanlakhani: does the true iteration do anything here?

zeeshanlakhani (commented):
Here's my first set of comments/thoughts @taspelund. I'd also find ways to reference some of the RFCs through links in fn public docs.

bgp/src/session.rs
taspelund: It's definitely only used to facilitate logging of collision resolution results (not just which connection won, but also for what reason it won), but I wanted to keep the resolution function pure instead of passing in a Logger.

Do you have other ideas on how we could return winner + criteria here instead of using an enum like this? I'm certainly open to other options, I just didn't see anything immediately obvious

bgp/src/session.rs
taspelund: ...yep, sure do!  Also, looking at the rest of the FSM it seems like we only ever increment this in one place (within `OpenSent`)... I'll look at adding this call inside handle_open so it gets bumped during the unhappy paths in a centralized manner.

bgp/src/session.rs
taspelund: The logic being tested isn't dependent on the value of `deterministic_collision_resolution`, so I can yank out the loop and just run the test once

bgp/src/session.rs
zeeshanlakhani: Yeah, you may be able to make this work as a struct, but I'm not truly against the enum here either. We should link the RFCs here as these are public docs.

bgp/src/session.rs
taspelund: Sort of. It's now in both the main log string and in one of the slog keys. In theory I'm not opposed to putting the resolution in its own key so we can do some more structured log searching (e.g. w/ rhai via looker), but the new display string isn't super friendly for that IMO. I'll yank out the slog key.

...I'll also shorten that log string, it's awful.

bgp/src/session.rs
taspelund: Let me noodle on this one a little bit. I like how concise it is, but I'm not sure it's quite as easy to understand

bgp/src/session.rs
taspelund: Let me look into this. I just pushed a couple commits to address the more minor suggestions, but I will circle back

bgp/src/session.rs
taspelund: Let me look into this. I just pushed a couple commits to address the more minor suggestions, but I will circle back

